PR #44, opened 2022-04-02
feat(core): add View core component

mrjackdavis:
I found the need for a more primitive component than `Stack` when using `styled` and in some other circumnstances. I found myself overriding `YStack` to be... well not a stack.

WDYT?


natew:
Looks good, I'll get this in alongside the beta PR.

natew:
Just out of curiosity, was it a default style on Stack that made you want View? Or something else?

natew:
I think since Stack is basically just a renamed View, going to leave this one. I will test if styled((react-native').View) works.

PR #94, opened 2022-06-07
support for radix colors steps

seeden:
added support for original radix step1-12 variables


natew:
Only thing I'm worried about is web initial CSS size. May try and get one more optimization there in place before this.

seeden:
If you are worried about size we can use property for it which can add it.

natew:
This snuck into master, in the future may still make it configurable.

PR #233, opened 2022-09-23
Add Radix License

penx:
Radix license requires that this "be included in all copies or substantial portions of the Software"


penx:
Closing as already present in https://github.com/tamagui/tamagui/blob/master/LICENSE

PR #234, opened 2022-09-23
feat(tailwind): Tamagui tailwind gives you Tailwind style variants and Stack + Text

natew:
May not touch this for a bit, but it's a good thing to have even if only for experimental sake, and since it will use tons of variants, it may motivate me to improve `styled()` flattening on the compiler.

Allows for:

```tsx
import { Stack } from '@tamagui/tailwind'

export default () => <Stack w-0 />
```

### DIFF
--- a/packages/tailwind/package.json
+++ b/packages/tailwind/package.json
@@ -1,0 +1,39 @@
+{
+  "name": "@tamagui/tailwind",
+  "version": "1.0.1-beta.174",
+  "sideEffects": [
+    "*.css"
+  ],
+  "source": "src/index.ts",
+  "types": "./types/index.d.ts",
+  "main": "dist/cjs",
+  "module": "dist/esm",
+  "module:jsx": "dist/jsx",
+  "files": [
+    "src",
+    "types",
+    "dist"
+  ],
+  "scripts": {
+    "build": "tamagui-build",
+    "watch": "tamagui-build --watch",
+    "lint": "eslint",
+    "lint:fix": "yarn lint --fix"
+  },
+  "dependencies": {
+    "@tamagui/core": "^1.0.1-beta.174"
+  },
+  "peerDependencies": {
+    "react": "*",
+    "react-dom": "*"
+  },
+  "devDependencies": {
+    "@tamagui/build": "^1.0.1-beta.174",
+    "react": "*",
+    "react-dom": "*"
+  },
+  "publishConfig": {
+    "access": "public"
+  },
+  "gitHead": "a49cc7ea6b93ba384e77a4880ae48ac4a5635c14"
+}
--- a/packages/tailwind/src/Stack.tsx
+++ b/packages/tailwind/src/Stack.tsx
@@ -1,0 +1,7 @@
+import { Stack as TStack, styled } from '@tamagui/core'
+
+import { variants } from './variants'
+
+export const Stack = styled(TStack, {
+  variants,
+})
--- a/packages/tailwind/src/Text.tsx
+++ b/packages/tailwind/src/Text.tsx
@@ -1,0 +1,7 @@
+import { Text as TText, styled } from '@tamagui/core'
+
+import { variants } from './variants'
+
+export const Text = styled(TText, {
+  variants,
+})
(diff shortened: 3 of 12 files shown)


natew:
The typescript types may help as a way to get the full list of props structured?

natew:
If anyone feels motivated to copy all those presets (🥵) would be lovely :)

nandorojo:
The issue with this approach for adoption (I think) is that you don't get the useful tailwind VSCode plugin etc. which you'd otherwise get with the `className`/`tw` prop.

natew:
What does that get you beyond typescript? Shows colors right?

Its possible to do a single tw variant function as an alternate approach.

alitnk:
Maybe there's potential for integrating [Typewind](https://github.com/Mokshit06/typewind) here?

natew:
@alitnk Tamagui optimizer won't understand that, but with variants we can really have any API we want:

```tsx
<Stack tw="px-0" />
<Stack px-0 />
```

nandorojo:
would `tw` work with spaces?

jerryjfwang:
we're thinking about migrating from NativeWind to Tamagui. would the most "drop-in" replacement be to define variants that map to tailwind classes?

nandorojo:
Not really, because it wouldn’t work well with responsive design. But if you have no media queries then I assume that’s fine

natew:
You could make media query variants right? Maybe not understanding them as much. 

I think if you want this to work like this:

```
<Stack px-0 />
```

I would do just a single catch all functional variant and basically do: val.split to get the prefix and suffix, dynamically build the style. That should be pretty compact relatively.

Isos9:
Hi @natew, I would like to help for the presets on this, how can I do it on this branch ? do I need to fork ?

redbar0n:
for reference: https://github.com/tamagui/tamagui/discussions/114

jerryjfwang:
> You could make media query variants right? Maybe not understanding them as much.
> 
> I think if you want this to work like this:
> 
> ```
> <Stack px-0 />
> ```
> 
> I would do just a single catch all functional variant and basically do: val.split to get the prefix and suffix, dynamically build the style. That should be pretty compact relatively. On my phone I can type this out more later.

Hey @natew,

A functional variant wouldn't give type safety over all possible tailwind style props, right? To do so, we'd have to hardcode each prop like you've done with width (eg: w-0, w-1, etc.) on the tailwind branch?

natew:
You’d just have to type it manually but it would save bundle size

natew:
I think if we really wanted to do it right we’d have to figure out a way to have code removal and variant removal or shaking, but that’s such an intense task especially with code splitting. But for most apps functional variant is maybe 10kb max on top of Tamagui so not bad

jerryjfwang:
Got it — that makes sense. What's the best practice for manually adding the types? Appreciate the help 🙏

jerryjfwang:
> You’d just have to type it manually but it would save bundle size

@natew how do you type the props manually? I've tried to do this two ways without success:

Method #1: defining the argument type for the catch all variant function
```
type TailwindProps = "h-0" |  ... 679 more ... | "min-h-60";

type tailwindVariants = {
  "...": (val: TailwindProps) => Record<string, string | number>
}

const StyledText = styled(Text, {
  variants: {
    ...variants,
    ...tailwindVariants,
  } as const,
});
```

Method #2: Asserting the type of variants
```
type TailwindProps = "h-0" |  ... 679 more ... | "min-h-60";
type TailwindPropVariants = {
  [K in TailwindProps]: {
    true: Record<string, string | number>;
  };
};

type AllVariants = typeof variants & TailwindPropVariants

const StyledText = styled(Text, {
  variants: {
    ...variants,
    ...tailwindVariants,
  } as unknown as AllVariants,
});
```

natew:
@jerryjfwang missed this comment. you can probably use the internal createComponent API:

CleanShot 2023-09-11 at 21 36 17@2x

It's like styled but a bit more low level, it takes 3 options: <Props, Ref, BaseProps>

hmmm... now looking at it I may need to export a better helper here...

PR #237, opened 2022-09-25
Rename Native to React Native in benchmarks, for clarity

redbar0n:
To avoid people thinking the benchmark is against pure native (iOS/Android).


natew:
Since this was out of date just cherry picked some in, thanks.

PR #255, opened 2022-10-03
Update ButtonProps to support "value"

### LINKED ISSUE #252: [bug] allow property "value" on Button component
**Describe the bug**
Allow property "value" for `Button` component

**To Reproduce**

```ts
<Button value="list">list</Button
```

Currently this does not work / typescript breaks. But there are certain scenarios where the value of `<button>` needs to be defined.

dohomi:
add `value` to ButtonProps. closes #252 

@natew is that sufficient for making it work?

### DIFF
--- a/packages/button/src/Button.tsx
+++ b/packages/button/src/Button.tsx
@@ -49,6 +49,7 @@ export type ButtonProps = Omit<TextParentStyles, 'TextComponent'> &
      * adjust internal space relative to icon size
      */
     scaleSpace?: number
+    value?: string
   }
 
 export const ButtonFrame = styled(ThemeableStack, {


dohomi:
@natew can you have a look at this small change? Integration of Quill requires the "value" prop on <button/>

bang88:
IMO `Button·  should not have this prop. but you CAN extend from it.

dohomi:
@bang88 Quill uses the value and its from the web point of view definitely a property which should be supported: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/button#attr-value
For the native side I am not sure though..

bang88:
The native side doesn't support this prop. We could define a custom Button with a value prop or create a Rich editor component in your package.

dohomi:
Would it make sense to use Platform specific handler?

bang88:
Not sure 😄

natew:
I think yea in this case for now just fork button and add it for that use case, we need to sync with react native people to see what API they would want to make this work. Closing this but we can open an issue.

PR #284, opened 2022-10-17
Lucide Icons

sohcah:
- Creates a new `@tamagui/lucide-icons` package, based on the `@tamagui/feather-icons` library
- Replaces usages of feather icons with lucide-icons throughout most examples and on the site
- Adds Lucide Icons demo page to site

Does not yet update the starters to use lucide-icons since I cannot install `@tamagui/lucide-icons` in them until it's published.


natew:
Did a few small changes and merged via another branch - I think it gave credit properly if not idk how but happy to fix.

PR #337, opened 2022-11-05
fix(site): fixed typo

MichaelKilbane:
- Added a fix to the core configuration page, correcting:
  "The you'll just want to bring along any fonts..." -> "Then you'll just want to bring along any fonts..."


natew:
Thank you

PR #339, opened 2022-11-05
feat(site): Add design system docs

natew:


nandorojo:
This is cool. Small detail, would you consider keeping Radix's 1-based index for colors (1 - 12 instead of 0 - 11)? Not a big thing I suppose, but I've gotten used to it over time

PR #363, opened 2022-11-14
fix(core): delete weight/style when face applies

MichaelKilbane:
Delete font weight and style when the font family is being overridden due to a native font face.

This is due to a bug where font weights 700 or above will look for a bold variant that doesn't exist, and fall back on the platform font.


natew:
Can you confirm this with a variety of tests cases? We have some testing stuff strewn around the repo, but i'd like start getting snapshot tests of stuff like this. Could be easy to add one using the vite setup.

I'd also need to run this and see the difference in effect. This merge area tricky, really appreciate you making these PRs btw.

PR #364, opened 2022-11-15
fix(list-item): Make sure icon after spacer does not use the full width if not content will end up being cropped

sync:
### Before
Screenshot 2022-11-15 at 3 39 48 pm

### After
Screenshot 2022-11-15 at 3 40 13 pm


natew:
Wouldn't we want something like flex space around / between? Haven't tested this yet.

sync:
> Wouldn't we want something like flex space around / between? Haven't tested this yet.

yes probably space between is also good.I think the the labels / content is taking the full width so they are pushing the icon to the right already.

PR #372, opened 2022-11-17
Improve `ThemeInverse`

nandorojo:
Currently, `ThemeInverse` resets the component tree, which isn't good for cases of getting conditionally rendered. It also double renders on native for Web's SSR support, which is fixed here by deriving the theme rather than triggering a new render.


natew:
I like the direction, `core` no longer depends on react-native so just have to make some tweaks.

natew:
Cut a release let me know if it works

PR #373, opened 2022-11-18
perf(core): avoid tree recalculations on changing themeInverse/theme

natew:


packages/core/src/views/ThemeInverse.tsx
```diff
@@ -1,32 +1,40 @@
...
-    return children as JSX.Element
+  if (isWeb) {
+    // for SSR must be in effect
+    useIsomorphicLayoutEffect(() => {
+      setName(toggleTheme(themeName || defaultTheme || ''))
+    }, [defaultTheme, themeName])
   }
 
-  return <Theme name={name}>{children}</Theme>
+  return props.disable ? null : name
```
nandorojo, line 27: Does this toggle the theme on native? I think it should be: 

```tsx
props.disable ? null : isWeb ? name : toggleTheme(name)
```

But maybe better to not have so many ternaries.

packages/core/src/views/ThemeInverse.tsx
```diff
@@ -1,32 +1,40 @@
...
-  disable,
-}: {
-  children: React.ReactNode
+export type ThemeInverseProps = {
   disable?: boolean
-}) => {
+}
+
+const toggleTheme = (themeName: string) => {
+  return themeName.startsWith('light') ? 'dark' : 'light'
```
nandorojo, line 12: I wonder what the expected behavior is here. 

If the theme is `dark_red`, do I want `light_red` or `light`? Or would it be `<ThemeInverse reset />` to get it to be `light`? Current behavior is probably fine, but to think about it before I get too deep...

PR #374, opened 2022-11-19
fix(core): Cleanup and improve reverse logic, allow reverse on useThe…

natew:
…me, and make inverse move from sub-themes (dark_red reversed => light_red instead of just light)


packages/core/src/hooks/useTheme.tsx
```diff
@@ -273,13 +240,13 @@ export const useChangeThemeEffect = (
     }
   }
 
-  if (!isServer) {
+  if (!isWeb || !isServer) {
```
nandorojo, line 243: unless i’m missing something, there shouldn’t be a need to run an effect on native, since all state can be derived from dimensions.

packages/core/src/views/Theme.tsx
```diff
@@ -3,17 +3,15 @@ import React, { memo, useMemo } from 'react'
...
 export const Theme = memo(function Theme(props: ThemeProps) {
-  const { name, theme, themeManager, themes, className } = useChangeThemeEffect(
-    props.name,
-    props.componentName,
-    props
-  )
+  const { name, theme, themeManager, themes, className } = useChangeThemeEffect(props)
+
+  if (props.inverse) {
+    console.warn(`wtf is this`, name, className, theme)
```
nandorojo, line 13: is this an oversight lol

natew: Haha yea still wip but good to note. I have a lint rule should fail for ci

natew: But evidently isn’t

PR #378, opened 2022-11-21
fix config weight usage to be string

aleedmanath:
as discussed on discord, examples were showing font weight in a number rather than string -- the former which is turned into pixels and not picked up as a font.

i also changed the font `face` key to be a string. 

the first commit is for docs fix, the second commit changes font face to string in code if you want to keep that too.

### DIFF
--- a/apps/site/data/docs/core/configuration.mdx
+++ b/apps/site/data/docs/core/configuration.mdx
@@ -80,8 +80,8 @@ const interFont = createFont({
     // ...
   },
   weight: {
-    4: 300,
-    6: 600,
+    4: '300',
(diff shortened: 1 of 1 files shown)


natew:
The face one shouldn't change, all object keys are strings even if they are defined as numbers.

natew:
Minor nit but single quotes too on the docs plz

aleedmanath:
> The face one shouldn't change, all object keys are strings even if they are defined as numbers.

ah, wow! JS :) my fault, in CLJS there's various type of map keys

natew:
thank you!

PR #384, opened 2022-11-25
Fixed map key not provided issue of Subtitle

loveholly:


natew (approved):

natew:
Thanks! Was driving me crazy but never looked into the file.

loveholly:
> Thanks! Was driving me crazy but never looked into the file.

I found it just when I was debugging `TextInput` Component in the documentation site😂.

PR #398, opened 2022-12-02
Ease the dependency on @expo/match-media

alisherks:
I wanted to bring attention to the hard dependency on @expo/match-media. IMO core module should not force the bare react-native developer to install the expo module, which requires changes to the native code. 

In this PR I drafted an option to make the match-media function pluggable. It will simplify the installation process for react-native developers who might not even use media props.

@natew, let me know how we can tackle this issue.

### DIFF
--- a/apps/site/data/docs/core/use-media.mdx
+++ b/apps/site/data/docs/core/use-media.mdx
@@ -3,6 +3,18 @@ title: useMedia
 description: Designing for different size screens.
 ---
 
+`@tamagui/core` doesn't provide media capabilities out of the box to native apps. To enable media queries, you need to provide matchMedia implementation by calling `setupMatchMedia` in your configuration file:
+```tsx
+import { createMedia } from '@tamagui/react-native-match-media'
+
+export default createTamagui({
+  media: createMedia({
+    xs: { maxWidth: 660 },
+    // ...
+  })
+})
+```
+
 Define your media rules in the media object of your `tamagui.config.ts`:
 
 ```tsx line=2-6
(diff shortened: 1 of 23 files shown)


natew:
Happy to accept this change, authorized you for vercel as well on this PR.

We'd need the following steps:

1. Update docs to note this
2. Make `packages/tamagui` automatically set this up by default
3. Update apps/site, apps/kitchen-sink, starters/*, make sure they work

If you can make those changes happy to accept.

nandorojo:
If we're removing the dependency altogether, a better implementation would be to pull over the polyfill in a tree-shaken native file. I also don't think conditional `require` in `try`/`catch` is well-supported.

natew:
Yea was going to comment as much, this would best be done as some sort of @tamagui/media-query-plugin or something and yes use `.native`.

nandorojo:
One option could be to make a match-media alternative that doesn’t use expo, and tell tamagui users to alias from expo’s to the other one in their native babel config. This would still presumably require another native library, though.

natew:
Can we have RN Dimensions API as an option for a plugin? I’d guess that’s an easier install but does it have downside?

nandorojo:
I was wondering that too — presumably Expo doesn’t use it for a reason, though I don’t know what tbh. Is device rotation relevant to us? I haven’t looked closely at their code, but can’t we just compare height and width to know the rotation? I may be missing something though. It’s possible dimensions doesn’t account for rotation?

natew:
Looking better too btw, I'd say ideally we add `@tamagui/react-native-media-driver` and have it be the default for `tamagui` and just use Dimensions, and then we may not even need the expo one (or we can add that too as an option).

natew:
image

Lets mimic this API surface.

So `media` should accept the existing object for backward compat and just work on the web, for native must use `media: createMedia({ sm: {} })` imported from a driver. We can warn if not the MediaDriver type on native.

alisherks:
Added default `matchMedia` implementation as 'react-native-media-driver' based on Dimensions API (probably, not accurate as native implementation). Internally, it does the same thing as previously but API looks nicer.

I tried to hide `setupMatchMedia` as an internal API and build a contract with a configuration object (similar to animation drivers), but couldn't figure out how to properly type it to support backward compatibility.

Currently, the warning message should show up when the user defines media options inside the configuration.

Issue #400, opened 2022-12-02
[bug] Custom Button breaks build


apperside:
Hello,
I am trying to create a custom button implementation starting from the sample taken here

https://tamagui.dev/docs/components/button/1.0.0-beta.161#customization-advanced

But when I add that file and export it from /packates/ui/index.ts, it doesn't build anymore.

What I did is creating a project from scratch with `npm create tamagui-app@latest`, adding a Button.tsx to ui package, and export it. 
Here is the result

<img width="1630" alt="Screenshot 2022-12-02 at 19 10 58" src="https://user-images.githubusercontent.com/5955338/205358354-cdc32439-446f-4add-ab14-cb47be1118b3.png">

Here is a reproduction erepo

https://github.com/apperside/tamagui-custom-button-bug

**TAMAGUI VERSION: 1.0.1-beta.226

**System Info (please complete the following information):** 

``` bash
~/dev/tamagui-button/tamagui-button git:(main) npx envinfo --system --npmPackages --binaries --browsers

  System:
    OS: macOS 13.0
    CPU: (16) x64 Intel(R) Core(TM) i9-9880H CPU @ 2.30GHz
    Memory: 20.48 MB / 16.00 GB
    Shell: 5.8.1 - /bin/zsh
  Binaries:
    Node: 16.17.0 - /usr/local/bin/node
    Yarn: 3.2.3 - ~/.nvm/versions/node/v16.17.0/bin/yarn
    npm: 8.15.0 - ~/.nvm/versions/node/v16.17.0/bin/npm
  Browsers:
    Chrome: 107.0.5304.121
    Firefox: 107.0
    Safari: 16.1
  npmPackages:
    @babel/runtime: ^7.18.9 => 7.20.6 
    @manypkg/cli: ^0.19.1 => 0.19.2 
    @types/react: ^18.0.15 => 17.0.21 
    @types/react-native: ^0.69.6 => 0.69.15 
    eslint: ^8.21.0 => 8.28.0 
    prettier: ^2.7.1 => 2.8.0 
    turbo: ^1.4.2 => 1.6.3 
    typescript: ^4.7.4 => 4.9.3 
```

I can also attach the output of `DEBUG=tamagui yarn web` but it's really a lot of stuff and I cannot paste it here

jonsherrard:
If you add '@my/ui' to `useTM` in `next.config.js` and things will start working https://github.com/tamagui/tamagui/pull/401

natew:
[...]

PR #401, opened 2022-12-02
fix(starter): Add `@my/ui` package to Next.js transpile modules

jonsherrard:
Fixes: https://github.com/tamagui/tamagui/issues/400 and errors in the `ui` folder containing `You may need an appropriate loader to handle this file type,`


natew:
Thanks for finding this, was a bit involved made the proper fix here 8cbb64e00527dc0d45fefa6aeee4b82551a0ef97

PR #404, opened 2022-12-03
fix(core): Find border radius tokens over space tokens

natew:
Sorry had a bad master branch locally, third times the charm...

Had to do this inside the src dir, @nandorojo 🙏 happy to accept it as your if you want to update your PR


nandorojo:
@natew no worries on accepting as my PR, but I noticed that I'm still getting the same bug on `rc.0`. Is that expected?

PR #411, opened 2022-12-05
fix(core): Hash themeKeys in getThemeCSSRules

peterjskaltsis:
It looks like we are hashing correctly and processing out `.` characters etc for the base colors in `tokens`, but not for theme token names.

This was resulting in situations where you'd see `<div class="--backgroundsecondary1a ..."/>` but in the CSS panel, no CSS variable `--backgroundsecondary1a` existed.

My solution is to hash the themeKeys when processing theme css rules via `createTamagui()` -> `getThemeCSSRules()`.

So far I haven't seen this affect any other pieces.


natew:
Merged here, need to merge differently with my cli to make it automatically show this pr merging :/

https://github.com/tamagui/tamagui/commit/93ea6c0b1b4217e550d8f044920605ce3fbe6cf3

PR #412, opened 2022-12-05
Use built in transpilePackages

pronevich:


natew:
Still figuring out github but merged this with a fix for yarn lock

PR #430, opened 2022-12-12
[Issue] pressStyle={{ bg: ... }} has issues on native

skyriverbend:
There's no way to get rid of the grey background color that occurs when pressing on pressable text in native.

You'll notice in this example that it works as expected on Web, but in Native we get a grey background color no matter what. The example with `bg: 'red'` seems to work, but we still end up with a grey border.

This grey background color behavior seems to be default in React Native. I did some basic research and saw that you can remove it like this:

```
  <TouchableOpacity onPress={() => console.log('pressing text')} activeOpacity={1}>
    <RNText style={{ color: 'white' }}>Press this text</RNText>
  </TouchableOpacity>
```

Just wondering how hard it would be to implement this sort of thing in Tamagui so that we can configure the background pressStyle using tamagui component props?

Thanks!

### DIFF
--- a/starters/next-expo-solito/packages/app/features/home/screen.tsx
+++ b/starters/next-expo-solito/packages/app/features/home/screen.tsx
@@ -5,6 +5,7 @@ import {
   H1,
   MyComponent,
   Paragraph,
+  SizableText,
   Separator,
   Sheet,
   XStack,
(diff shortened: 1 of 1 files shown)


natew:
Fixed in [a8832e2](https://github.com/tamagui/tamagui/pull/438/commits/a8832e221968dee8ecff9ece73ae37e5c08d9f17)

PR #431, opened 2022-12-12
fix: let useCurrentColor access token colors

peterjskaltsis:
Currently when working with icons inside components and using the `getThemedIcon`, you are unable to pass color tokens from `tokens.color` which have not been declared explicitly in `createTheme({ ... })`.

This PR adds a fallback to the `useCurrentColor` hook to also look for the color in `tokens.color` when returning the current color.


peterjskaltsis:
I also noticed we're not using color in `useGetThemedIcon`. Is that intended @natew ?

natew:
Great find!

PR #464, opened 2022-12-29
build: run ci checks on pull requests

iBotPeaches:
I'm noticing PRs aren't running CI - the workflow is only hooked to "push" right now for branches local to this project. This should allow incoming PRs to also run the `checks.yaml` workflow.


natew:
Thanks, was wondering about that.

PR #493, opened 2023-01-04
Add Phosphor Icons

mwood23:
Adds phosphor icons as an option for developers. There's a lot more icons than `lucide` and there is not `phosphor-static` package so the PR is a bit chunky.

I did run into a hydration error when developing on the docs site, but it didn't look related to this PR.

Totally understand if you don't want to take on the extra maintenance burden of two icon packs.

One feature I added was a context API for icons. I think that's something valuable to move into `lucide-icons` as well.

https://user-images.githubusercontent.com/13633613/210508165-678a3ba8-3a5b-479a-b60b-b6023d617b94.mp4


mwood23:
Published as a package in case someone needs to use this now: https://www.npmjs.com/package/tamagui-phosphor-icons

dohomi:
@mwood23 nice package do you have it published in Github as well?

dohomi:
great package, the `weight` prop makes a big benefit compared to lucide icon set 👍

mwood23:
Published this on my personal repo and added docs on how to keep perf high dealing with all the icons. You can find it here: https://github.com/mwood23/tamagui-phosphor-icons

natew:
Btw would love to merge this actually, would just need to change the strategy to not build the icons up-front but instead generate them on pre-publish, and have a test added that ensures they work fine in a simple render before pre-publish completes.

I can help contribute the test part.

chen-rn:
> Published this on my personal repo and added docs on how to keep perf high dealing with all the icons. You can find it here: https://github.com/mwood23/tamagui-phosphor-icons

This works really well :) Great job!
Can't wait for this to become a part of the main build!

dohomi:
any chance of getting this included in @tamagui namespace?

natew:
@dohomi we need a way to have packages that aren't updated often and are heavier, because otherwise this will weigh down the repo quite a bit. Maybe a community maintained `tamagui/phosphor-icons` or something.

PR #506, opened 2023-01-06
fix: update starters to use transition in next theme change handler

### LINKED ISSUE #504: [bug] SSR hydration error with Suspense and OS dark mode color preference
**Describe the bug**

When a Next.js application using tamagui (with the recommended theme swapping method) renders a suspense boundary containing Tamagui components, and the user has dark mode color preference set `prefers-color-scheme: dark`, the following error is thrown:

> This Suspense boundary received an update before it finished hydrating. This caused the boundary to switch to client rendering. The usual way to fix this is to wrap the original update in startTransition.

**To Reproduce**

See bug reproduction repo here: https://github.com/nderscore/tamagui-suspense-demo

Changes from the starter:

* Nest a [very-bare-bones suspense boundary](https://github.com/nderscore/tamagui-suspense-demo/blob/main/packages/app/components/SuspenseBoundary.tsx) component inside global providers:
  * https://github.com/nderscore/tamagui-suspense-demo/blob/main/packages/app/provider/index.tsx#L10
* Add timer-delayed mock suspense API call to user details page:
  * https://github.com/nderscore/tamagui-suspense-demo/blob/main/packages/app/features/user/detail-screen.tsx#L24

nderscore:
Fixes #504

* Update the theme provider examples in the starter projects to call `startTransition` in their change handlers
  * Here is the fix as applied to the bug reproduction repo: https://github.com/nderscore/tamagui-suspense-demo/tree/fix
  * (the `site` app for the tamagui homepage already does this)
* Remove package resolution locking `@types/react` to an old 17.x version which predates the existence of `startTransition`
  * I might be missing some prior knowledge as to why this was there in the first place, but I didn't notice any issues after removing it.

### DIFF
--- a/apps/starter/apps/next/pages/_app.tsx
+++ b/apps/starter/apps/next/pages/_app.tsx
@@ -1,13 +1,13 @@
 import '@tamagui/core/reset.css'
 import '@tamagui/font-inter/css/400.css'
 import '@tamagui/font-inter/css/700.css'
+import 'raf/polyfill'
 
 import { NextThemeProvider, useRootTheme } from '@tamagui/next-theme'
 import { Provider } from 'app/provider'
 import Head from 'next/head'
-import React, { useMemo } from 'react'
+import React, { startTransition } from 'react'
 import type { SolitoAppProps } from 'solito'
-import 'raf/polyfill'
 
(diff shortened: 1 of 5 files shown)


natew:
Thanks!

PR #525, opened 2023-01-11
cancelAnimationFrame / floating-ui fixes

### LINKED ISSUE #514: [bug] Select - SelectScrollButton scrolls indefinitely after initial pointer hover
Seen on: v1.0.3

WEB only.

**Describe the bug**

After hovering either a `SelectScrollDownButton` or `SelectScrollUpButton`, the list of select items will scroll indefinitely until it reaches the end of the list.

**To Reproduce**

Can be seen happening on demo for `Select` on the docs site: https://tamagui.dev/docs/components/select/1.0.0

Make the height of your browser window relatively small, then interact with the select scroll button (trigger an enter and leave)

**Investigation**

After adding some logs here...

```diff
diff --git a/packages/select/src/SelectScrollButton.tsx b/packages/select/src/SelectScrollButton.tsx
index 851cda66c8..5dc4cf0dab 100644
--- a/packages/select/src/SelectScrollButton.tsx
+++ b/packages/select/src/SelectScrollButton.tsx
@@ -133,10 +133,12 @@ const SelectScrollButtonImpl = React.memo(
           top={y || 0}
           width={`calc(${(floatingRef?.current?.offsetWidth ?? 0) - 2}px)`}
           onPointerEnter={() => {
+            console.log('enter', frameRef.current)
             statusRef.current = 'active'
             let prevNow = Date.now()
 
             function frame() {
+              console.log('frame', frameRef.current)
               if (element) {
                 const currentNow = Date.now()
                 const msElapsed = currentNow - prevNow
@@ -171,6 +173,7 @@ const SelectScrollButtonImpl = React.memo(
             frameRef.current = requestAnimationFrame(frame)
           }}
           onPointerLeave={() => {
+            console.log('leave', frameRef.current)
             statusRef.current = 'idle'
             cancelAnimationFrame(frameRef.current)
           }}
```

Based on the following output in the console:

```
...
frame 1202
leave 1221
frame 1221
...
```

It would seem that `cancelAnimationFrame` is being called before the the rAF callback, yet the rAF callback still executes 🤔 

It can be worked around by checking the value of `statusRef.current` inside the rAF callback, but that doesn't seem like it should be necessary - and `statusRef.current = 'idle'` would probably need to be added to the other places where the `frameRef` gets cancelled.

### LINKED ISSUE #519: [bug] Select - floating-ui autoUpdate runs continuously after closing select
Last seen on: 1.0.4

**Describe the bug**

In the web implementation of the `Select` component, after opening the select once and closing it, `floating-ui`'s `autoUpdate()` function gets called continuously from that point onwards, despite the select's popover being closed.

**To Reproduce**

1. Open any page with a `<Select>` component on it, such as the [docs page for the `Select` component](https://tamagui.dev/docs/components/select/1.0.0).

1. Run this snippet in devtools console to add logging to all `requestAnimationFrame`/`cancelAnimationFrame` calls:

```js
(() => {
  const RAF = window.requestAnimationFrame;
  // just in case, get native cancelAnimationFrame implementation:
  const iframe = document.createElement('iframe');
  iframe.style.display = 'none';
  document.body.appendChild(iframe);
  const CAF = iframe.contentWindow.cancelAnimationFrame;

  // add logging wrappers:
  window.requestAnimationFrame = (callback) => {
    const frame = RAF(callback);
    console.log('RAF', frame, callback);
    return frame;
  };

  window.cancelAnimationFrame = (frame) => {
    console.log('CAF', frame);
    return CAF(frame);
  };
})();
```

1. Open the select

1. Close the select

1. See `autoUpdate` still getting called continuously

nderscore:
Fixes #514
Fixes #519

* Restore native `cancelAnimationFrame` on web
  * The `tamagui` package was adding a `try/catch` wrapper to `cancelAnimationFrame`, but this wrapper referenced itself (it shadows its own variable name) instead of the original `cancelAnimationFrame` method, causing an infinite loop whose error gets swallowed.
  * This fixes the scroll buttons on the select as well as the never-ending `autoUpdate` after close
* Revert the change from cf2951a6
  * Change was not necessary and doesn't seem to follow examples from floating-ui docs
* Pass `open` prop to `useFloating()` hook call in `SelectScrollButton`
  * Not 100% sure this is necessary, but it felt right 😅

### DIFF
--- a/packages/select/src/SelectImpl.tsx
+++ b/packages/select/src/SelectImpl.tsx
@@ -112,7 +112,7 @@ export const SelectInlineImpl = (props: SelectImplProps) => {
   const { x, y, reference, floating, strategy, context, refs } = useFloating({
     open,
     onOpenChange: setOpen,
-    whileElementsMounted: open ? autoUpdate : undefined,
+    whileElementsMounted: autoUpdate,
     placement: 'bottom-start',
     middleware: fallback
       ? [
--- a/packages/select/src/SelectScrollButton.tsx
+++ b/packages/select/src/SelectScrollButton.tsx
@@ -78,6 +78,7 @@ const SelectScrollButtonImpl = React.memo(
       const frameRef = React.useRef<any>()
 
       const { x, y, reference, floating, strategy, update, refs } = useFloating({
+        open: open && isVisible,
         strategy: 'fixed',
         placement: dir === 'up' ? 'top' : 'bottom',
         middleware: [offset(({ rects }) => -rects.floating.height)],
--- a/packages/tamagui/src/setup.ts
+++ b/packages/tamagui/src/setup.ts
@@ -14,8 +14,10 @@ if (typeof requestAnimationFrame === 'undefined') {
   globalThis['requestAnimationFrame'] = setImmediate
 }
 
+const cancelAnimationFrame = globalThis.cancelAnimationFrame
+
 // for vite / Animated.spring()
-global.cancelAnimationFrame = (x) => {
+global.cancelAnimationFrame = (x: number) => {
   try {
     cancelAnimationFrame(x)
   } catch {


natew:
You sure we shouldn't turn it off, it seems to imply here:

image

https://floating-ui.com/docs/react

Issue #529, opened 2023-01-12
[bug] npm create tamagui@latest not working since this morning


rvasseur31:
**To Reproduce**

```
npm create tamagui@latest
```

```
> yarn install

➤ YN0000: ┌ Resolution step
➤ YN0000: └ Completed
➤ YN0000: ┌ Fetch step
➤ YN0000: └ Completed in 0s 791ms
➤ YN0000: ┌ Link step
➤ YN0007: │ turbo@npm:1.6.3 must be built because it never has been before or the last one failed
➤ YN0007: │ esbuild@npm:0.15.6 must be built because it never has been before or the last one failed
➤ YN0007: │ core-js-pure@npm:3.26.1 must be built because it never has been before or the last one failed
➤ YN0007: │ myapp-monorepo@workspace:. must be built because it never has been before or the last one failed
➤ YN0009: │ myapp-monorepo@workspace:. couldn't be built successfully (exit code 1, logs can be found here: /private/var/folders/z6/f9ls1kwd3lq20kc9v3ng1s2h0000gn/T/xfs-32f12261/build.log)
➤ YN0000: └ Completed in 9s 855ms
➤ YN0000: Failed with errors in 10s 914ms
[tamagui] error installing with yarn
 yarnpkg exited with non-zero code: 1
node:child_process:960
    throw err;
    ^

Error: Command failed: node /Users/rvasseur/.npm/_npx/d01fabec74a282ea/node_modules/create-tamagui/dist/index.js 
    at checkExecSyncError (node:child_process:885:11)
    at Object.execSync (node:child_process:957:15)
    at Object.<anonymous> (/Users/rvasseur/.npm/_npx/d01fabec74a282ea/node_modules/create-tamagui/run.js:6:26)
    at Module._compile (node:internal/modules/cjs/loader:1218:14)
    at Module._extensions..js (node:internal/modules/cjs/loader:1272:10)
    at Module.load (node:internal/modules/cjs/loader:1081:32)
    at Module._load (node:internal/modules/cjs/loader:922:12)
    at Function.executeUserEntryPoint [as runMain] (node:internal/modules/run_main:82:12)
    at node:internal/main/run_main_module:23:47 {
  status: 1,
  signal: null,
  output: [ null, null, null ],
  pid: 35986,
  stdout: null,
  stderr: null
}

Node.js v19.2.0
npm ERR! code 1
npm ERR! path /Users/rvasseur/Projects/poc
npm ERR! command failed
[...]

PR #530, opened 2023-01-12
fix(starter): fix the `npm create tamagui` command

rvasseur31:
I don't know why but this devDependencies has disappeared mysteriously

This PR is link to https://github.com/tamagui/tamagui/issues/529


natew:
Managed to sneak this in already but thanks for the PR!

PR #556, opened 2023-01-20
[Issue] Popover broken for native screens with/without Adapt

thomasstevens89:
The popover demo is broken on native devices (admittedly only tried IOS) when not using an `<Adapt>` for a popover. When using adapt, the popover will correctly display a sheet but only for devices matching `when="$sm"`.

If you click the popover button on a `> $sm` device, nothing renders.


natew:
This is in the docs:

https://tamagui.dev/docs/components/popover

image

natew:
Maybe plans to support in the future but not high priority.

PR #557, opened 2023-01-20
[Issue] Popover and Sheet do not work on web with the css animation driver

thomasstevens89:
This is likely known by Nate but isn't documented anywhere.

Sheet (which Popover Adapt also depends on) assumes the animation driver can handle `Spring` animations. It may even only work with the `animations-react-native` package and not any others.


natew:
It doesn't assume spring btw, just assumes it's implemented, the driver does need some fleshing out.

natew:
Making an issue, PR is only for if you have a reproduction.

PR #566, opened 2023-01-23
feat(native): preferred color scheme

iconjone:
Per [discussion](https://discord.com/channels/909986013848412191/1066854123095474186), starter only had light support, and if dark theme was forced, there were visual errors. Changes allow for automatic light/dark theme preference.


natew:
Awesome, thanks

PR #567, opened 2023-01-23
[Issue] add missing "types" property to package.json "exports" maps

coffeebeats:
Currently, TypeScript fails to find declarations for `tamagui` (and many `@tamagui/*` packages) when using the `Node16` or `NodeNext` [module resolution strategy](https://www.typescriptlang.org/tsconfig#moduleResolution) (i.e. using ES modules). The reason for this is because all `package.json` files are missing the [`types` field within their `exports` maps](https://www.typescriptlang.org/docs/handbook/esm-node.html#packagejson-exports-imports-and-self-referencing); this is only necessary when declarations aren't located next to the JavaScript files.

This PR:

- adds the missing field to all relevant `package.json` files
- adds a missing `StyleObject` import to `packages/static/types/extractor/extractMediaStyle.d.ts` (I'm not sure why the `build` action adds this only after making the prior change)
- removes an invalid `./react-native` entry point from the `@tamagui/core` `exports` map (this file no longer exists)

NOTE: I'm happy to create an example repository which demonstrates the issue, if that helps.

### DIFF
--- a/packages/avatar/package.json
+++ b/packages/avatar/package.json
@@ -25,6 +25,7 @@
   "exports": {
     "./package.json": "./package.json",
     ".": {
+      "types": "./types/index.d.ts",
       "import": "./dist/esm/index.js",
       "require": "./dist/cjs/index.js"
     }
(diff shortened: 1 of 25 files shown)


natew:
Beautiful, thanks

PR #573, opened 2023-01-23
Idea: `unstyled` prop

nandorojo:
I feel that the styles/colors on many elements are too tailored to the Tamagui website. It takes a lot of guessing and checking to undo them and build them from scratch. Rather than implement a massive feature, I think an easy solution is an `unstyled` variant that "undoes" default styles set by pre-styled elements. 

This PR serves as a draft of how that might look.

A key element for me here is the colors. I don't know what each color corresponds to or where to edit them properly. I find that the `borderFocus` and `$color` and `$backgroundStrong` etc colors are meant for the Tamagui site, but I don't exactly know the right way to configure them for my own.

Thus, I prefer to only use the `$color1`-`$color12` scales from Radix, and to override the built-in `light` and `dark` themes from Tamagui which do not follow the Radix spec. 

I think a studio will also help in general for understanding the color themes better. But an `unstyled` prop would go a long way.

### DIFF
--- a/packages/button/src/Button.tsx
+++ b/packages/button/src/Button.tsx
@@ -99,6 +99,7 @@ export const ButtonFrame = styled(ThemeableStack, {
     },
       
     unstyled: {
+      // reset all styles
       true: {
         focusable: false,
         hoverTheme: false,
--- a/packages/button/src/Button.tsx
+++ b/packages/button/src/Button.tsx
@@ -97,6 +97,22 @@ export const ButtonFrame = styled(ThemeableStack, {
         pointerEvents: 'none',
       },
     },
+      
+    unstyled: {
+      focusable: false,
+      hoverTheme: false,
+      pressTheme: false,
+      backgrounded: false,
+      borderWidth: 0,
+      borderColor: 'transparent',
+      justifyContent: 'flex-start',
+      alignItems: 'flex-start',
+      flexWrap: 'nowrap',
+      flexDirection: 'column',
+      pressStyle: {},
+      focusStyle: {},
+      hoverStyle: {}
+    }
   } as const,
 
   defaultVariants: {
--- a/packages/button/src/Button.tsx
+++ b/packages/button/src/Button.tsx
@@ -99,19 +99,21 @@ export const ButtonFrame = styled(ThemeableStack, {
     },
       
     unstyled: {
-      focusable: false,
-      hoverTheme: false,
-      pressTheme: false,
-      backgrounded: false,
-      borderWidth: 0,
-      borderColor: 'transparent',
-      justifyContent: 'flex-start',
-      alignItems: 'flex-start',
-      flexWrap: 'nowrap',
-      flexDirection: 'column',
-      pressStyle: {},
-      focusStyle: {},
-      hoverStyle: {}
+      true: {
+        focusable: false,
+        hoverTheme: false,
+        pressTheme: false,
+        backgrounded: false,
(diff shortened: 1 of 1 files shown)


mwood23:
Nate and I have talked about something similar to this. One of the challenges I've had with Tamagui is there's two levels to the design system and I really want there to be a middle layer.

- Core (exists): Here’s style prop go have fun and we’ll codegen types+ handle theming
- Primitives: This is radix or theme-ui (this is the missing piece)
- Tamagui (exists): This is Chakra or Tailwind

A primitive could be:
- No style tokens
- No variants (including size)
- Only raw color values
- Business logic within a hook (things like Select, Accordion, or Dialog would get a lot here)

I think an unstyled prop could also get us there. I'd worry it'd become a maintenance burden over time compared to a clearer separation of concerns. Some components like `Dialog` are closes whereas components like `Select` force a lot theme props on you.

natew:
Ahh I see doing it with a variant is a good idea.

How does this compare to having `unset` just like how CSS does it? So `<Button unset />` turns off all styles?

I guess you want some styles to stick around?

Because `styled(Square, { unset: true })` almost doesn't make sense, may as well use `Stack`, but it does make sense in other areas like in a variant, or if it accepts arrays:

```tsx
const Sqaure2 = styled(ColorfulSquare, {
  variants: {
    unstyled: { true: { unset: true } }
  }
})

const ColorfulCircle = styled(ColorfulSquare, {
  unset: ['borderRadius']
})
```

mwood23:
The main issue I see for this approach is that the unstyled variant needs to cascade down. So for button, you'll also need it for `ButtonText` since there is a `$color` variable there. Similarly, components like `Select` will need the same treatment.

If you added this prop, I'm not sure what the consumption experience looks like because you'll still need to create a custom component based on the docs: https://tamagui.dev/docs/components/button/1.0.0#customization-advanced.

Another potential option is utilizing a factory component that bootstraps the unstyled version of all components. Then `tamagui` layers on top via the theme system. Chakra utilizes a pattern similar to this: https://chakra-ui.com/docs/styled-system/chakra-factory

nandorojo:
I'd pass `unstyled` to the `ButtonText` too, and compose it like `Button.Text = styled(ButtonText, { unstyled: true, ... })`. 

Then to consume it:

```tsx
<Button>
  <Button.Text />
</Button>
```

I think the simplicity of this `unstyled` option is nice, since the styles are all colocated. It doesn't imply "resetting all styles". Rather, "remove the special styles set by Tamagui".

natew:
Yea this makes sense. Debating if doing something more fancy like having a `@tamagui/button/unstyled` import that avoids even adding the styles, but not sure it's worth the cost of maintenance for the benefits of slightly smaller bundles.

natew:
Started on this, almost working for Button. Fixed some nice correctness stuff in variants.

natew:
Here https://github.com/tamagui/tamagui/pull/647

alitnk:
We're exploring a new headless API here #972, basically:
- `import {Button} from @tamagui/button  // or tamagui` is the default tamagui styled button, just like before
- `import {Button} from @tamagui/button/headless` is a headless button, no styles, no automatic sizing, very primitive

Also we're making buttons more composite, so there's `Button.Text` and `Button.Icon` in the new API. All backwards-compatible.

natew:
Yes headless is coming, plus unstyled still will exist for the styled versions. Closing as it's already partially complete.

PR #618, opened 2023-01-29
feat: checkbox component

### LINKED ISSUE #609: Checkbox
Almost identical to Radix:

https://www.radix-ui.com/docs/primitives/components/toggle-group

- But instead of <Checkbox.Root> it should be <Checkbox>.
- There should be a `native` prop on the root that on the web will just output to a plain old web checkbox.

Requirements:

- Add docs page to the site.

alitnk:
WIP

Closes #609


natew:
You may want to use `SizableStack` so we can support `size` prop, it'd be easy to test with `yarn sandbox` and set up a demo in packages/demos

alitnk:
> You may want to use `SizableStack` so we can support `size` prop, it'd be easy to test with `yarn sandbox` and set up a demo in packages/demos

Done - thanks for mentioning that.

alitnk:
I tried adding the Radix tests but couldn't set up Vitest and Tamagui to run them, unfortunately. Would be great if you could point me to some existing component tests so I can use them as a reference.

Would love to hear your feedback on the PR @natew

alitnk:
Also, I'm not sure what version I should go with for the mdx file's name. I went with [`1.2.0.mdx`](https://github.com/tamagui/tamagui/blob/91bb5629d71db6c4c3d2dba867c32ab2d0df9338/apps/site/data/docs/components/checkbox/1.2.0.mdx) assuming this will be released in the next minor version.

natew:
@alitnk that works, authorized you to deploy on vercel, will review this soon.

natew:
CleanShot 2023-01-30 at 19 23 08@2x

Seeing this on Safari.

alitnk:
@natew created an issue for that. It's an issue with icon components not dealing with size tokens properly.

natew:
Things looking pretty good overall, thanks for this. Testing with the new icon fix once it deploys.

natew:
B6B6044C-13C8-4144-9FF5-CBA210AB6E31

Icon size needs to scale there’s a pattern in button for this. 

And on mobile no focus style making the border darker

alitnk:
> You may want to use `SizableStack` so we can support `size` prop, it'd be easy to test with `yarn sandbox` and set up a demo in packages/demos

So the icon should be passed as a prop to the checkbox indicator, right?

natew:
Frame should take size and pass to indicator through context.

alitnk:
Alright so I played around with some approaches. I'm leaning towards the third one but would like to know your opinion first @natew.
### 1. Passing icon as children to `<Checkbox.Indicator />`
  - It's how Radix handles it
  - Got size forwarding with `asChild` to work _but_ it does not work when trying to handle the indeterminate state. here's an example

Works: 
```tsx
<Checkbox size={props.size} scaleIcon={2}>
  <Checkbox.Indicator asChild>
    <MinusIcon /> // is properly sized
  </Checkbox.Indicator>
</Checkbox>
```
Does not work:
```tsx
<Checkbox size={props.size} scaleIcon={2}>
  <Checkbox.Indicator asChild>
    {checked === 'indeterminate' && <MinusIcon />}
    {checked === true && <CheckIcon />}
  </Checkbox.Indicator>
</Checkbox>
```

### 2. Passing icon as prop to `<Checkbox.Indicator />`
- Kind of similar to how `<Button />` handles it

Would look something like:
```tsx
<Checkbox size={props.size} scaleIcon={2}>
  <Checkbox.Indicator indicatorIcon={checked === true ? <CheckIcon /> : checked === "indeterminate" ? <MinusIcon /> : null} />
</Checkbox>
```

### 3. Have an additional `<Checkbox.Icon />`
- No need for user to use `asChild` anymore
- Checked and indeterminate are supported
- Size is handled in the special checkbox icon component. gets the size from context.

Would look like this:
```tsx
<Checkbox size={props.size} scaleIcon={2}>
  <Checkbox.Indicator>
    {checked === 'indeterminate' && <Checkbox.Icon icon={<MinusIcon />} /> }
    {checked === true && <Checkbox.Icon icon={<CheckIcon />} /> }
  </Checkbox.Indicator>
</Checkbox>
```

alitnk:
Pushed fixes for mobile border focus and the icon pattern. @natew 

> I wonder why asChild isn't working here is it because it only expects on component?

Yes, `Slot` supports only one child. I'm sure we could hack a way around it though.

alitnk:
Cleaned up the PR.

natew:
Looks very close. I want to implement your proposal 1 though as that's the right API. Hopefully I can look this weekend and we can release. Something seems off with `asChild` (probably logic in Slot) if you do want to check.

natew:
Let's make the examples just show 3 size up to $4, and default the last one to be checked.

alitnk:
@natew made the first proposal work without the use of `asChild` - similar to [how `<Group />` handles its children](https://github.com/tamagui/tamagui/blob/1125dbdf93f6c56de1b30af08f0f9994ae0e7b12/packages/tamagui/src/views/Group.tsx#L73). thoughts?

natew:
Problem with mapping children is it doesn't allow HOC, if I do:

```tsx
// imagine it does some stuff to change props etc
const MyIndicator = (props) => <Chekbox.Indicator {...props} />
```

We don't know that it will pass the props down. In general it's something to avoid, I get away with it in Group but honestly I should make Group like this:

```tsx
<XGroup>
  <XGroup.Item />
</XGroup>
```

That way it really knows how to pass the props (to Item children) rather than doing a direct loop...

Couldn't size be passed to Indicator via context, and then Indicator can map over the children? That seems ok enough. And maybe a `disablePassSize` or something option.

natew:
Oh i see you do do the map in Indicator, that makes sense.

We could also have a `acceptsSize` static property optional on components so it knows to pass it or not, I think we do this already in some cases, see `isUnspaced`.

alitnk:
> We don't know that it will pass the props down. In general it's something to avoid, I get away with it in Group but honestly I should make Group like this:
> 
> ```tsx
> <XGroup>
>   <XGroup.Item />
> </XGroup>
> ```
> 
> That way it really knows how to pass the props (to Item children) rather than doing a direct loop...

This looks very similar to my third proposal.

> We could also have a acceptsSize static property optional on components so it knows to pass it or not, I think we do this already in some cases, see isUnspaced.

Tried wrapping the icon in `<Unspaced />` and it works (styles don't get passed).

Also, just added`disablePassStyles`. Named it differently since it passes color as well as it uses  `useGetThemedIcon()` but could rename it to `disablePassSize`.

natew:
Can we make the native prop work with this? https://css-tricks.com/almanac/properties/a/accent-color/

natew:
Almost there!

We need to actually size down things, it shouldn't be so big. Because size === Button height by default, and checkbox/radio should be smaller than a button. See the helper stepTokenUpOrDown and how it's used to do this, lets try sizing down 2 to start.

natew:
Why `useTheme({ name: 'active' })`? Since it's web-only we can likely just pass string `var(--background)` but if we do use useTheme() seems like we just use their current theme?

natew:
CleanShot 2023-02-06 at 19 13 17@2x

Size looks off, the $4 is huge, it should be the size of $2.

alitnk:
@natew thanks for the feedback. pushed fixes for both mentioned issues. 

> Size looks off, the $4 is huge, it should be the size of $2.

I agree it does look off. I tried to sync it with `<Switch />`'s size. setting $4 as the size of $2 makes $1 and $2 almost non-existent.

alitnk:
Scaled down the size of checkbox. Let me know if you think it needs more adjustments.

PR #626, opened 2023-01-30
fix: tamagui website issue on windows

### LINKED ISSUE #625: [bug] Tamagui site doesn't work properly on Windows
**Describe the bug**

The pages that use the [`getAllFrontmatter`](https://github.com/tamagui/tamagui/blob/c7fb6573f1cad7b5ff428c73697b521a997c5fa8/apps/site/lib/mdx.ts#L21) function don't work on windows. Must be an issue with Windows using backslashes instead of slashes. 

**To Reproduce**

Run `yarn site` on Windows and try navigating to the docs page; You'll get a 404 error.

I've tried the same on Linux and it works with no issues. This only happens on Windows.

alitnk:
certified Windows moment

`DATA_PATH` has backslashes on Windows and slashes on Linux. Tried the fix and it works both on Windows and Linux.

Closes #625

### DIFF
--- a/apps/site/lib/mdx.ts
+++ b/apps/site/lib/mdx.ts
@@ -27,7 +27,7 @@ export const getAllFrontmatter = (fromPath: string) => {
       const { data, content } = matter(source)
       return {
         ...data,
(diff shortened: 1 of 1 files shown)


natew:
ty

PR #628, opened 2023-01-30
feat(Accordion): Create Accordion Component

### LINKED ISSUE #604: Accordion
The API should closely mimic Radix:

- https://www.radix-ui.com/docs/primitives/components/accordion

Requirements:

- If no Tamagui library exists, either just avoid a library and write it yourself or make a `@tamagui/` library only if it helps other futuer components.
- Animations need to work across all the animation drivers.
- Add docs page to the site.

TheEhsanSarshar:
TODO
- [x]  Create a Collapsible Component
- [x]  Create a Collection Component
- [x]  Create an Accordion Component
- [x]  Make it accessible / performant
- [x]  Test on web/mobile platforms
- [x]  LGTM

closes #604


natew:
Fine to have collapsible as well here.

natew:
Unfortunately we run into the same problem as a few other components currently in progress, where native doesn't have an idea of querying the DOM to find the order.

What we could do.... is maybe have a wrapper that uses onLayout or something, calls back through context to the parent with it's reference and dimensions.

Then the parent can listen for all the dimensions, then compare them based on `flexDirection` to determine their order.

TheEhsanSarshar:
> Unfortunately we run into the same problem as a few other components currently in progress, where native doesn't have an idea of querying the DOM to find the order.
> 
> What we could do.... is maybe have a wrapper that uses onLayout or something, calls back through context to the parent with it's reference and dimensions.
> 
> Then the parent can listen for all the dimensions, then compare them based on `flexDirection` to determine their order.

do you want to elaborate more on this, actually, I didn't get what's the issue? why do we need to know the order of the elements?

natew:
@TheEhsanSarshar not sure actually but it's what Collection is doing with the querySelector parts. Maybe it's only for accessibility and we can leave it out on native.

TheEhsanSarshar:
@natew I guess you're right it's only for Accessibility, at least with Accordion

josephto:
Is there anything I can do to help move this along? Would love to be able to use a Tamagui Accordion component

TheEhsanSarshar:
> Is there anything I can do to help move this along? Would love to be able to use a Tamagui Accordion component

Today, I'll try to finish the remaining parts.
In meantime if you test it on native devices that would be great 😊

josephto:
I've been developing with the code from this PR copied into my local repo and it's been working beautifully in both iOS and Android emulators

natew:
@josephto nice! That's a great endorsement. @TheEhsanSarshar want me to fix up merge conflicts, is this ready for re-review?

TheEhsanSarshar:
@natew yeah please do that

TheEhsanSarshar:
It's ready for review

natew:
@TheEhsanSarshar could you add a docs page for it? Looking at it now.

natew:
Ok playing with it, lets get the demo to be closer to [radix](https://www.radix-ui.com/docs/primitives/components/accordion):

- Lets show animations (let me know if there's troubles there)
- Can we show the chevron/arrow indicator like the do
- The "focus" border on the headers when you click them


I made a few fixes on my branch `TheEhsanSarshar-create-accordion-component` you can merge back in here.

natew:
Also just saw this in Framer Motion:

https://www.framer.com/motion/layout-animations/

Notice LayoutGroup. Not sure if necessary here but seems relevant for Accordions. We don't have layout animations ofc but maybe still applies? Not sure.

TheEhsanSarshar:
thanks @natew for your review.
sure I'll check all those things that you said.

TheEhsanSarshar:
My only concern about animation part is that it should work with all the drivers that we have.
or maybe for layout animations we might need to have some sort of a reusable mechanism that we might later need for other components as well.

LeanderGangso:
Really looking forward to this component. Any estimate for when it's ready? 😁

TheEhsanSarshar:
because we have some layout animation ( height animation when the accordion expands or collapses). we aren't super confident to merge this PR because of some API changes that might happen later.

right now I'm working on a tiny layout animation library, hopefully, I can finalize it soon.

other parts of the Accordion are ready, if it's possible you can use this branch 🙂 
till we finalize that part

doodle-codes:
I would only implement animations in the accordion for web-use components. Still, for the native devices, I would remove the styling feature because React Native does not truly support animations unless you use a third-party library. I honestly believe that this will become too problematic to maintain. I understand that for user experience, it's a nice to have feature, but it's not a critical one.

ammarfaris:
Hi, I'm a bit confused, so technically Accordion can be used (via PR's code), but it will execute without collapsible animations? is that true? thanks @natew @TheEhsanSarshar and the rest of the team and contributors for making cross-platform styling easier.

TheEhsanSarshar:
@ammarfaris we'll try to merge that this week

packages/collapsible/src/Collapsible.tsx
```diff
@@ -1,248 +1,191 @@
...
+ * -----------------------------------------------------------------------------------------------*/
+
+const TRIGGER_NAME = 'CollapsibleTrigger'
+
+type CollapsibleTriggerElement = TamaguiElement
+interface CollapsibleTriggerProps extends StackProps {
+  children: ReactNode | ((props: { open: boolean }) => ReactNode)
+}
+
+const CollapsibleTriggerFrame = styled(ThemeableStack, {
```
natew, line 100: can we throw this into the `unstyled: false` and make unstyled generally work?

packages/demos/src/AccordionDemo.tsx
```diff
@@ -0,0 +1,58 @@
...
+    <Accordion overflow="hidden" width="$20" type="multiple">
+      <Accordion.Item value="a1">
+        <Accordion.Trigger
+          focusStyle={{
+            borderColor: '#fdfdfd',
+          }}
+          flexDirection="row"
+          justifyContent="space-between"
+        >
+          {({ open }) => (
```
natew, line 15: we don't use render callbacks anywhere else, but i'm not against them... the alternative would be a hook i guess? thoughts? @nderscore @nandorojo

natew: @alitnk

packages/demos/types/index.d.ts
```diff
@@ -1,48 +1,90 @@
...
-export * from './TokensDemo';
-export * from './ToastDemo';
-export * from './ToastDuplicateDemo';
-export * from './TabsDemo';
-export * from './TabsAdvancedDemo';
-export * from './ToggleGroupDemo';
-export * from './useOnIntersecting';
-//# sourceMappingURL=index.d.ts.map
\ No newline at end of file
+export { AlertDialogDemo } from './AlertDialogDemo'
```
natew, line 1: whats up with this change?

packages/collapsible/src/Collapsible.tsx
```diff
@@ -1,248 +1,191 @@
...
+  CollapsibleContentElement,
+  CollapsibleContentProps
+>((props: ScopedProps<CollapsibleContentProps>, forwardedRef) => {
+  const { forceMount, children, __scopeCollapsible, ...contentProps } = props
+  const context = useCollapsibleContext(CONTENT_NAME, __scopeCollapsible)
+
+  return (
+    <AnimatePresence {...contentProps}>
+      {forceMount || context.open ? (
+        <CollapsibleContentFrame {...contentProps}>{children}</CollapsibleContentFrame>
```
nderscore, line 172: It looks like the ID for linking the collapsible region to the trigger via `aria-controls` might be missing here?
```suggestion
        <CollapsibleContentFrame
          id={context.contentId}
          {...contentProps}
        >
          {children}
        </CollapsibleContentFrame>
```

packages/demos/src/AccordionDemo.tsx
nderscore: I'm not sure of a better way to expose internal state of the accordion to each collapsible region while still allowing consumers to write inline JSX.

I don't see why there couldn't also be a hook/context provider option as well though to support non-inline use cases.

alitnk: I also don't see a problem with having both render function and hook other than there being two ways to do the same thing.

Hook might need to take a value argument (`const { open } = useAccordionState('a1')`

packages/demos/types/index.d.ts
TheEhsanSarshar: auto generated ƪ(˘⌣˘)ʃ
let me check it again

packages/demos/src/AccordionDemo.tsx
TheEhsanSarshar: having both hook and render functions is a great idea
render function is especially helpful for performance

packages/collapsible/src/Collapsible.tsx
TheEhsanSarshar: nice catch

packages/demos/src/AccordionDemo.tsx
nandorojo: I like it. I wish the other components had this option too. I always want to close popovers etc without managing the state myself. I think it should always be optional, but it’s a nice way to access internal state.

packages/animate-presence/src/AnimatePresence.tsx
```diff
@@ -47,7 +47,7 @@ function onlyElements(children: ReactNode): ReactElement<any>[] {
   // We use forEach here instead of map as map mutates the component key by preprending `.$`
   Children.forEach(children, (child, index) => {
     if (isValidElement(child)) {
-      if (!child.key) {
+      if (!child.key && Children.count(children) > 1) {
```
nandorojo, line 50: Does this belong in this PR?

packages/demos/src/AccordionDemo.tsx
nandorojo: Could also add something like `.ItemExpanded` which only renders children when it’s open or something. But the render function is a useful way to animate the transition.

packages/collapsible/src/Collapsible.tsx
TheEhsanSarshar: actually `aria-control` comes from props.

packages/animate-presence/src/AnimatePresence.tsx
TheEhsanSarshar: while working with the accordion I faced this warning and fixed it here. 
yea maybe create another PR for that?

packages/demos/src/AccordionDemo.tsx
natew: sounds good overall, thanks @TheEhsanSarshar for quick work here

Issue #627, opened 2023-01-30
[bug] RN named colors on flattened/compiled components not working


thomasstevens89:
Works on native: `<XStack backgroundColor="$red5" />`
Does not work on native: `<XStack backgroundColor="red" />`

Web is unaffected.

natew:
Fixed 1.1.12

Issue #623, opened 2023-01-30
[bug] Document site has an ugly scrollbar in dark mode


Envov:
<img width="1529" alt="image" src="https://user-images.githubusercontent.com/48908684/215425183-9fbb7eaf-d27b-42a0-a410-045081f54972.png">
It should be something like github：
<img width="641" alt="image" src="https://user-images.githubusercontent.com/48908684/215425403-3a24e39e-d9ec-4c93-a92e-a871bf2c7e67.png">

natew:
Fixed 1.1.12

PR #629, opened 2023-01-30
fix(static): fix named css colors not loading ehen native when flatte…

natew:
…ning when using TAMAGUI_TARGET=native


natew:
Fixes https://github.com/tamagui/tamagui/issues/627

natew:
Fixes #623

PR #631, opened 2023-01-31
feat: Tabs & @tamagui/tabs component + separate @tamagui/group package

### LINKED ISSUE #630: Tabs
Close to Radix UI, but Tamagui utils and libraries.

https://www.radix-ui.com/docs/primitives/components/tabs

- Should support a background indicator that floats behind and can animate, like the vercel.com logged in toolbars at top
- Should support sizing/spacing like all other components do.
- Horizontal scroll option when overflows

alitnk:
Closes #630

### DIFF
--- a/apps/kitchen-sink/ios/.xcode.env.local
+++ b/apps/kitchen-sink/ios/.xcode.env.local
@@ -1,1 +1,0 @@
-export NODE_BINARY="/Users/n8/Library/Application Support/fnm/node-versions/v16.14.2/installation/bin/node"


alitnk:
@natew This still needs some docs work and refactoring, but I'd appreciate it if you could give me some feedback on its current state.
I've added a bunch of examples including:
- Tabs buttons w/ `XGroup`
- Animated tabs
- Underlined tabs
- Highlighted tabs (like Vercel's dashboard)

Here's a short video of the examples:

https://user-images.githubusercontent.com/35243344/218334371-600f6df9-95ee-4421-a311-dbadc93a2c36.mp4

natew:
The demo is amazing, just like what I'd expect. I can play with it a bit more. Will check out the code soon.

natew:
@alitnk can you fix up conflicts + yarn and I'll approve the deploy here

natew:
Yea let me see about that one, it's tricky so worth me seeing if there's some better solution.

alitnk:
- Separated `Group` into its own package (`@tamagui/group`) - not really related to this PR; I could separate it into another PR if needed.
- Added a `useGroupItem` hook for better handling some edge cases.
- `TabsList` and `TabsTrigger` are now extending `Group` and `GroupItem`, so the `scrollable` prop on `Group` is also applicable here. Overall it's more consistent and there's less boilerplate for groups.
- Removed the `RovingIndicator` component and added examples on how to achieve the same using the new callback props on `Trigger`. Reduces complexity on the tabs core and gives more control to the user for custom use cases.

@natew I think this is ready for a review.

natew:
Playing with it now and making some patches on a branch you can merge back here. Overall it seems great.

I can't seem to use shorthand properties on many of the components. EG Tabs.List, Tabs.Trigger.

natew:
The `onSelectedLayoutChange` and `onHoveredLayoutChange` does this just mean when it hovers it gives the current layout right? Seems alright, first impression was this is all a bit verbose with the many useState hooks.

Only thing I can think that would maybe work is:

```
onInteraction={(type, layout) => setState(prev => ({ ...prev, [type]: layout })}
```

That way you can have a single onInteraction passed to all three, and a single useState call? Not 100% but just an idea.

alitnk:
Continuing on #754

Issue #632, opened 2023-01-31
[bug] Passing size tokens for icons


alitnk:
Passing size tokens (`$1`, `$2`, etc.) on icons passes the token to the DOM

Usage: 
![image](https://user-images.githubusercontent.com/35243344/215680522-4dece1b4-5075-44dc-8e3d-1c470f82d4ea.png)

What actually ends up happening: 
![image](https://user-images.githubusercontent.com/35243344/215677072-64565a36-153f-4876-97b4-aa1a37bccd8d.png)

Related to the Safari issue in checkbox demo, https://github.com/tamagui/tamagui/issues/618#issuecomment-1409778378_

alitnk:
So I guess we should either fix the icon components to deal with size tokens or change the docs.

alitnk:
Closed in favor of bf954b1ad0d917ef2a1ff3676ae175c298058865

PR #634, opened 2023-01-31
fix(lucide-icons): size token not getting resolved

alitnk:
`getVariable()` was used to resolve a size token. Assuming that's the wrong function I went ahead and used `getSize()` from `@tamagui/get-size` instead.

Closes https://github.com/tamagui/tamagui/issues/632 and fixes the issue with Safari on the Checkbox PR (https://github.com/tamagui/tamagui/pull/618)

### DIFF
--- a/packages/lucide-icons/package.json
+++ b/packages/lucide-icons/package.json
@@ -21,7 +21,8 @@
   },
   "license": "MIT",
   "dependencies": {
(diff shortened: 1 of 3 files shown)


natew:
Thanks for checking into this, I implemented a bit diff just to keep it simple and improved getTokens a bit.

PR #636, opened 2023-01-31
feat(switch): add native prop

### LINKED ISSUE #613: [Feature] Add native prop to Switch
On iOS and Android they have native Switch elements. We can add:

```tsx
<Switch native>{/* ... *}</Switch>
```

Where on native apps it will render to the React Native Switch. Should be done inside Switch.native.tsx to avoid bundle size on web.

54m1r:
Just a draft for the native prop for the switch component like described in #613

hopefully i chose the right approach and understood the issue correctly, its's btw my first pr :)

Closes #613


packages/switch/src/Switch.tsx
```diff
@@ -189,6 +193,10 @@ export const Switch = withStaticProperties(
           }, [props.id, setChecked])
         }
 
+        if(native && (Platform.OS === 'ios' || Platform.OS === 'android')) {
```
natew, line 196: use `isNative` from `@tamagui/constants`

packages/switch/src/Switch.native.tsx
```diff
@@ -0,0 +1,25 @@
...
+  return (
+    <Switch
+      accessibilityRole="switch"
+      aria-checked={props.checked}
+      aria-labelledby={props.labeledBy}
+      aria-required={props.required}
+      data-disabled={props.disabled ? '' : undefined}
+      disabled={props.disabled}
+      value={isChecked}
+      ios_backgroundColor="#3e3e3e"
```
natew, line 18: Try out using the new `useStyle()` hook and taking in the color props from the parent so you can get the right colors.

natew:
@54m1r let me know if that makes sense! Thanks again for this I'll check it over more soon but lmk when you make those chnages.

54m1r:
I will check it today. I'm sorry, unfortunately I was too busy in the last days.

54m1r:
I am sorry to get back to you so late. After merging main into this branch the switch component doesn't work anymore because I think the types were destroyed and I couldn't fix the error

natew:
Closing, sorry :/, we'll get to this though somewhat soon.

PR #639, opened 2023-01-31
Docs(Core/Themes): Added (Light) post-fix

AlandSleman:
Added (Light) post-fix to the custom light theme.
And a few colors doesn't exist so i just commented them.

### DIFF
--- a/apps/site/data/docs/intro/themes.mdx
+++ b/apps/site/data/docs/intro/themes.mdx
@@ -275,17 +275,18 @@ import { tokens } from './tokens'
 
 const light = createTheme({
   background: '#fff',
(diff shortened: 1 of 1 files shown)


natew:
What's the thing about A5, A6? You can just delete the commented stuff. Technically you can do any typed theme so the docs just need to be clear on that - you can do you own and even a superset of the core ones we recommend.

You just have to change the types which should be doable now.

AlandSleman:
Oh my bad i was importing tokens from @tamagui/theme-base, Also i totally forgot you could create your own tokens.

AlandSleman:
So i think there is no need to change anything in the docs.

PR #642, opened 2023-02-01
feat: add RadioGroup component

### LINKED ISSUE #608: RadioGroup
Almost identical to Radix:

https://www.radix-ui.com/docs/primitives/components/radio-group

But: Instead of `<RadioGroup.Root>` it's just `<RadioGroup>`.

NOTE: This and ToggleGroup likely share the same underlying logic.

There's should be a `native` prop on the Indicator that would render to an actual `radio` on the web, on Native eventually it can render to something like the iOS style list of items with check marks but that can be later on.

Requirements:

- Add docs page to the site.

foyarash:
Add RadioGroup component

Fixes #608


natew:
Note there's two people just opened PRs back to back here. I don't have track of everything but I don't believe I said more than whoever puts up good work I'll assign to. Neither of the PRs are actually super far, since they mostly copy docs and the rough component API but don't have actual filled in functionality yet, but @azeez1776 seems just a bit further.

Not sure what the best strategy is, if you can reach out on Discord we can figure out.

natew:
Oh @foyarash I missed the file was showing "too big to diff" so missed that. Looks like you are pretty far along.

natew:
Looking better, there needs to be a focus outline like on the Checkbox PR.

Also, we need to actually size down things, it shouldn't be so big. Because size === Button height by default, and checkbox/radio should be smaller than a button. See the helper `stepTokenUpOrDown` and how it's used to do this, lets try sizing down 2.

natew:
Can we make the `native` prop work with this? https://css-tricks.com/almanac/properties/a/accent-color/

foyarash:
> Can we make the `native` prop work with this? https://css-tricks.com/almanac/properties/a/accent-color/

Would that be in an inline style ?

natew:
@foyarash yea inline style should work i think

natew:
@foyarash we're ready for this one, could you rebase and cleanup conflicts and if you can check over the checkbox make sure they are doing the same sizing logic. We can release this weekend if you get any time.

natew:
Made a few last changes on a new branch and merged it in there. Thanks for the great work here!

PR #644, opened 2023-02-02
Worked on the Radio group component assigned

### LINKED ISSUE #608: RadioGroup

azeez1776:
Closes #608 
Worked on creating a Radio Group component.
Worked on creating it's documentation.
Followed the practices used by other components.


natew:
Note there's two people just opened PRs back to back here. I don't have track of everything but I don't believe I said more than whoever puts up good work I'll assign to. Neither of the PRs are actually super far. Not sure what the best strategy is, if you can reach out on Discord we can figure out.

PR #646, opened 2023-02-03
fix: jsdoc wrong default value for button prop

alitnk:
1. it had wrong default value
2. it had two separate comments - I merged them


natew:
ty

PR #650, opened 2023-02-04
Feat/togglegroup

azeez1776:
:star: Feat: ToggleGroup Feature Closes #607 

- Added the ToggleGroup Component.
- Added Documentation for the component.
- No using Radix libraries.

Issue #607, opened 2023-01-28
ToggleGroup


natew:
Almost identical to Radix:

https://www.radix-ui.com/docs/primitives/components/toggle-group

But instead of `<ToggleGroup.Root>` it should be `<ToggleGroup>`.

Requirements:

- No using Radix libraries, Tamagui has equivalent libraries for most.
- Add docs page to the site.
- Simple tests would be appreciated.

glocore:
I'd like to work on this one!

natew:
@glocore sounds good, open a draft PR that references this and we can consider it yours.

natew:
With some initial real progress

azeez1776:
Hey @natew can I work on this.

natew:
@azeez1776 go for it I’ll reserve for you for next 48 hours. So long as you show decent progress within then it will be yours.

azeez1776:
Thanks @natew , got it

PR #655, opened 2023-02-05
feat: roving focus package

alitnk:
#608 (PR: #642), #630 (PR: #631) and #607 (PR: #650) need this for keyboard navigation


alitnk:
Sorry for the Vercel ping Ehsan. I noticed you've already ported `collection` so I just cherry picked your commit.

alitnk:
@natew I think this is ready.

natew:
I see `radix` in the code.

Not the most happy with the size of this, feels like collection is very abstraction-heavy. I think that and not sure what Slot is doing can't we just pass props? But I can merge this if you feel those are worth the cost.

natew:
Yea just fix up conflicts and remove radix, I think it's ok. It doesn't seem to have any carve-out for native doesn't native not need this?

alitnk:
> Not the most happy with the size of this, feels like collection is very abstraction-heavy. I think that and not sure what Slot is doing can't we just pass props? But I can merge this if you feel those are worth the cost.

I also agree it could probably be simpler. We can just change collection later as we don't use it directly (except for the accordion PR)

> It doesn't seem to have any carve-out for native doesn't native not need this?

The purpose of this package is focus management of groups by left-right or up-down buttons (you can see it in use in #631) - I don't know if there's an equivalent for that on native. But if you have an idea please let me know.

packages/roving-focus/types/RovingFocusGroup.d.ts
```diff
@@ -0,0 +1,48 @@
+import { Stack } from '@tamagui/core';
+import * as React from 'react';
+/**
+ * A custom hook that converts a callback to a ref to avoid triggering re-renders when passed as a
+ * prop or avoid re-executing effects when passed as a dependency
+ */
+declare function useCallbackRef<T extends (...args: any[]) => any>(callback: T | undefined): T;
```
amaranth, line 7: Isn't this basically the `useEvent` already defined in `@tamagui/use-event`?

alitnk: Replaced it with `useEvent`. Thanks for pointing out.

natew:
Lets make `.native.ts` versions of these that exports empty functions.

alitnk:
> Lets make `.native.ts` versions of these that exports empty functions.

@natew when porting components to tamagui, there are use cases like this:
``` 
<RovingFocusGroup asChild ...>
```
The current implementation allows these to work. I guess we could use a Slot instead of the Stack there.

We could also just return children, but it wouldn't be a 1-1 to web implementation anymore and we should probably only use it on `.web.ts` files at that point. It's your call though.

natew:
I think it's fine as is then, I'll merge this unless you have any objection.

Issue #599, opened 2023-01-27
[bug] Pass onPress to a component that wraps an extractable component results in double onPress call


foyarash:
# Reproduction

The issue can be reproduced here https://github.com/foyarash/tamagui-weather-demo/tree/119d30a238c7c7920ac97947b17f6ce9020cb42a

There is a button "Hello" on the home page, clicking on it results in a double `press` log.

https://github.com/foyarash/tamagui-weather-demo/blob/119d30a238c7c7920ac97947b17f6ce9020cb42a/src/components/Button.tsx#L6 Replacing `Box` by `BoxFrame` at this line will get around the issue.

# Expected result

Calling `onPress` on any component should result on a single `onPress` call upon a single press event

# Note

`onPressIn` and `onPressOut` events are not affected

lobor:
I try on your repo, very strange, can you try by replace @tamagui/core by tamagui package ?
also, if you want use theme with button, you can remove Theme component
```typescript
<ButtonFrame theme="primary_Button" size="$md" {...props}>
```

lobor:
Today i had this bug with not extractable component, is not resolved

lobor:
should be closed when #656 released

PR #656, opened 2023-02-05
fix(core): onClick exist only on web #599

lobor:
See issue #599


lobor:
It might be better to create an onClick prop directly

```typescript
onClick: attachPress
  ? (e) => {
    unPress()
    onClick?.(e)
  }
  : undefined,
onPress: attachPress
  ? (e) => {
      unPress()
      onPress?.(e)
    }
  : undefined,
```

Issue #498, opened 2023-01-04
[bug] Web - Click on Switch component submit form


lobor:
Hi

I use switch in form for cross mobile/browser component. The bug appear on browser (web).
When use switch in form, this click submit form
```typescript
import { Form, Switch } from "tamagui";

export const Demo = () => {
  return (
    <Form
      onSubmit={(e) => {
        e.stopPropagation();
        console.log("submited");
      }}
    >
      <Switch />
    </Form>
  );
};
```
In the browser, all buttons that have no `type` attribute submit the form.
The simple fix for web is add `type="button"` on all button by default and leave to choice with props to modify this attribute. `type="submit"` for submit form

example:
```typescript
import { Form, Switch, Button } from "tamagui";

export const Demo = () => {
  return (
    <Form
      onSubmit={(e) => {
        e.stopPropagation();
        console.log("submited");
      }}
    >
      <Switch />
      <Button type="submit">Submit</Button>
    </Form>
  );
};
```

Also, this example does not work on mobile, because `type="submit"` not exists.
I have to add handleSubmit to button and onSubmit to form (web)  for work on cross platform
example:
```typescript
import { Form, Switch, Button } from "tamagui";

export const Demo = () => {
  const handleSubmit = () => {
    console.log("Hello i'm lobor");
  }

  return (
    <Form onSubmit={handleSubmit}>
      <Switch />
      <Button type="submit" onPress={handleSubmit}>Submit</Button>
    </Form>
  );
};
```

@natew What do you think ? 

I love Tamagui, it's a great cross platform library. it's the only one I've found with a really smart theme system, thought about our needs.

natew:
Form isn't documented / released for this reason, leaving open.

natew:
I do like adding the `type` property though, will do that soon.

lobor:
May be you can add Component `Form.Trigger` for onSubmit event

lobor:
Hi @natew 
I have create FormTrigger component in my ui packages, i try to include in repo and create PR but i can run sandbox

so here this code, can you give me feedback please ?

```typescript
import {
  GetProps,
  Stack,
  TamaguiElement,
  styled,
  withStaticProperties,
  Scope,
  createContextScope,
  composeEventHandlers,
  YStack,
  YStackProps,
} from "tamagui";
import { forwardRef } from "react";

const FORM_NAME = "Form";

export const FormFrame = styled(Stack, {
  name: "Form",
  tag: "form",
});

[...]

PR #657, opened 2023-02-05
feat(Form): create @tamagui/form and add Form.Trigger #498

lobor:
see #498


natew:
Looks like a great start that we can iterate on. Docs page subtitle needs fixing.

Wondering how we should work with `type="submit"`. Typically on web a Button with type="submit" will send the onSubmit event. It's possible we could make this work but we'd have to hook into Button, but that wouldn't be too bad. But may need Yet Another New Package :)

Just figure since we're here may as well do that.

There's also the BubbleInput a variety of inputs use, not sure if we want to do anything there.

lobor:
ok, I'll edit that for the documentation.

For the web it might be necessary to have a condition in order to use type="submit" automatically, and use the context on the native. The problem with submit is that any other button with submit can submit the form, my dev's idea was to block that.

What exactly is BubbleInput?

natew:
Something that Radix does https://github.com/tamagui/tamagui/blob/9658f5e3e1964d2bcee3494a25f33e5f0c142619/packages/switch/src/Switch.tsx#L230, web only thing, from what I can tell it just sends a normal input check event so I guess this should just work?

natew:
If you can fix conflicts this should be good for release :)

lobor:
it's fix

lobor:
i fixed conflitct

lobor:
@natew i resolve conflict
When do you think that PR will be merged ?

natew:
Made some cleanups and merged!

natew:
Thanks for this 🙏

PR #664, opened 2023-02-06
fix(select): add missing dependency on sheet

iBotPeaches:
While ingesting select on its own in our design kit. It crashed out missing a reference to sheet, which is used in component here: https://github.com/tamagui/tamagui/blob/eaeff66a1f8cd6e30c2b1e70c51db9994a1f74f0/packages/select/src/Select.tsx#L14

Had some issues making this PR as 1.2.6 is out, but not pushed into master. So I'm pointing this to "next", which appears to be released already.


natew:
ty

PR #665, opened 2023-02-06
fix: native double render

nandorojo:
There's no reason to double render here on web


natew:
I think this looks right, because this is just giving us SSR safe... initial render is always at `initialValue` (which ideally should be user-configurable) and then after hydration it'll update to the real value to avoid hydration mis-matches?

natew:
Oh right I see the != now, thanks.

PR #667, opened 2023-02-06
docs(site) - community - add @criszz77/luna to starter repos section

criszz77:
This PR adds [🌘 Luna template](https://github.com/criszz77/luna) to the starter repos on the community page. 

It's basically a starter template with vanilla React Native, Next.js and Tamagui, but no monorepo -  everything is in one app.

Would be very happy if my work helps other people.


natew:
Thanks! I forgot to get around to that my bad.

PR #668, opened 2023-02-06
withTM Update

goatbarn:
Just an update on the docs regarding fixes for "Cannot use Import statement outside a module" when using libraries with native code and installing in the app directory.


natew:
🙏

Issue #619, opened 2023-01-29
[bug] Vite SSR Can't Resolve ESM


tazsingh:
Hey Nate,

Thanks for the other Vite fixes in #479! It's now gotten me past those errors and onto this one.

I've managed to widdle it down to the following repo, where you can see the patch that worked for me for that 1st error and onto the next with essentially the same pattern: https://github.com/tazsingh/tamagui-ssr-vite

Note, the patch doesn't include changing this line:

https://github.com/tamagui/tamagui/blob/master/packages/normalize-css-color/package.json#L31

to: `"import": "./dist/esm/index.mjs",`

Essentially, I believe that Vite CLI is a bit more forgiving than Node's module resolution. Annoyingly, Node seems to want everything to be `.mjs` files for ESM, while Vite doesn't seem to mind as much. Since I'm trying to load the project via Vite for SSR, it's hitting Node's module resolution which is constantly tripping over the `.js` files not being `.mjs` files.

Additionally, it's odd to me that `"type": "module"` doesn't seem to matter much here, and separately removing the `"exports"` map entirely from `package.json` is also a fix 🤷 Thought I'd mention it in case you're debugging. The patch above is what I've discovered to be the path-of-least-resistance to change all ESM output to use `.mjs` files. Hope that helps 🙏 

**To Reproduce**

Check out my repo

`yarn install`

`yarn dev:ssr`

Notice the next bug, that isn't fixed by the patch above. Feel free to remove the patch to see the issue it's patching over.

natew:
Woof, change to mjs is almost not doable for the hassle. All the module stuff has been a pain to get it working backwards compat.

tazsingh:
This is only for the ESM build, which has always used `.mjs` and is backwards compatible to Node 14 but goes all the way back to Node 8: https://nodejs.org/docs/latest-v14.x/api/packages.html#packages_determining_module_system

The alternatives which aren't backwards compatible would be changing the `package.json` `"type"`, as that would affect ESM and CJS. However, per my testing above, that doesn't seem to matter anyway

I agree the whole ESM thing is a hassle though! 😅

JoviDeCroock:
Hey Nate,

This is indeed a troublesome issue, we have had many issues with `urql` as well. The gist of it comes down to that plain `node` usage demands `.mjs` for anything that uses `ESM` type import/export. As Taz mentions, there are ways around this by marking your package as `type: 'module'` but then the `cjs` usage requires a `.cjs` extension.

[...]

PR #669, opened 2023-02-06
fix(build): provide both .mjs and .js

JoviDeCroock:
In this PR we offer both a `.mjs` as well as a js build, this so that we can put mjs in the export-maps so we can adhere to strict node resolution for all node-versions which need mjs to show them that we are using import/export. The reason for also supplying a js here is because tamagui uses dist/esm in the module property. This will make it so that older webpack/.. versions can still automatically pick what they can resolve according to their configuration.

Resolves https://github.com/tamagui/tamagui/issues/619
CC @tazsingh

### DIFF
--- a/packages/avatar/package.json
+++ b/packages/avatar/package.json
@@ -26,7 +26,7 @@
     "./package.json": "./package.json",
     ".": {
       "types": "./types/index.d.ts",
-      "import": "./dist/esm/index.js",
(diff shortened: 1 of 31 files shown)


nderscore:
@natew `@tamagui/core/reactNativeTypes` should probably get a `.mjs` extension in the export paths as well:

https://github.com/tamagui/tamagui/blob/master/packages/core/package.json#L62

Issue #638, opened 2023-01-31
[bug] expo-camera breaks next app


patcito:
How to reproduce:
- yarn add expo-camera
- import {Camera} from "expo-camera"
- add "expo-camera" to transpiled modules in next.config.js
- use the` <Camera />` tag in a screen
- refresh screen manually in browser

The error I get is: "SyntaxError: Cannot use import statement outside a module". Navigating to screen works but when refreshing or on first load I get this error.

Note, this works in solito.dev vanilla starter without tamagui.

Here is the repo to reproduce from starter https://github.com/patcito/tamagui-next-expo-camera

dohomi:
@patcito make sure that `expo-camera` is part of transpiled modules in your next.config.js

goatbarn:
I'm having the same issue after adding expo packages to setup my auth provider (expo-auth-session, expo-random, and expo-web-browser). All added to next config, just like OP.

Also, OPs repo DOES have expo-camera listed in transpileModules.. it's listed first...unless I'm crazy (or he edited?).

Is there anyway to further debug this message...other than looking at which library shows up in the stack?

patcito:
@dohomi as noted by @goatbarn, I've already added "expo-camera" to next.config.js here https://github.com/patcito/tamagui-next-expo-camera/blob/main/apps/next/next.config.js#L78

dohomi:
oh sorry I overlooked the first line (I only add new modules to the bottom and not top..). I have similar problems with expo modules like expo-av and expo-document-picker. What helped in my case was using [`transpile-module`](https://github.com/martpie/next-transpile-modules) instead of the built in feature. There is actually a discussion about it in Discord where I mentioned this. Most probably that will fix and can be used as a workaround

natew:
Lets use the discord for things that aren't really tamagui related

felicio:
Repost:

> I was able to make it work by removing tamagui-extras which I wasn't using, I may use it in the future though so would be nice if you find a fix 🙂
> – https://discord.com/channels/909986013848412191/1070796779555147856/1071027668864217088

dohomi:
@felicio open an issue at discord or https://github.com/dohomi/tamagui-kitchen-sink a lot has changed since february

felicio:
> @felicio open an issue at discord or https://github.com/dohomi/tamagui-kitchen-sink a lot has changed since february

[...]

PR #676, opened 2023-02-07
Further withTM revisions, and errant import/require typo fix

goatbarn:


natew:
Is transpileModules broken or something in next?

natew:
Or is this for next < 13.1? If so we should note the two options and the starter should keep the modern one.

starters/next-expo-solito/apps/next/next.config.js
```diff
@@ -2,6 +2,7 @@
 const { withTamagui } = require('@tamagui/next-plugin')
 const withImages = require('next-images')
 const { join } = require('path')
+const withTM = require('next-transpile-modules')
```
theonetheycallneo, line 5: @goatbarn This is deprecated, you can simply pass it directly into next config now:

https://nextjs.org/blog/next-13-1#built-in-module-transpilation-stable

theonetheycallneo (changes requested):

natew:
Yea I think this is not advisable anymore, I changed the readme back to use the new option.

starters/next-expo-solito/apps/next/next.config.js
goatbarn: See [https://github.com/tamagui/tamagui/issues/638#issue-1564910300](url). Many Expo modules (possibly since its SDK 47?) do not work without this.

What's the correct path forward then?

goatbarn: > Or is this for next < 13.1? If so we should note the two options and the starter should keep the modern one.

I'm all for keeping it within next realm but think it is worth noting at least. I'll go check and see if this is an issue noted with next for expo modules in particular and if its tracked then yeah, certainly just a note would do. Until they fix it, if so.

PR #683, opened 2023-02-08
docs: remove invalid link reference

alitnk:
The link leads to nowhere. I couldn't find any equivalent to [the old file that has been removed](https://github.com/tamagui/tamagui/blob/c3beeff4902388a420549abc608f1af7607658c9/packages/site/constants/themes.ts) in the site app.


natew:
Yea I can bring this back shortly with the much better examples coming up.

PR #702, opened 2023-02-13
feat: add disabledStyle prop

foyarash:
- [x] Add `disabledStyle` prop
- [x] Handle `disabledStyle` for web with CSS compiler
- [x] Handle `disabledStyle` on native
- [x] Add documentation for the prop


natew:
@foyarash this seems like a decent addition. Trying to think about it a bit more. I know we could use themes here and just have a disabled theme it looks for.... but then again this is very simple.

natew:
landed this!

PR #704, opened 2023-02-14
fix: pass hitSlop prop to view props

foyarash:
`hitSlop` props was indeed used in the `usePressability` hook, but it also needs to be passed as a prop to the created View.

This PR fixes this


natew:
amazing, thank you

Issue #701, opened 2023-02-13
`styled(Button)` incompatible with `<Group />`'s borderRadius


alitnk:
Button works just fine in `XGroup` and `YGroup`, but when using `styled(Button)` it fails to apply the border radius and it just uses the button's border radius.

It's important to note that this hasn't happened in sandbox (uses Vite) but happened in the documentation app, the `site` app (uses Next.js) - guessing maybe something with the ordering of classes might be involved?

![image](https://user-images.githubusercontent.com/35243344/218512558-95ea10a6-dbcd-46c4-bc23-684fccd11aba.png)
![image](https://user-images.githubusercontent.com/35243344/218512572-fefc63b6-2178-48b2-b2ca-7841f2b5261c.png)

idrakimuhamad:
I'm getting this too. In fact, it seems like using `styled` to extend the button broke some of it internal props.

```jsx
const CustomButton = styled(Button)

<YStack space bg="#dedede" p="$5">
  <Text>Styled Button</Text>
  <XGroup size="$8">
    <XGroup.Item>
      <CustomButton backgroundColor="green">Hello</CustomButton>
    </XGroup.Item>
    <XGroup.Item>
      <CustomButton backgroundColor="green">Hello</CustomButton>
    </XGroup.Item>
    <XGroup.Item>
      <CustomButton backgroundColor="green">Hello</CustomButton>
    </XGroup.Item>
  </XGroup>
  <Text>Tamagui Button</Text>
  <XGroup size="$8">
    <XGroup.Item>
      <Button backgroundColor="green" color="azure">
        Hello
      </Button>
    </XGroup.Item>
    <XGroup.Item>
      <Button backgroundColor="green" color="azure">
        Hello
      </Button>
    </XGroup.Item>
    <XGroup.Item>
      <Button backgroundColor="green" color="azure">
        Hello
      </Button>
    </XGroup.Item>
  </XGroup>
</YStack>
```
![Simulator Screen Shot - iPhone 14 - 2023-03-31 at 10 05 18](https://user-images.githubusercontent.com/1400323/229004401-d3bb45d4-33f3-4420-9b3b-5993b011b3e5.png)

Some prop doesn't do anything on the button itself, like `size` and `background`

```jsx
<YStack space bg="#dedede" p="$5">
  <Text>Styled Button</Text>
  <YGroup>
    <YGroup.Item>
      <CustomButton backgroundColor="green" size="$2">
        Should have size 2
      </CustomButton>
[...]

PR #705, opened 2023-02-14
feat(group): new API using `reforest`

alitnk:


alitnk:
This works but it has a regression when it's used with a `separator` (e.g. `<Group separator={<Separator/>} ...>`). Will need to find an alternative to `spacedChildren()` or rewrite it to work with `reforest`.

natew:
@alitnk if you can figure out the separator stuff would be nice. Can we make this backwards compat? Or should we move to a new component like `Group` instead?

alitnk:
Fixed the separator. Not sure how backward compatibility would work... maybe detect if children are `<XGroup.Item`s and make it behave differently and otherwise default to the old impl?

alitnk:
Thinking out loud here, I can think of a few options:
- Check if the first child is <XGroup.Item and use the new implementation if that's the case
- A prop to opt-in to the new implementation for now and gradually remove the old one and just move to that in v2 maybe... like `<XGroup itemChildren><XGroup.Item>...</XGroup.Item></XGroup>`
- Create a separate Group component, so no backward compatibility.
- Apply the old impl styles only on non-`XGroup.Item` components. will have to try this one out...

alitnk:
Last one seems to work, will push it in a bit.

alitnk:
Still, this doesn't seem to fix #701 - border radius issue with custom button still happening in the `site` app but not in `sandbox`.

alitnk:
Found a regression - working on it...

natew:
Wrote in discord, seeing SSR doubling:

https://user-images.githubusercontent.com/12100/221380605-cc842567-d7bf-4824-bdf6-c9a221893df3.mp4

also it doesn't seem to fall back to the old behavior if you don't use .Item, we could do that by having .Item look for context and call a callback to say "i'm mounted" and then the parent can switch the mode

PR #724, opened 2023-02-20
feat: Toast

### LINKED ISSUE #605: Toast
The component API should somewhat look like Radix:

https://www.radix-ui.com/docs/primitives/components/toast

- Keyboard shortcuts should work like Radix

Some important differences:

- There should be an imperative helper to show/hide:

```tsx
import { createToast } from '@tamagui/toast'

const { toast, useToasts } = createToast({
  native: true,
  // can set some default values:
  timeout: 10,
})

toast.show({ title: ``, message: ``, icon: ``, type: ``, timeout: 2 })
toast.hide()
toast.hide(1)
toast.hide('all')
```

Note the `native` property:

- Can be set to true/false, or `['web', 'ios']` or `['native']` for example. If set like this it will try and use the platform-native toasts.
- For web this is an actual notification

---

For the `useToast`:

```tsx
const toasts = useToasts() // returns most recent on top

const elements = toasts.map(toast => {
   return (
     <Toast>
        <Toast.Title>{toast.title}</Toast.Title>
        <Toast.Description>{toast.description}</Toast.Description>
     </Toast>
   )
})

return (
  <Toast.Viewport anchor="top-center">
    {elements}
  </Toast.ViewPort>
)
```

---

Docs:

The end user should be able to use enterStyle/exitStyle and <AnimatePresence> to do animate things so showing in docs would be great.

---

Requirements:

- Animations need to work across all the animation drivers.
- Add docs page to the site.

alitnk:
Closes #605


natew:
https://github.com/emilkowalski/sonner Has very similar API to ours but some other nice things too.

natew:
The stacked toasts btw is the demo we should show will be really cool.

packages/portal/types/GorhomPortal.d.ts
```diff
@@ -59,8 +59,14 @@ export interface PortalHostProps {
      */
     name: string;
     forwardProps?: Record<string, any>;
+    /**
+     * Useful when trying to animate children with AnimatePresence.
+     *
+     * Not a part of gorhom/react-native-portal
+     */
+    render?: (children: React.ReactNode) => React.ReactElement;
```
alitnk, line 67: Did this since can't simply wrap toasts in `AnimatePresence` as they get forwarded to the portal host. Do let me know if you can think of a better way to do this, @natew.

alitnk:
Closing in favor of #810

PR #731, opened 2023-02-23
docs: Mini-fix website starter guide

kaiiserni:
Per the website docs one needs to run "yarn dev" which is not an available script. Replaced with "yarn web" and "yarn native" to avoid confusion.


natew:
ty!

Issue #601, opened 2023-01-28
[bug] sheets gets blank on input focus (native expo)


patcito:
Happens on native android at least. Focusing on input text inside a sheet dialog makes the whole sheet go blank.

**To Reproduce**
Just paste the example from the doc https://tamagui.dev/docs/components/dialog/1.0.0
Run on expo android for example, open the dialog and press on the Name input
Sheet goes blank

Demo:
https://user-images.githubusercontent.com/26435/215271937-b10db209-b846-417f-b7e3-b46b7718d7e8.mp4

**System Info:** 

Output of `npx envinfo --system --npmPackages --binaries --browsers`
System:                                                                                                                                               
    OS: Linux 5.15 Ubuntu 22.04.1 LTS 22.04.1 LTS (Jammy Jellyfish)                                                                                     
    CPU: (8) x64 11th Gen Intel(R) Core(TM) i7-1185G7 @ 3.00GHz                                                                                         
    Memory: 1.63 GB / 31.06 GB                                                                                                                          
    Container: Yes                                                                                                                                      
    Shell: 5.1.16 - /bin/bash                                                                                                                           
  Binaries:                                                                                                                                             
    Node: 19.3.0 - ~/.nvm/versions/node/v19.3.0/bin/node                                                                                                
    Yarn: 3.2.3 - ~/.nvm/versions/node/v19.3.0/bin/yarn                                                                                                 
    npm: 9.2.0 - ~/.nvm/versions/node/v19.3.0/bin/npm                                                                                                   
  Browsers:                                                                                                                                             
    Brave Browser: beta
    Chrome: 108.0.5359.124
    Firefox: 109.0
  npmPackages:
   @babel/plugin-proposal-export-namespace-from: ^7.18.9 => 7.18.9 
    @babel/runtime: ^7.18.9 => 7.20.7  
    @ethersproject/providers: ^5.7.2 => 5.7.2 
    @ethersproject/shims: ^5.7.0 => 5.7.0 
    @expo/vector-icons: ^13.0.0 => 13.0.0 
    @expo/webpack-config: ^0.17.2 => 0.17.4 
    @gnosis.pm/safe-modules-deployments: ^1.0.0 => 1.0.0 
    @gnosis.pm/safe-react-gateway-sdk: ^3.5.2 => 3.5.2 
    @manypkg/cli: ^0.19.1 => 0.19.2 
    @react-native-async-storage/async-storage: ^1.17.11 => 1.17.11 
[...]

PR #749, opened 2023-03-02
fix(sheet): size frame rerender when keyboard showing

lobor:
Related to #601 

When the keyboard appear, the size frame is set like height - keyboardHeight, which causes onBlur event on input, and hide keayboard.

In my solution, when the keyboard is show, not trigger resizing.
I have may be a bug, when keyboard is open and focus on input, i can't press any way for onBlur input and hide keyboard (test on android), i should click on keyboard's hide button


natew:
Thanks 🙏

PR #751, opened 2023-03-02
Docs: Install react-native-reanimated with expo cli

G-Ray:
`react-native-reanimated` v3.0.0 has been released two days ago.
This version is not yet compatible with expo.

To be safe, I think we should use `expo install react-native-reanimated` to always install a compatible version of this library.


natew:
Nice idea, thanks.

PR #753, opened 2023-03-03
Lobor fix601

natew:


packages/use-keyboard-visible/src/useKeyboardVisible.ts
```diff
@@ -0,0 +1,22 @@
+import { useEffect, useState } from 'react'
+import { Keyboard } from 'react-native'
+
+export const useKeyboardVisible = () => {
+  const [isKeyboardVisible, setKeyboardVisible] = useState(false)
```
nandorojo, line 5: thoughts on using a ref here?

PR #754, opened 2023-03-04
Alitnk tabs 630

natew:
Small fixes for #631


alitnk:
@natew 

> The onSelectedLayoutChange and onHoveredLayoutChange does this just mean when it hovers it gives the current layout right? Seems alright, first impression was this is all a bit verbose with the many useState hooks.
> 
> Only thing I can think that would maybe work is:
> 
> ```
> onInteraction={(type, layout) => setState(prev => ({ ...prev, [type]: layout })}
> ```
> That way you can have a single onInteraction passed to all three, and a single useState call? Not 100% but just an idea.

Definitely agree having three event handlers is verbose. Removed them and added `onInteraction`  as suggested, but having 1 useState is not possible, unless we store both states in the same state.

> I can't seem to use shorthand properties on many of the components. EG Tabs.List, Tabs.Trigger.

I'm a bit lost on why this is happening... Can you point me to a direction, maybe?

PR #755, opened 2023-03-04
Tabs 630

alitnk:
Well, #631 can't be re-opened (oops) and I don't have write access to #754 - so continuing here.


alitnk:
@natew 

> The onSelectedLayoutChange and onHoveredLayoutChange does this just mean when it hovers it gives the current layout right? Seems alright, first impression was this is all a bit verbose with the many useState hooks.
> 
> Only thing I can think that would maybe work is:
> 
> ```
> onInteraction={(type, layout) => setState(prev => ({ ...prev, [type]: layout })}
> ```
> That way you can have a single onInteraction passed to all three, and a single useState call? Not 100% but just an idea.

Definitely agree having three event handlers is verbose. Removed them and added `onInteraction`  as suggested, but having 1 useState is not possible, unless we store both states in the same state.

> I can't seem to use shorthand properties on many of the components. EG Tabs.List, Tabs.Trigger.

I'm a bit lost on why this is happening... Can you point me to a direction, maybe?

natew:
Seeing regression here, I'll try and fix this up...

https://user-images.githubusercontent.com/12100/222930650-23503a32-4266-4e2e-8720-908158557da9.mp4

natew:
For some reason I don't see that bug locally even running things in prod mode...

I'm making more changes on a new branch. Idk how to Github, it doesn't let me push here for some reason. Going to search and see what's the right way to do this..

natew:
For some reason it's complaining I can't use shorthands on the Trigger itself.

One thing also, idk about using Button directly... It's pretty aggressive to force on people who may want to customize things a lot. But idk how hard it is to change. And it does look good by default.

CleanShot 2023-03-04 at 12 39 44@2x

natew:
https://github.com/tamagui/tamagui/pull/758

PR #756, opened 2023-03-04
Fix react-native-svg elements casing

prichodko:
The correct casing of gradients is capitalized – [linearGradient](https://developer.mozilla.org/en-US/docs/Web/SVG/Element/linearGradient) and [radialGradient](https://developer.mozilla.org/en-US/docs/Web/SVG/Element/radialGradient).

Current casing makes browsers ignore the elements.


natew:
Ah good catch, thank you.

PR #765, opened 2023-03-07
[issue] style inconsistency in production

ebg1223:
To reproduce:
cd starters/next-expo-solito
yarn
yarn build && yarn web:prod && yarn web:prod:serve

Open the page. Click the button on home screen and you are redirected to a page with 3 buttons, but at the top of the window. 

Refresh the page, and it renders correctly in the center of the window.

See screenshots folder in the starter for example.


ebg1223:
I believe it has something to do with alignItems={"center"} on the YStack on the homescreen. Without setting this, rendering is normal.

ebg1223:
After some more testing, it seems like the issue has to do with how style props are passed from parent to child component when static rendered next-js vs. how they are rendered client-side. I have some more examples of some idiosyncrasies between the 2 versions.

natew:
@ebg1223 thanks for the detailed write up, I'll take a look at this soon.

ebg1223:
@natew thank you!!

natew:
This should be fixed on latest, lmk

PR #769, opened 2023-03-08
[fix] Style regression on Switch frame

colinhemphill:
Between version 1.6.3 and version 1.7.0 the styles on Switch frames were commented out as if `unstyled` was always applied. Noticed that an `unstyled` prop also stopped being passed through.

| BEFORE | AFTER |
|---|---|
| Screenshot from Switch page with switch frame not styled  | Screenshot from Switch page with frame styles returned  |


natew:
🙏

PR #773, opened 2023-03-09
Fix role value

### LINKED ISSUE #761: [bug] Label component invalid native prop (warning)
**Describe the bug**
 `Label` component warning on native:  
 `Warning: Failed prop type: Invalid prop role of value label supplied to Text`. 
I think it needs something similar to what was made in commit: https://github.com/tamagui/tamagui/commit/817814c5f671240a524603435f40c1c1f4aff218

**To Reproduce**
Add a Label component on native and the warning will be displayed.

**System Info (please complete the following information):** 
OS: iOS (Expo 48)

mphill:
The valid values for role are listed here:

https://reactnative.dev/docs/accessibility#role

label is not one of them, so it's producing an error in the console that appears to have started with expo 48.

This fixes #761 

Based on the description, heading seemed to be the most fitting.

### DIFF
--- a/packages/label/src/Label.tsx
+++ b/packages/label/src/Label.tsx
@@ -108,7 +108,7 @@ const LabelComponent = React.forwardRef<typeof LabelFrame, LabelProps>(
     return (
       <LabelProvider id={id} controlRef={controlRef}>
         <LabelFrame
-          role="label"
+          role="heading"
           id={id}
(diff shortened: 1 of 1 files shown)


mphill:
Another solution to this is:

`role={isWeb ? 'label' : 'heading'}`

natew:
I think this works for now.

PR #774, opened 2023-03-09
fix: remove src on npm tarball

lobor:
Remove src in files field in package json for all packages.

Not import src file in tarball npm
image

### DIFF
--- a/packages/accordion/package.json
+++ b/packages/accordion/package.json
@@ -10,7 +10,6 @@
   "module": "dist/esm",
   "module:jsx": "dist/jsx",
   "files": [
-    "src",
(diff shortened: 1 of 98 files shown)


natew:
My thought was it maybe nice for people in development mode to be able to see the source, but I can see how this slows down CI and weighs down production builds so probably a good idea...

lobor:
in my project, nextjs in docker, image size ~ 300Mo before migrate to tamagui
with tamagui ~ 800Mo
I don't know how tamagui src file take in docker image, but i think is good for production. if people need source, they can go on github on clone repo i think

PR #775, opened 2023-03-09
feat(build): remove sourcesContent from generated source maps, leaving only path reference to src

nderscore:
Right now, all three builds of each tamagui package (esm/cjs/jsx) contains source maps that have `sourcesContent` defined. This means there are three extra copies of the sourcecode included in each npm packge. 

Since the source map already contains a reference linking to the `../../src` path, the `sourcesContent` is likely duplicative. Removing it would significantly reduce the size of every package.

This change doesn't seem to impact the behavior of linking in VSCode or of source maps support on the web.

---

* Updates `@tamagui/build` to pass [`sourcesContent: false` to esbuild](https://esbuild.github.io/api/#sources-content)

### DIFF
--- a/turbo.json
+++ b/turbo.json
@@ -6,7 +6,13 @@
     },
     "build": {
       "dependsOn": ["^build"],
-      "inputs": ["tsconfig.json", "package.json", "src/**"],
+      "inputs": [
+        "tsconfig.json",
+        "package.json",
+        "src/**",
+        "tamagui-build.js",
+        "externalNodePlugin.js"
+      ],
       "outputs": [".next/**", "dist/**", "types/**"]
     },
     "start": {


nderscore:
Also probably worth mentioning: 

From my experience, turborepo's cache doesn't respond to changes in `@tamagui/build`

I think it's because the output of its build script is always `true` and there are otherwise no outputs from the build of the build package, since it has no `.next/**`/`dist/**`/`types/**`

natew:
@nderscore Great idea on the turbo inputs, if you want to add that here that's cool.

natew:
I think we've decided to keep them, it's super useful for development to have the original source, in fact I wouldn't use any UI kit myself that didn't publish like this because otherwise it's simply to hard for anyone to onboard and understand what's going on.

That said, for production it's a pain. I wish there was some blessed way to indicate or publish two versions, but I don't know of anything. If you wanted to avoid bloat in Docker you could run a script to remove all the src dirs, but not to avoid installing in the first place.

If anyone knows any better lmk.

nderscore:
> I think we've decided to keep them, it's super useful for development to have the original source, in fact I wouldn't use any UI kit myself that didn't publish like this because otherwise it's simply to hard for anyone to onboard and understand what's going on.

This PR does not remove the original source from packages. It only removes the additional copy of the source that's baked into each source map in the `sourcesContent` field.

PR #787, opened 2023-03-14
fix(sheet): not change render on rezise sheet

lobor:
see https://discord.com/channels/909986013848412191/994937898908516362/1081409443457138748


natew:
Yea we just need to test that flickering isn't an issue? Not idea what flickering but I assume web.

lobor:
@nderscore @natew 
I remove isResizing, i test on web and android, this work as expected with DemoSheet.
I rebased master
It's ready to review/test

PR #792, opened 2023-03-15
Docs: fix formatting of list-item

peterpme:
i didn't like the formatting on the website so i fixed it

also had javascript errors (forgot an extra paren when copying it over)


natew:
Thanks i should check more closely next time.

PR #796, opened 2023-03-15
fix(themes,theme-base): remove $-0 from space tokens

nderscore:
This PR removes `$-0` from the default set of `space` tokens.

As far as I can tell, the `$-0` token should never be used. `-0px` isn't a valid value in the context of margin or padding - or at the very least, its not meaningfully distinguishable from `0px`.

Removing it should prevent accidental usage of the token from IDE autocomplete/intellisense and reduce some unnecessary noise.


natew:
Looks solid I’ll test and merge tonight.

packages/config-default/src/index.ts
```diff
@@ -62,12 +62,15 @@ export function getDefaultTamaguiConfig() {
...
+  const spaces = Object.entries(size).map(
+    ([k, v]) =>
+      [
+        k,
+        Math.max(0, v <= 16 ? Math.round(v * 0.333) : Math.floor(v * 0.7 - 12)),
+      ] as const
+  )
 
-  const spacesNegative = spaces.map(([k, v]) => [`-${k}`, -v])
+  const spacesNegative = spaces.slice(1).map(([k, v]) => [`-${k.slice(1)}`, -v])
```
nderscore, line 73: Just realized I accidentally added `.slice(1)` to the _key_ here where it doesn't belong - the space keys are not prefixed in this file.

Will revert that piece and repush.

PR #803, opened 2023-03-17
Expo SDK-48 Dependency Fix

jmurphy4:
### Description
The package.json for the static folder had a reference to an expo package that was out of date. The package (@expo/match-media) is just used for testing in this directory, so I simply moved the package to be a dev dependency. 

The issue would arise when upgrading to the new Expo SDK 48, and then running npx expo-doctor. (See pics below)
image
image

When I moved it to be a dev dependency, the error is gone! This was also discussed in [this thread](https://discord.com/channels/909986013848412191/1086003512472772708) in the discord. 

Not sure if there is something else I am missing, I ran all of the tests in the readme. 

If there is anything I missed let me know! 
Thanks!


natew:
I think we can move this to the static-tests folder now but this works, merging and thank you!

Issue #801, opened 2023-03-16
[bug] Input & TextArea border issue


ionutcirja:
1. When I try to set a custom border width for Input and TextArea, on field focus, the field moves 1px to the left.
2. I cannot override border width on focus style.

Thank you

seeden:
Probably connected to https://github.com/tamagui/tamagui/issues/725 as well

natew:
This should be fixed, but also the negative margin trick is specific to our default styles. You can reset it with `unstyled` prop, or just override it manually.

natew:
Should be fixed, lmk if still an issue.

PR #807, opened 2023-03-18
fix(input): default focusable style false value

lobor:
related issue #801 
I'm not sure it's the good way for fix it, but it's workaround which work


natew:
This makes it so when it's *not* focusable it has focus style?

natew:
@lobor Releasing 1.7.11 here in a few minutes, sorry I didn't merge this but needed to do it a different way. Let me know if that fixes it for you though and thanks for your help!

lobor:
> This makes it so when it's not focusable it has focus style?
yes exactly

PR #813, opened 2023-03-20
tamagui root: bump react-native-web to 0.18.10 from 0.18.9

peterpme:
just tried new starter kit, seemed like this wasn't upgraded from resolutions. thanks!

the result of running `npx expo start`
```
Some dependencies are incompatible with the installed expo version:
  react-native@0.71.4 - expected version: 0.71.3
  react-native-web@0.18.9 - expected version: ~0.18.10
```


peterpme:
That's a good point @la55u, I believe the reason is because this is what `npx expo install --fix` returned and I usually abide by their rules

peterpme:
i confirmed this and react-native-web 0.18.10 is in fact what expo calls for. there's another issue around react-native 0.71.4 vs. 0.71.3 but i will open a separate PR for that since there's more to it

natew:
Thanks for the fix!

PR #816, opened 2023-03-21
fix: [#559] ensure keysOrdered is tokenized

### LINKED ISSUE #559: [bug] Custom space/size tokens break the size prop for components
**Describe the bug**

[The docs](https://tamagui.dev/docs/core/configuration#overview) show spacing as an object of key values.

```js
const size = {
  0: 0,
  1: 5,
  2: 10,
  // ....
}
```

Doing that produces the correct types, however, it breaks in this function so size does not work for any component. Slider is a good example.

```ts
export const stepTokenUpOrDown = (
  type: 'size' | 'space',
  name: SizeTokens | SpaceTokens | string = '$true',
  shift = 0,
  bounds = [0]
) => {
  const tokens = getTokens(true)[type]
  const keysOrdered = tokensKeysOrdered.get(tokens) || Object.keys(tokens)
  const min = bounds[0] ?? 0
  const max = bounds[1] ?? keysOrdered.length - 1
  const currentIndex = keysOrdered.indexOf(name)
  if (name === '$true') {
    shift += shift === 0 ? 0 : shift > 0 ? 1 : -1
  }
  const index = Math.min(max, Math.max(min, currentIndex + shift))
  const key = keysOrdered[index]
  return tokens[key] || tokens['$true']
}
```

If you look at the base themes there's a lot more complexity and the map prefixes the keys with `$`. However, if you prefix the keys the types will be wrong. 

```ts
export const size = {
  $0: 0,
  '$0.25': 2,
  '$0.5': 4,
  '$0.75': 8,
  $1: 20,
  '$1.5': 24,
  $2: 28,
  '$2.5': 32,
  $3: 36,
  '$3.5': 40,
  $4: 44,
  $true: 44,
  '$4.5': 48,
  $5: 52,
  $6: 64,
  $7: 74,
  $8: 84,
  $9: 94,
  $10: 104,
  $11: 124,
  $12: 144,
  $13: 164,
  $14: 184,
  $15: 204,
  $16: 224,
  $17: 224,
  $18: 244,
  $19: 264,
  $20: 284,
}

type SizeKeysIn = keyof typeof size
type Sizes = {
  [Key in SizeKeysIn extends `$${infer Key}` ? Key : SizeKeysIn]: number
}
type SizeKeys = `${keyof Sizes extends `${infer K}` ? K : never}`

const spaces = Object.entries(size).map(([k, v]) => {
  return [k, sizeToSpace(v)]
})

```

Not sure of the best path forward here beyond copying all of the stuff from the theme base. I attached a video of it live and a repo.

What makes this work in `theme-base` is cheating on the types :(

```ts
size: size as any as Sizes,
```

**To Reproduce**

Repo here: https://github.com/mwood23/tamagui-size-tokens-bug

https://user-images.githubusercontent.com/13633613/213885390-33b64e26-030d-46c4-9e6b-544d430268c8.mp4

wkoutre:
Fixes #559

Ensures that `keysOrdered` is tokenized (each `key` is prefixed with `"$"`).

Admittedly, this patches a side effect from a separate issue:

- `createVariables` _always_ removes the `"$"` prefix
- `tokensKeysOrdered` stores `size` and `space` tokens without the prefix

So, I'm not sure what the better approach is:

- the fix in this PR
- changing how `tokensKeysOrdered` [is set here](https://github.com/tamagui/tamagui/blob/master/packages/web/src/createVariables.ts#L51) by doing

```
tokensKeysOrdered.set(res, Object.keys(tokens).map((tokenWithoutPrefix) => `${tokenWithoutPrefix}`))
```

### DIFF
--- a/packages/get-size/src/index.ts
+++ b/packages/get-size/src/index.ts
@@ -12,7 +12,8 @@ export const stepTokenUpOrDown = (
   bounds = [0]
 ): Variable<number> => {
   const tokens = getTokens({ prefixed: true })[type]
-  const keysOrdered = tokensKeysOrdered.get(tokens) || Object.keys(tokens)
+  const maybeTokenizedKeysOrdered = tokensKeysOrdered.get(tokens) || Object.keys(tokens);
+  const keysOrdered = maybeTokenizedKeysOrdered.map((maybeTokenizedKey) => maybeTokenizedKey.charAt(0) === "$" ? maybeTokenizedKey : `$${maybeTokenizedKey}`);
   const min = bounds[0] ?? 0
   const max = bounds[1] ?? keysOrdered.length - 1
   const currentIndex = keysOrdered.indexOf(name)


natew:
I'll merge this just to get into the next version, likely your alternative idea is marginally better but I'm seeing we set `tokensKeysOrdered` in two different areas now and perhaps the one in createVariables is actually unnecessary so I want to look into that.

PR #826, opened 2023-03-23
chore(readme): fix starter yarn command

la55u:
there is no `yarn:prod` command, was it meant to be `yarn web:prod`?


natew:
Good catch, thank you.

PR #831, opened 2023-03-24
feat: create @tamagui/react-hook-form

lobor:


la55u:
I absolutely love the idea that Tamagui _could_ be used with react-hook-forms, but I worry that adding it as an official package might introduce a lot of extra maintenance work. The codebase is already massive and I think the focus should be on simplifying things (both for contributors and users), streamlining the workflows and improving performance rather than adding more features that IMO not closely related to the goal. React-hook-forms is a beast in itself and is a very opinionated choice and might not be for everyone's liking and with it officially being supported, feature request will inevitably come to integrate other form libraries too which (I think) would be overkill.

I don't know the internal roadmap of tamagui, sorry if it's already been approved. And I definitely don't mean to underestimate the work being done here by @lobor it is brilliant, I'm just here to state my opinion that I'd advocate for this to be an unofficial, community package rather than inside tamagui.

natew:
My French isn’t so good :) but I honestly don’t think I can merge this for the reasons stated, yes. But that said, we have a lot of community members that maintain packages and if you put it on your own repo I’ll happily add it to the community page. 

In the more distant future I’d like to put together a blessed set of really high quality components that is paid, simply due to the extreme cost of maintenance. I still think this package is a bit too low level for that, but would be happy to collaborate and find a way to have people contribute and benefit. But that’s ways off.

natew:
And I do very much appreciate this work, if you pull it into a repo I’ll send out a tweet as well!

lobor:
Ok. I move this in extra repo and deploy it in package.

lobor:
I begin de create package here 

https://github.com/lobor/tamagui-react-hook-form

dohomi:
@lobor I started long time ago a package `tamagui-extras` which includes `react-hook-form` and `tamagui`
https://tamagui-extras.vercel.app/?path=/docs/ui-forms-reacthookform--documentation
https://github.com/dohomi/tamagui-kitchen-sink#integration-of-react-hook-form

Feel free to try it out and lets join forces in case you miss anything.

PR #837, opened 2023-03-26
addFont function creation

TheEhsanSarshar:
- [x] - make it work properly
- [x] - migrate to a package


natew:
This looks really good, going to release it without documentation and we can give it a run and document later on.

natew:
Maybe we can add `<Beta />` and put it in docs now so we don't lose track.

PR #843, opened 2023-03-26
Update Vite guide example link

horacioh:
not sure if this is the correct example to link to, but from the list of apps this one seems to be the correct one :)


natew:
thank you

PR #851, opened 2023-03-27
feat: add moti driver

nandorojo:
Need to test this still, but would help to get a general review first to make sure I'm in the right direction.

It's best to view this by hiding formatting changes. I saved a few files which triggered prettier to fix their spacing.

Screenshot 2023-03-27 at 2 25 30 PM

### DIFF
--- a/apps/site/data/docs/intro/installation.mdx
+++ b/apps/site/data/docs/intro/installation.mdx
@@ -39,6 +39,16 @@ Be sure to set the environment variable `TAMAGUI_TARGET` to the string `web` or 
 }
 ```
 
+Or for native you'd do something like this:
+
+```sh
+TAMAGUI_TARGET=native react-native start
+```
+
+```sh
+TAMAGUI_TARGET=native expo start
+```
+
 If you're targeting a native app and installing `tamagui` (not core), you'll need to [set up `expo-linear-gradient` first](https://github.com/expo/expo/tree/sdk-47/packages/expo-linear-gradient) as Metro will try and find it even if you don't import it.
 
 ### Config file


packages/animations-moti/src/createAnimations.tsx
```diff
@@ -0,0 +1,134 @@
...
+       * ```
+       */
+      const [animationKey, inlineTransition] = Array.isArray(props.animation)
+        ? props.animation
+        : [props.animation]
+      const animateOnly = props.animateOnly ?? []
+      let transition = animations[animationKey as any]
+      if (!transition) {
+        // @natew is this necessary? probably safer, but it will work with moti without setting this from the theme...
+        return null
```
natew, line 49: so basically saying we didn't find any animation config, what would happen without it? as is its a conditional hook problem

natew: yea seeing inlineTransition its likely fine to not have it bail

packages/animations-moti/src/createAnimations.tsx
```diff
@@ -0,0 +1,134 @@
...
+        }
+      }, [style, ...animateOnly])
+
+      const motified = useMotify({
+        transition,
+        animate,
+        // @natew
+        // can you confirm that i don't need to pass these?
+        // from: pseudos?.enterStyle as any,
+        // exit: pseudos?.exitStyle as any,
```
natew, line 102: yep no need they are handling higher level, we can go back and upgrade with a more efficient one later

packages/animations-moti/src/createAnimations.tsx
```diff
@@ -0,0 +1,134 @@
...
+
+        usePresenceValue: (presence as any) ?? undefined,
+      })
+
+      return useMemo(
```
natew, line 107: we can test this but i have a feeling we wont ever memo, because i don't memo the style object passed in via createComponent, i can revisit doing that but if i remember its just a bit rare for the memo to actually work, worth checking into at some point though

packages/toast/src/ToastImpl.tsx
```diff
@@ -248,6 +248,7 @@ const ToastImpl = React.forwardRef<TamaguiElement, ToastImplProps>(
     const AnimatedView = (driver['NumberView'] ?? driver.View) as typeof Animated.View
 
     const animatedStyles = useAnimatedNumberStyle(animatedNumber, (val) => {
+      'worklet'
```
natew, line 251: btw wheres this stuff coming from? seems unrelated?

packages/animations-moti/src/createAnimations.tsx
nandorojo: oh i see, we use that to find if it animates at all or not…it could be `‘animation’ in props` to determine whether or not it should add animations. this means it would animate even if you don’t have a config with this name at the root. but that’s probably not desired, so the current behavior is fine

packages/animations-moti/src/createAnimations.tsx
nandorojo: got it, i just did added memo because i saw it in another drive

packages/toast/src/ToastImpl.tsx
nandorojo: this is necessary for any reanimated drivers to work when using useAnimatedNumberStyle. you can’t pass non-worklets to useAnimatedStyle, so any function that ends up in a UAS hook must be a worklet

nandorojo: for non-reanimated drivers, this is a no-op

PR #856, opened 2023-03-28
feat(animations): animation hook support for css driver

alitnk:
Some notes when using Sheet if using the css driver:
- Exiting animation would require an `<AnimatePresence>` wrapping the sheet
- User should provide the `animation` key if working with the css driver. We could provide this internally however...
- Each gesture change performs a state change in this driver, so a lot of re-renders. Might be ok with some and not with others... a note on disabling gestures on css driver would be nice on docs

I could add these notes in the Sheet docs if you think this is ready to merge, @natew


natew:
Would be kind of fun to have a way to use different drivers.

Perhaps we could have const animations = createAnimations({})

and then <YStack animation={animations.bouncy} />

If that worked you could load the css driver lazily and then we can show demos that let you toggle between the two...

PR #858, opened 2023-03-28
fix(core): make font props work independently

### LINKED ISSUE #845: [bug] fontSize not converted to valid size overriden from a variant
Considering the following component:

```
import { styled, Text as TamaguiText } from "@tamagui/core";

export const Heading = styled(TamaguiText, {
    name: "Heading",
  color: "$color",

variants: {
    type: {
      "myVariant": {
        fontFamily: "$myFont",
        fontSize: "$24",
        lh: "$24",
        fow: "$bold"
      },
    }
  } as const
})
```

If i do
```
 <Heading type="myVariant" fontSize="$20">
        H1
      </Heading>
```
it will not convert `fontSize="$20"` to a valid font size, unless in explicitly provide a font family to the props.

**Expected behavior**

The fontSize prop should just override the fontSize from the variant

alitnk:
Closes #845 
Ignore the branch name... [the plan changed along the way](https://github.com/tamagui/tamagui/issues/845#issuecomment-1486860978).

### DIFF
--- a/packages/core-test/getSplitStyles.web.test.tsx
+++ b/packages/core-test/getSplitStyles.web.test.tsx
@@ -44,33 +44,27 @@ describe('getSplitStyles', () => {
 
   test(`fonts get merged properly`, () => {
     const CustomText = styled(Text, {
+      name: 'Heading',
+      color: '$color',
+
       variants: {
         type: {
           myValue: {
-            fontFamily: '$mono',
-            fontSize: '$10',
+            fontFamily: '$body',
           },
         },
-      },
+      } as const,
     })
-    const case1 = simplifiedGetSplitStyles(CustomText, {
+    const splitStyles = simplifiedGetSplitStyles(CustomText, {
+      fontSize: '$1', // should be able to resolve $body from the variant and find the css value
       type: 'myValue',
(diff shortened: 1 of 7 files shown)


alitnk:
Something's wrong with the tests... working on it.

packages/core-test/getSplitStyles.web.test.tsx
```diff
@@ -33,6 +42,31 @@ describe('getSplitStyles', () => {
...
+    })
+    const splitStyles = simplifiedGetSplitStyles(CustomText, {
+      fontSize: '$1', // should be able to resolve $body from the variant and find the css value
+      type: 'myValue',
+    })
+
+    expect(
+      splitStyles.rulesToInsert
+        .find((rule) => rule.property === 'fontSize')
+        ?.value.startsWith('var(--')
```
natew, line 66: Lets make test more thorough and check for the right value output, test for non-variant cases, and maybe even nested styled().

alitnk: Added more - media and pseudos still failing so commented out for now. Those should be easy to support if we get rid of `getSubStyle`

PR #864, opened 2023-03-29
refact: move addTheme, updateTheme to @tamagui/load-theme package 

TheEhsanSarshar:
- [x] LGTM


TheEhsanSarshar:
I think we need to rename `@tamagui/add-font` to `@tamagui/load-font` as well.

TheEhsanSarshar:
Right now I marked `addTheme` and `loadTheme` as deprecated. and also this new package `@tamagui/load-theme` is not accessible in `tamagui` because of naming conflicts.
we should remove those deprecated functions first ( maybe in a minor version ), and then add this new package in tamagui namespace

natew:
I sort of like `add` better than `load` because in the future we can add `remove` which is more symmetrical.

TheEhsanSarshar:
you mean this ?
`import { addTheme, updateTheme,  removeTheme } from '@tamagui/add-theme'`

the only reason that changed the name from `add-font` to `load-font` was having those extra functions

natew:
Ah yea technically it's generic... maybe we can do `@tamagui/theme`. That's kind of nice. Same for `@tamagui/font`. Thinking if there's some other overlap but I think it's fine.

TheEhsanSarshar:
@natew both packages are renamed to `@tamagui/font` and `@tamagui/theme`

natew:
Hey @TheEhsanSarshar thanks so much for this, I can fix these conflicts but if you get time to would be great and I'll hit merge.

TheEhsanSarshar:
@natew it's done

PR #866, opened 2023-03-30
feat(sheet): new `moveOnKeyboardChange` prop

alitnk:
Fixes issues of several users who mentioned this on the Discord server

### DIFF
--- a/apps/site/data/docs/components/sheet/1.9.18.mdx
+++ b/apps/site/data/docs/components/sheet/1.9.18.mdx
@@ -1,0 +1,153 @@
+---
+title: Sheet
+description: A simple sheet component
+name: sheet
+component: Sheet
+package: sheet
+---
+
+# Sheet
+
+<Description>A bottom sheet that slides up.</Description>
+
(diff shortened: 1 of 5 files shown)


mphill:
Thank you!

mphill:
The main problem will be that [react native avoid softinput](https://github.com/mateusz1913/react-native-avoid-softinput) requires dev builds to work on expo, which will cause some issues and confusion for sure.

We had a problem with inputs in the sheets where the sheet would start to slide up, but lose focus. So you were never able to enter input. This appears to be fixed now!

The primary issue appears to be that the keyboard avoiding view pushes the sheet up, not the whole window. It's like the keyboard avoiding view only understands how to attach to the highest z-index view.  That seems like a problem in RN for sure.

alitnk:
Yeah that makes sense. Still haven't tried the soft input library but if it works it's nice to have on docs as a recommended library to integrate tamagui with. We'd have an easy way (the prop) and a better way (avoid soft input lib) so people can choose.

PR #869, opened 2023-03-30
Move Text below SizeableText

peterpme:
this is kind of a nitpick / dumb suggestion but when i look at this top "Text" and see the serif-style font for the first example, i can't help but think "font rendering is broken on the web" lol

i would consider replacing Text with "Basic Serif Font" or moving it to the bottom to suggest that the font is properly loading

image


natew:
Let's just remove from the demo

peterpme:
Removed!

PR #873, opened 2023-03-31
fix(animations): add getAnimationDriver as fallback to useAnimationDriver

alitnk:
Users seem to have issues using this and are getting the `Must set animations in tamagui.config.ts at SheetImplementation2 ` .

Not sure why that is happening. Maybe context is not available on the first render and this is throwing? 

P.S. ignore the branch name as it's not a css driver issue, but a general animations one


natew:
Btw imports in general should use the .js suffix in the filename (unless they reference .native or .web which breaks native metro)

wezter96:
It seems like this issue might have resurfaced?
I got it now when upgrading to 1.131 or higher @natew

PR #880, opened 2023-04-01
light docsearch active color change to be more visible

Sabin-Subedi:
Previously the issue was when on light mode the active color of the text and icons on doc search was white which in white background weren't visible. currently i have changed the color to default more visible color.


natew:
ty

PR #882, opened 2023-04-01
feat: next.js + expo router starter

TheEhsanSarshar:
- [x] work on native
- [x] work on web (rnw)
- [ ] use next.js new app directory which integrates seamlessly with expo-router app directory (99%) navigation reusability


TheEhsanSarshar:
I think It's better to keep this PR to use expo-router with the current next.js directory.
because the new next.js app directory is still in beta, we need to have a stable starter app for the expo-router.
and for the next.js new app directory + expo router, we can have a separate PR.
so in this case we have both a stable and a beta starter

and of course, we still have navigation reusability 😃

TheEhsanSarshar:
@natew what's your thoughts on this

natew:
This sounds good to me, I know you can mix and match the new app directory anyway so migration should be doable. Happy to merge just expo router to start.

TheEhsanSarshar:
@natew it's ready for review

starters/next-expo-router/apps/expo/index.js
```diff
@@ -0,0 +1,9 @@
+import 'expo-router/entry'
+// import { registerRootComponent } from 'expo'
```
natew, line 2: lets remove commented code here

natew:
Looks like tests broke. So does expo just find the stuff in `app/`? I thought you'd have to add FS routes to expo just like you do with next and re-export or no?

TheEhsanSarshar:
@natew thanks for reviewing
it seems there is an issue in `yarn install` let me check that.
yea expo look for the content of the `/app`

TheEhsanSarshar:
sorry not sure if I get your last comment correctly, yeah we just re-export those routes just like next.js, and in case if we need to define tabs or stacks or any react-navigation configs we do that inside `_layout.tsx`

TheEhsanSarshar:
follow up this PR 
https://github.com/tamagui/tamagui/pull/959

PR #889, opened 2023-04-03
fix(toggle-group): update focus style to be more visible

alitnk:


natew:
Can we make the border also a bit more prominent especially light mode it's hard to detect.

PR #890, opened 2023-04-03
fix(radio-group): better focus and press style indications

alitnk:


alitnk:
The press could happen outside the dot. We can make the dot change color too but won't happen if user clicks outside the indicator (but still inside the wrapper)

Is that fine?

natew:
I think for now that worksOn Apr 3, 2023, at 8:06 AM, Alireza Zamani ***@***.***> wrote:﻿
The press could happen outside the dot. We can make the dot change color too but won't happen if user clicks outside the indicator (but still inside the wrapper)
Is that fine?

alitnk:
> I think for now that works

Done in cb417c412c169f0e616e9bc84f9c477226f2c2f3

PR #899, opened 2023-04-05
Allow for `alt` attr in Image

VladyslavKochetkov:
It'd be nice to allow `alt` attribute to satisfy a11y linter - and you know - to help people the a11y is supposed to help.

### DIFF
--- a/packages/image/src/Image.tsx
+++ b/packages/image/src/Image.tsx
@@ -32,6 +32,7 @@ type BaseProps = Omit<
   width: number | string
   height: number | string
   src: string | StyledImageProps['source']
+  alt?: string
 }
 
 export type ImageProps = BaseProps & Omit<StackProps, keyof BaseProps>


natew:
Try out `accessibilityLabel`. Eventually we can diverge `web` from `core` and support all web props there, but that requires a bit more effort.

VladyslavKochetkov:
Yeah I understand that accessibilityLabel is there, but from my experience most projects using eslint a11y which by default enabled https://github.com/jsx-eslint/eslint-plugin-jsx-a11y/blob/db64898fa591f17827053ad3c2ddeafdf7297dd6/docs/rules/alt-text.md and setting accessibilityLabel doesn't satisfy that rule.

natew:
Yea I'd like us to move to web-first props, I think RN is doing this in general. Having two of the same prop isn't ideal but we could mark the RN ones as deprecated. That would need to be in a PR along with the right logic to make sure it adapts it to native.

VladyslavKochetkov:
Already - can't wait to see it, thanks @natew

PR #913, opened 2023-04-07
Add option --declaration-root to tamagui-build

dohomi:
I had the requirement to modify the declarationDir of esbuild. With this PR users can run
```
tamagui-build --declaration-root
```
To create declarations into the root folder. This is specific if library authors wants to export multiple folders of one package like:
```
import {} from 'tamagui-extras '
import {} from 'tamagui-extras/submodule'
```


dohomi:
@natew I was digging deeper into tamagui-build to support type generation into the root folder. Currently I run a separate task via `tsc` but I guess it would be ideal if this option would make it into `tamagui-build` directly.

PR #920, opened 2023-04-07
feat: change animation config, driver

TheEhsanSarshar:
- [x] change configs of the same driver
- [ ] change driver


TheEhsanSarshar:
right now you can change the configuration of the same animation driver, but switching to another animation driver is a challenge here.
the challenge is with React hooks. because each driver uses different hooks, and changing hooks at runtime is not possible in React.
there are two solutions that came to my mind for this
 - moving away from hooks and using simple functions
 - or making the drivers use the same hooks internally for example `useAnimations` hook should have the same number and order of hooks in all drivers.

natew:
Ah yea changing them dynamically is tricky, did we have a use case for this? @alitnk did get multiple animation drivers working for the demos recently, you can see the tiny toggle top left that does it. It's been a bit since I looked at what he did there.

alitnk:
What I did basically was using context to store the driver AND change the `key=""` on the child so they don't break when we toggle them. Without the key it'd give some error about hooks changing.

It works fine for the purpose of our demos. Not sure the other cases where this feature would be useful for.

TheEhsanSarshar:
you mean the use case for changing the driver? I don't know I just picked a task from `next.mdx` 😁
yeah so basically by changing the key it re-mounts the component, not sure if it's a good idea to remount all the components inside the app.

TheEhsanSarshar:
other than demos which Alireza already found a solution for that tbh a clever solution, I think we don't have any use case for this, users don't change the driver.
I guess we can either close this PR or keep that part where it changes the configuration of the current driver.

alitnk:
Yeah remounting is definitely not ideal, but I can't come up with a better way given how animation hooks work.

We should also organize the tasks on `next.md` better... 😶

TheEhsanSarshar:
I guess that's the only way to skip the strict laws of React Hooks 😄
anyway, I think it's better to close this PR, it just adds some complexity.

PR #930, opened 2023-04-10
Fix: elevation android

### LINKED ISSUE #897: Elevation not working on Android
The `elevation` prop doesn't seem to be working on Android. It render appropriate shadow on iOS.

```jsx
<TCard bg="$bg.mainBg" elevation="$4" p="$2" mb="$5">
  <XStack jc="space-between">
    <TransferShortCut />
  </XStack>
</TCard>
```

Furthermore, manually setting elevation in `style` also doesn't work. I tried it on the `Card` and `Stack` component, so I assume the rest are the same.

Tamagui v1.9.15

TheEhsanSarshar:
fix #897


natew:
@TheEhsanSarshar I think our elevation is different from android elevation though? Technically we just turn elevation into a shadow ourselves. So I'm wondering is the shadow we create just not working for some reason?

I think it'd be confusing to have Android use the built-in android elevation but the rest don't, but maybe the `native` prop or something could do this down the road.

TheEhsanSarshar:
@natew The only way to set shadow in Android is using the native `elevation` style property, it doesn't support shadowRadius or other properties.

not sure if somewhere else we are converting those shadow properties into platform compatible primitives, but we need to set `elevation` somewhere for Android to work.

or as you said we can use some sort of native prop like `androidElevation`

natew:
Oh interesting! Wow I had no idea, shows how much I do Android dev... also crazy they can't support that? The `* 2` seems a bit arbitrary but if it seems to match up well that's fine, we ideally would have some platform detection here (does it warn on web/ios?)

TheEhsanSarshar:
yeah, that's crazy and really a pain, for custom shadows you need to use SVGs and Linear gradients.

I come up with that arbitrary number `* 2` by checking this shadow generator [react native shadow generator ](https://ethercreative.github.io/react-native-shadow-generator/) and also some search on the internet.

ok, I'll make it to only add on Android

TheEhsanSarshar:
not ready yet!

VladyslavKochetkov:
Just commenting on this as well, I've seem a few other libraries do the same thing as @TheEhsanSarshar is suggesting, (I think including native base), so I think it's fine to ship. The concept being that a shadow that is off is better than no shadow at all, and doing `2 * ` gives a close enough result. Maybe just add a note somewhere in the docs that says the Android doesn't support shadow colors?

natew:
Yea I think this is fine, elevation is our own custom thing in tamagui anyway and we're working on headless for everything so lmk once ready and I'll test this out.

TheEhsanSarshar:
I think to avoid confusion it's better to use `elevationAndroid`. especially it will help to override that `* 2` number if doesn't always work.

TheEhsanSarshar:
`elevationAndroid` is working right now, `elevation` variant internally uses it, and also you can set its directory as a prop to override the result of the `elevation` variant.
@natew I can add some docs if you like this implementation. otherwise happy to hear your suggestions about the API

TheEhsanSarshar:
@natew it's ready

natew:
Looks good I'll test and merge soon

PR #946, opened 2023-04-13
fix: update theme not merging bug

TheEhsanSarshar:


natew:
Since updateTheme is a rare use case I think we can get away with changing the behavior here. So long as we add “replaceTheme” that just uses that flag internally to avoid merge.

natew:
Replace theme = old functionality, update theme we change how it works to be merge by default.  On Apr 13, 2023, at 8:23 AM, Ehsan sarshar ***@***.***> wrote:﻿
ok so you mean we should create a replaceTheme and use the update flag there?

TheEhsanSarshar:
right now `addTheme` can do the job of replaceTheme as well. 
I can add replaceTheme function that internally just call `addTheme` like we do for `updateTheme` ?

natew:
Yes the implementation is the same just share a function they both call. Very minor difference.

TheEhsanSarshar:
@natew please take a look if you got some free times

natew:
Looks good on first glance, I'll test it in a bit.

natew:
I guess for SSR if you call these before load it will be odd right, maybe break? Since it could happen anytime during a render too. We can make a rule / document it's client only for now as that's the primary use case. Docs updating would be great here too.

TheEhsanSarshar:
@natew please have a look

natew:
@TheEhsanSarshar final q have you tested on native? btw i invited you to be part of the tamagui github org so you dont have to work off a fork and i can test your changes much easier locally.

natew:
because on thing i think we miss on native is we need to trigger re-render of themes that use that. i think our existing examples had the same missing feature too, since they re-render in the same component. but if you added a sub-component or sibling component that is memoized or something i assume it won't actually update. this may even be an issue on web for animated ones which also would need ThemeManager.notify() to run i assume, or something like that.

alitnk:
Tried on iOS and working fine as far as I can tell. 

Did you add the new rectangle btw? @TheEhsanSarshar 
image

TheEhsanSarshar:
@alitnk sorry for not replying 
yeah I added that

natew:
I didn’t see it in the source code where is it? We should remove it’s confusing.

TheEhsanSarshar:
@natew let me check that again 
I guess I made a mistake in doc

TheEhsanSarshar:
added that extra box to show how update-theme is working. 
maybe instead of `set random color` it should be `update left box` or something

PR #954, opened 2023-04-14
fix(toast): bugs

alitnk:
fixes:
- events and multiple viewports where only the last viewport's events work
- variant types not being correctly inferred
- `additionalInfo` not being typed on `useToastState()`
- ability to pass `viewportName` to `useToastController()` 

also deprecates `additionalInfo` in favor of `customData`

### DIFF
--- a/packages/toast/src/ToastImpl.tsx
+++ b/packages/toast/src/ToastImpl.tsx
@@ -34,11 +34,6 @@ const ToastImplFrame = styled(ThemeableStack, {
     unstyled: {
       false: {
         focusable: true,
-        focusStyle: {
-          outlineStyle: 'solid',
-          outlineWidth: 1,
-          outlineColor: 'red',
-        },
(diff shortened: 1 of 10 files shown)


packages/toast/src/ToastImpl.tsx
```diff
@@ -164,17 +160,22 @@ const ToastImpl = React.forwardRef<TamaguiElement, ToastImplProps>(
     const closeTimerRemainingTimeRef = React.useRef(duration)
     const closeTimerRef = React.useRef(0)
     const { onToastAdd, onToastRemove } = context
+
+    const viewport = React.useMemo(() => {
```
nandorojo, line 164: there’s no benefit to memoizing this btw

alitnk: Fair enough. I have a bad habit of memoizing everything just in case...

PR #960, opened 2023-04-15
fix(site): show all icons in demo by removing unnecessary slice

la55u:
I removed the `.slice(0,835)` from the code to correctly list newly added icons too.


natew:
ty

PR #963, opened 2023-04-16
fix: TextInput some missing, and deprecated props

TheEhsanSarshar:
updated text input based on the latest changes in React Native Web

[ReactNativeWeb:TextInput](https://github.com/necolas/react-native-web/blob/master/packages/react-native-web/src/exports/TextInput/index.js)

### DIFF
--- a/packages/react-native-web-lite/src/TextInput/index.tsx
+++ b/packages/react-native-web-lite/src/TextInput/index.tsx
@@ -45,9 +45,7 @@ const setSelection = (node, selection) => {
     const { start, end } = selection
     try {
       node.setSelectionRange(start, end || start)
-    } catch (e) {
-      // ok
-    }
+    } catch (e) {}
   }
 }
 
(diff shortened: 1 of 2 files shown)


natew:
@TheEhsanSarshar looks good, lmk when ready

natew:
we probably want to deprecate react-native-web-lite, not sure if rnw properly supports concurrent mode on some things still (especially ScrollView) but maintaining the fork is a lot of hassle for us, and looks like they started making some progress there finally too.

natew:
@TheEhsanSarshar actually bringing in some of the new style changes to tamagui itself will be a huge help, see these commits:

CleanShot 2023-04-16 at 07 01 04@2x

eg transform string syntax looks cool

TheEhsanSarshar:
@natew the changes to `react-native-web-lite` is ready, but I should change tamagui itself as well.
I thought `rnwl` is our source for cross-platform stylings.
where do we use `rnwl` ? is it for expo web ?

natew:
@TheEhsanSarshar nope i just was frustrated the rnw didn't support concurrent mode or tree shaking, and also i needed to use some internal libraries inside it without the rest. so you get a bit better tree shaking and concurrent mode support, but other than that its not necessary. but yes getting tamagui to support all the new RN stuff is great to follow on. i'll merge this and test it out.

PR #978, opened 2023-04-18
fix(site): remove code causing hydration issues

alitnk:


natew:
This only fixes because incidentally, the real issue is something to do with Button you can see it gets a different `fow` className on server.

PR #980, opened 2023-04-19
Update tsconfig.json of starter

dohomi:
If users want to publish projects then "preserve" will fail for example Storybook to compile the ui elements.

### DIFF
--- a/starters/next-expo-solito/packages/ui/tsconfig.json
+++ b/starters/next-expo-solito/packages/ui/tsconfig.json
@@ -4,9 +4,7 @@
     "src"
   ],
   "compilerOptions": {
-    "composite": true,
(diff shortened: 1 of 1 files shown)


la55u:
Isn't it needed for the compiler or something like that?
Not sure how that option works though just remembered it being mentioned:
https://discord.com/channels/909986013848412191/909986013848412194/1082383221133680761

dohomi:
@la55u in `@tamagui-extras/*` mono repo this flag preserve that Storybook was able to parse the library.

natew:
Checked and it still works because tamagui-build builds to `dist/jsx`, but committing a different one with jsx: 'react-jsx' here.

PR #1014, opened 2023-04-23
feat: Factor out tamagui variables example to allow more customization

kaceycleveland:
Recently when using Tamagui, I realized the amount of work needed to build out some basic components in house. Originally the idea was to use `@tamagui/core` and build our components in house for our design system but its tough to do that because it requires a large chunk of time.

The next thing I tried was plucking out some components from `tamagui` like the `@tamagui/slider` component and using that however, despite eventually getting it working "ok", I am not super happy with how it turned out and now our theme variables are pulling in the default tamagui fields (`background`, `backgroundHover`, etc) which I do not want. I want people using our design library to only really be allowed to use our designated tokens in our library.

So that leads to this PR where I wanted to support overriding the default variables `tamagui` components are built on top of such that I can provide my own mapping to my own design system/tokens. Since the components need to know of the variables at build time, I opted to alias the imports and provide an optional custom override of the mappings. If there are other ways to do this, please let me know; I think this way it relatively clean/safe and allows for adopting most/all of `tamagui` components without needing to adhere to the variable namings!

If this is a valid solution, I can build this out and incrementally support more `tamagui` components and document usage.


natew:
Yes this is a valid path, but the only issue is that our extractor would need to support then parsing the imported variable usage so it could turn that into CSS at build-time. Now, granted, we don't extract *everything* at build time even now, but we should directionally go that way (for example, variants right now aren't optimized, but could be pretty easily with some logical improvements).

The good news is we do this sort of cross-module analysis already in the compiler. The bad news is it only works with the `whitelist` files you pass in, not with modules themselves. So we'd have to get that working. Good news again is that shouldn't be *too* hard, but it does require some prodding.

To build this out you'd look at how `importsWhitelist` works, then look at the `CallExpression()` part of `@tamagui/static` `createExtractor` works, then in the repo you could test it out by running `yarn sandbox:extract`, adding `//! debug-verbose` to the top of a the file you want to test it on in the ui, run `yarn watch` and basically add in the feature so that it ultimately gets down to CSS.

natew:
Looking at `getStaticBindingsForScope.ts` it may actually even support modules already.

Also not sure I love the build-step necessary to change it, we'd have to support that as well. What we could do to avoid some of this is:

1. Allow this to be configured at the `createTamagui` level, so define `defaultVariableNames` and pass in your object.
2. We'd have to change code then to map these in a few different places... probably mostly somewhere near `propMapper` but I'd have to look into it
3. Would avoid all need for static fancy stuff though

kaceycleveland:
That was actually more inline with my initial attempt but I realized it became more complex then that.

I had a new option in `createTamagui` but ultimately the part I stopped at was what sounds like touching `propMapper` or anything like that.

@natew What are your thoughts on just doing the mapping per component at the theme level? I sort've did this at the highest up point where I map my tokens to the `tamagui` ones in the theme definitions but I was thinking a new package that operates similar to the `Theme` component in which mappings to the `tamagui` variables can be done.

Something like...

```
import { Slider } from '@tamagui/slider'


export const MyCustomSlider = () => {
   return <TamaguiThemer mapping={{ background: myCustomToken }}><Slider /></TamaguiThemer>
}

```

I don't know the underlying details of how theming has been implemented so not sure if there are any performance concerns or anything like that.

kaceycleveland:
Looking at `propMapper`, I remember actually poking around this when messing with theming. I will take a stab at supporting some mapping here!

natew:
Appreciate the work here, we can probably come up with something better in the future.

PR #1015, opened 2023-04-24
Change sandbox to use lazy imports and Suspense to allow vite to work

kaceycleveland:
Small fix since the sandbox for vite seems to be broken for me (and probably others?)


natew:
ty

PR #1024, opened 2023-04-24
fix: Allow a config path to be passed without erroring

kaceycleveland:
In our monorepo setup there was an error being thrown that others in the Discord seemed to have hit. It may have solved a lot of users cases with some recent fixes, but I think for monorepos or custom config locations it was still broken because `getDefaultTamaguiConfigPath` would get called and error out despite the config being ignored later.

This should fix it but ultimately to come up with a test environment for this small fix would be more effort then the issue itself. Let me know if I should come up with a test somewhere but ultimately this is a fairly small fix.

### DIFF
--- a/packages/static/src/extractor/loadTamagui.ts
+++ b/packages/static/src/extractor/loadTamagui.ts
@@ -142,7 +142,7 @@ export async function getOptions({
     tsconfigPath,
     tamaguiOptions: {
       components: ['tamagui'],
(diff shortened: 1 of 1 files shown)


jonsherrard:
This has broken `next dev` for me, next build still works.

natew:
@jonsherrard I added better logging here, and made it only warn because this technically shouldn't break anything its just for a new feature. Let me know what it logs. It should work, the starter and site and other repos are working.

PR #1025, opened 2023-04-25
fix: Fix a previous fix for allowing custom config location

kaceycleveland:
I tested this one by creating an extra `tamagui.config.ts` in a folder and pointing the `config` field to it.

It was a variable that I just overlooked 🙃


natew:
It seems this totally ignores `existingPaths`? Seems like a bug. I can read this over more carefully and see what's going on in a second

natew:
pushed something to master though this shouldnt be causing issues in the app at most just a warning

PR #1040, opened 2023-04-26
add the remove type module step

imjulianeral:
The support for type module was removed because it was breaking things as discussed here: https://discordapp.com/channels/909986013848412191/1088725438001590282/1100863190579036211

So to get a successfull instalation of the vite plugin the developers must delete that config in their package.json.


natew:
Haha at the wording

PR #1072, opened 2023-04-30
remove .js extension form ToastImperative import in Toast.tsx

hamdiwanis:
Is "ToastImperative" import with .js is explicit for a reason? if not it broke my mono repo setup on web for some reason toast controller was returning empty object.


natew:
What broke?

hamdiwanis:
toast controller was returning empty object.

PR #1091, opened 2023-05-01
Fix Card Props

DavideSegullo:
I don't know if it is an intentional thing, but I noticed that the card, due to the spread of the 'props' object, was losing the variant for the 'size', I also added `styleable` type instead of `extractable` which is deprecated


natew:
ty

PR #1148, opened 2023-05-10
Improve Lucide Icons

DavideSegullo:
Hello, I’m working to improve lucide icons package, the goals of this PR are:

- [x] Update icon set with the latest release of Lucide Icons
- [x] Solve an inheritance problem in SVGs, currently I can't set a stroke width because the children overwrite the parent's width


DavideSegullo:
I did some tests and everything seems okay to me

Gabrola:
Thank you! Was just having the same problem with not being able to apply `strokeWidth` to the icons

Gabrola:
Thank you! Was just having the same problem with not being able to apply `strokeWidth` to the icons. I was about to submit a PR for this, but your solution is much better! It might be nice to add `strokeWidth` as a prop though where you can pass a size token. It's a very simple change in `IconProps.ts` and `themed.tsx`

natew:
Looks great and Gabrola seems like he has good points.

DavideSegullo:
> 

Yes, it's a good idea, I'll add it, thanks for your feedback

PR #1157, opened 2023-05-13
Finalize css driver layout animation remained parts

byteab:


packages/web/src/createComponent.tsx
```diff
@@ -782,7 +788,24 @@ export function createComponent<
       }
     }
 
-    content = createElement(elementType, viewProps, content)
+    content = createElement(
+      elementType,
+      viewProps,
+      children && layout
```
natew, line 794: this should come from the driver somehow, we're really trying to remove as much as possible from `web` so features always should add code from outside.

packages/cubic-bezier-animator/src/index.tsx
```diff
@@ -49,3 +51,25 @@ export function animate(param: AnimateProps) {
   }
   requestAnimationFrame(frame)
 }
+
+function getCubicBezierInput(str: string): [number, number, number, number] {
```
natew, line 55: similar idea could we make this generic? seems like it should almost be passed in somehow or configured so people can decide what it is / control it

packages/web/src/createComponent.tsx
TheEhsanSarshar: thanks for the review!
ok make senese

packages/web/types/types.d.ts
```diff
@@ -764,6 +765,8 @@ export type UseAnimationHook = (props: {
     pseudos: PseudoProps<ViewStyle> | null;
     onDidAnimate?: any;
     delay?: number;
+    childrenRefs?: MutableRefObject<(HTMLElement | View)[]>;
+    layout?: boolean;
```
TheEhsanSarshar, line 769: layout prop is here

TheEhsanSarshar:
WIP.

Issue #1143, opened 2023-05-10
it is weird that expo support web but building failed


tylerrrkd:
### Current Behavior

after run the script `yarn native`, it show the below tips

```bash
› Choose an app to open your project at http://[my ip address]:19000/_expo/loading                                                                           ng
› Metro waiting on exp://[my ip address]:19000
› Scan the QR code above with Expo Go (Android) or the Camera app (iOS)      

› Web is waiting on http://localhost:19000

› Press a │ open Android
› Press w │ open web

› Press j │ open debugger
› Press r │ reload app
› Press m │ toggle menu
```

though `tamagui` support web build with script `yarn web`, and when I just try to press `w`, it will throw an error

```bash
Web Bundling failed 5000ms
The package at "..\..\node_modules\next\dist\shared\lib\bloom-filter\base-filter.js" attempted to import the Node standard library module "crypto".       
It failed because the native React runtime does not include the Node standard library.
Learn more: https://docs.expo.dev/workflow/using-libraries/#using-third-party-libraries
```

I think `press w` may not be shown when I just run the script with `native` or devs may feel confused

### Expected Behavior

`press w` may not be shown when devs just run the script with `native` 

### Tamagui Version

```markdown
1.27.1
```

### Reproduction

```markdown
first running after `npm create tamagui@latest`
```

### System Info

```markdown
System:
    OS: Windows 10 10.0.19045
    CPU: (12) x64 Intel(R) Core(TM) i7-8700 CPU @ 3.20GHz
    Memory: 5.11 GB / 15.87 GB
  Binaries:
    Node: 16.13.2 - C:\Program Files\nodejs\node.EXE
    Yarn: 3.5.0 - ~\AppData\Roaming\npm\yarn.CMD
    npm: 8.1.2 - C:\Program Files\nodejs\npm.CMD
  Browsers:
    Chrome: 112.0.5615.139
    Edge: Spartan (44.19041.1266.0), Chromium (113.0.1774.35)
    Internet Explorer: 11.0.19041.1566
  npmPackages:
    @babel/runtime: ^7.18.9 => 7.21.0
    @manypkg/cli: ^0.19.1 => 0.19.2
    check-dependency-version-consistency: ^3.0.3 => 3.0.3
    eslint: ^8.21.0 => 8.39.0
    node-gyp: ^9.3.1 => 9.3.1
    prettier: ^2.7.1 => 2.8.8
    turbo: ^1.8.3 => 1.9.3
[...]

Issue #846, opened 2023-03-27
Can't start web app when SolitoImage is used by a component 


la55u:
When trying to start the nextjs app with `yarn web` from the root of the monorepo, I get the following error:

```

Tamagui Webpack Loader Error:
   createCSSVariable expected string: undefined
Error: createCSSVariable expected string: undefined
    at createCSSVariable (/home/la/Projects/tamagui-bug/node_modules/@tamagui/core-node/dist/index.js:5838:17)
    at variableToCSS (/home/la/Projects/tamagui-bug/node_modules/@tamagui/core-node/dist/index.js:10563:63)
    at /home/la/Projects/tamagui-bug/node_modules/@tamagui/core-node/dist/index.js:11016:78
    at createTamagui (/home/la/Projects/tamagui-bug/node_modules/@tamagui/core-node/dist/index.js:11086:9)
    at loadTamagui (/home/la/Projects/tamagui-bug/node_modules/@tamagui/static/dist/cjs/extractor/loadTamagui.js:155:40)
    at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
    at async Object.load [as loadTamagui] (/home/la/Projects/tamagui-bug/node_modules/@tamagui/static/dist/cjs/extractor/createExtractor.js:92:42)
    at async extractToClassNames (/home/la/Projects/tamagui-bug/node_modules/@tamagui/static/dist/cjs/extractor/extractToClassNames.js:78:3)
    at async Object.loader2 (/home/la/Projects/tamagui-bug/node_modules/tamagui-loader/dist/cjs/loader.js:62:23)

Missing theme, an error occurred when importing your config
Got config: [Function (anonymous)] {
  displayName: 'ProxyWorm - Check excludeReactNativeWebExports',
  _isProxyWorm: true
}
Looking for theme: displayName

```

## Reproduction
Repo: https://github.com/la55u/tamagui-solitoimage-repro
1. run `yarn`
2. run `yarn web`
3. you should see the script exited with the error above
 
Note that the only thing in the ui package is a simple component that uses SolitoImage. When I try to replace that with any other tamagui component it works.
Not sure if this is a tamagui or solito error, but posting it here because it's included in the starter so it's expected to work. 

Any help would be appreciated, thank you, love Tamagui <3

deggertsen:
Same issue here. When I import and use SolitoImage in my UI (creating an Image wrapper component) I get this when trying to run the web app.

```
Tamagui Webpack Loader Error:
[...]

PR #1173, opened 2023-05-16
(fix) EAS update / build / expo go compatibility in solito starter

benschac:
The current next-solito-starter fails in multiple places when creating an update with the expo eas service to view and share your tamagui project in the expo go app.

A couple of things I need to do to get the starter to work on expo go.
- remove expo-image: only works in development builds and sdk47. The starter doesn't use the package anywhere. (https://github.com/expo/expo/tree/main/packages/expo-image#installation)
- update eas with a runtime https://docs.expo.dev/eas-update/runtime-versions/#setting-runtimeversion
- Conditionally add toasts if not in expo go. Expo Go doesn't support native modules. IMO not having toasts in expo go outweigh the cost of not being able to use the starter in expo go.
- updated the app.json my app to `yourprojectname` to be a little bit more on the nose to the developer using the project to change the value.

Reproduce:
- `npm create tamagui` with changes
- add project_id
- `eas update --profile development`
- scan code from the console, open expo go to view starter app.

also fixes:
- fix: #1143 
- fix: (sorta?) #846 https://github.com/nandorojo/solito/blob/18b283a0dbdfdc78f9115a460e4bfbf9b3d57c33/src/image/use-solito-image.ts#L28 depends on `expo-image` which doesn't work with expo-sdk 48

### DIFF
--- a/starters/next-expo-solito/apps/expo/app.json
+++ b/starters/next-expo-solito/apps/expo/app.json
@@ -1,8 +1,8 @@
 {
   "expo": {
-    "name": "myapp",
-    "slug": "myapp",
-    "scheme": "myapp",
+    "name": "[yourprojectsname]",
+    "slug": "[yourprojectsname]",
+    "scheme": "[yourprojectsname]",
     "version": "1.0.0",
     "orientation": "portrait",
     "icon": "./assets/icon.png",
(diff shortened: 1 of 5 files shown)


starters/next-expo-solito/apps/expo/app.json
```diff
@@ -18,22 +18,24 @@
...
       "adaptiveIcon": {
         "foregroundImage": "./assets/adaptive-icon.png",
         "backgroundColor": "#FFFFFF"
       },
-      "package": "com.tamagui.myapp"
+      "package": "com.[yourprojectsname].app"
     },
     "web": {
-      "favicon": "./assets/favicon.png",
-      "bundler": "metro"
```
benschac, line 32: fixes: https://github.com/tamagui/tamagui/issues?q=is%3Aissue+is%3Aopen+expo+go

the end functionality is just turning of expo-web bundling which seems like what a developer would want using this starter with next

natew:
this is heroic work, thank you

matthewhausman:
@natew should also add 

`"plugins": [["expo-router", { "origin": "https://www.czero.today/" }]],`

to the expo config, I wasn't able to build the app without it

benschac:
> this is heroic work, thank you

ty 🥲

matthewhausman:
@benschac 

I still am not able to make an update successfully even with integrating the changes in this PR, adding the expo router plugin to eas config, etc...

(non native) Updates just cause app to be rendered as a black screen

benschac:
@matthewhausman - can you send a link with a reproduction of the issue?

matthewhausman:
Should have time to make one later today

matthewhausman:
@benschac 

https://github.com/matthewhausman/ios-preview-build-failure-reprod

matthewhausman:
Seems that the reprod actually updates, must be a problem with my personal project sorry!

PR #1195, opened 2023-05-21
fix(docs): link to tamagui config

stephdotnet:
Hi, thanks natew for this guide, it shows various ways to use tamagui styled system, it's really helpful.

There was juste a 404 on the link to tamagui config file in starter.

Thanks for your work


natew:
Oh good catch, thank you.

PR #1207, opened 2023-05-24
feat(web): add theme,platform based style prop

TheEhsanSarshar:
- [x] add web support
- [x] add native support
- [x] add typescript
- [x] final tests


natew:
so i cleaned up some stuff but one big problem:

`numberOfMediaStyles` and the whole `:root:root` thing shouldn't be how we do it when we have order.

That reduce over the props is slow, but also we can just use `usedKeys` to avoid even applying them at all and never have to mess around with `:root` at all. In fact I think we can save runtime and css size by avoiding that altogether. 

But we do need to *merge* the styles so its probably a little bit more complex than just using usedStyles. But yea basically thats the strategy.

TheEhsanSarshar:
@natew 
yeah that's definitely a better approach, we used that `usedKey` approach when shouldDoClasses is false.
and we can use the same approach for classes as well

PR #1236, opened 2023-05-31
fix(slider): allow Thumb size styles to be overridden

nderscore:
Currently, it's not possible to manually define dimension styles on `Slider.Thumb`:

```tsx
<Slider.Thumb
  width="$4"  // this does
  height="$2" // nothing! 😭 
/>
```

This is because inside the wrapper around that component, the `size` prop is always inserted _after_ any custom user-defined props.

This PR moves the spreading of user-defined props to be after any predefined style props, enabling consumers to define or override `width`/`height`/`minWidth`/`minHeight` styles without using the `size` prop.


natew:
Nice, we should make a `createSlider` headless style like `createSheet` is now next.

PR #1247, opened 2023-06-03
docs(styled): clarify comments on `stylable`

la55u:
Follow up to a discord discussion that pointed out a typo in the docs.

Also, I don't understand what this means, and thus can't rephrase it either, could you clarify it a bit?

> *Note that theme will automatically now happen above this HigherOrderStyledText so it's available if you use useTheme.*


TheEhsanSarshar:
the component that you pass to a styleable will also be wrapped with a `<Theme>`.
check this line. https://github.com/tamagui/tamagui/blob/99010f28c0611d85b75e14791e3fc55a352b99eb/packages/web/src/createComponent.tsx#L1004

PR #1265, opened 2023-06-07
update workspace

benschac:
https://yarnpkg.com/features/workspaces#workspace-ranges-workspace

I had some build issues where yarn was looking for package names in npm registry, when it should have been pulling a workspace.

using `workspaces:*` fixed the issue of npm checking a registry and 404'ing

### Steps
- `yarn` in root dir
-  `yarn watch` in root dir
- `yarn native` in root dir
- `yarn web` in root dir

both native:ios and web compile and run w/o issue.

### DIFF
--- a/starters/next-expo-solito/apps/expo/package.json
+++ b/starters/next-expo-solito/apps/expo/package.json
@@ -11,10 +11,10 @@
   },
   "dependencies": {
     "@babel/runtime": "^7.18.9",
-    "@my/ui": "*",
+    "@my/ui": "workspace:*",
     "@react-navigation/native": "^6.1.6",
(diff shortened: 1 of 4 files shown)


benschac:
tests are failing, seems like it unrelated to the change (many PRs have failing tests)

benschac:
I forgot to run `yarn` 🤦‍♂️.

natew:
I'm ok with the change but want to know why it errored, because in the tamagui monorepo we don't use this pattern and it works totally well. Only time you see errors if is you haven't built I think, or if the version doesn't match.

natew:
Granted this isn't published anywhere so it can get away with the looser workspace:, so i'll merge.

PR #1266, opened 2023-06-07
Allow absolute path to tamagui config

rawrmonstar:
Updates logic to check if the path passed into the extractor is an absolute path, if it is, uses that directly instead of prepending the current working directory.

I wasn't able to get kitchen-sink or site running locally so if someone that is able to run those could please double check I didn't break this I would appreciate it 🙇

### DIFF
--- a/packages/static/src/require.ts
+++ b/packages/static/src/require.ts
@@ -1,4 +1,4 @@
-import { isAbsolute, relative, sep } from 'path'
+import { relative, sep } from 'path'
 
 const nameToPaths = {}
 const Module = require('module')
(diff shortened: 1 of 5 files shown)


packages/static/src/require.ts
```diff
@@ -98,7 +98,7 @@ ${err.stack}
 
 You can see if it loads in the node repl:
 
-require("./${relative(process.cwd(), path)}").default
+require(${isAbsolute(path) ? path : `./${relative(process.cwd(), path)}`}").default
```
natew, line 101: this one shouldn't be necessary its just showing how to load the built config. in fact we could even make it always an absolute path potentially, but easier probably to revert

PR #1275, opened 2023-06-09
Move NextJS starter app dir user detail page into correct path

andrewmunsell:
Currently the starter template, inside of the app dir, has the user detail page at the incorrect path. It should be under a user folder to match the pages version and the Expo Router version of the app.


natew:
ty

PR #1277, opened 2023-06-11
accessibility improvement: native support web `aria-*` and `role`

TheEhsanSarshar:
as most of the core Tamagui components are using `aria-*` and `role` for accessibility, it means that accessibility is mostly broken in native, in this PR it's mapping those web-specific aria and role props to its React Native counterpart.

### DIFF
--- a/packages/web/src/helpers/getSplitStyles.tsx
+++ b/packages/web/src/helpers/getSplitStyles.tsx
@@ -19,6 +19,7 @@ import { useInsertionEffect } from 'react'
 import { getConfig, getFont } from '../config'
 import {
   accessibilityDirectMap,
+  accessibilityWebRoleToNativeRole,
   nativeAccessibilityState,
   nativeAccessibilityValue,
   webToNativeAccessibilityDirectMap,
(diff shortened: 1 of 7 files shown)


natew:
Wonderful stuff, could you refactor it so the web won't load it? Basically put it into it's own file that's imported where `.native.ts` includes it, but `.web.ts` (or non-prefixed) doesn't. If it saves some runtime cost you can also use `if (process.env.TAMAGUI_TARGET === 'native')`

TheEhsanSarshar:
done!

packages/web/src/helpers/getSplitStyles.tsx
```diff
@@ -243,7 +249,49 @@ export const getSplitStyles: StyleSplitter = (
       if (keyInit === 'userSelect') {
         keyInit = 'selectable'
         valInit = valInit === 'none' ? false : true
-      } else if (keyInit.startsWith('data-') || keyInit.startsWith('aria-')) {
+      } else if (keyInit.startsWith('aria-*') || keyInit === 'role') {
+        if (webToNativeAccessibilityDirectMap[keyInit]) {
```
natew, line 253: if process.env here to avoid this loading on web

natew: i think this would error on web with null webToNativeAccessibilityDirectMap right?

nderscore:
Doesn't React Native 0.71 natively support `role` and `aria-*` attributes already? 🤔 

https://reactnative.dev/blog/2023/01/12/version-071#web-inspired-props-for-accessibility-styles-and-events

TheEhsanSarshar:
@nderscore yeah version > 0.71.1 does support it. values are not fully compatible with each other tho. for example if you pass slider instead of adjustable it will crash

PR #1286, opened 2023-06-13
fix(web): `accessibilityRole` not overwriting `role`

alitnk:
would fix the failing test after merge + release


alitnk:
> nice!
> I guess we need the same thing for aria-* as well. since native will convert aria-* to accessibility* we need to put both aria-* and accessibility* in usedKey when either of those is being used.
> and the same is true for the web.
> 
> <SomeComponent                             
>      aria-lable="blabla"
>      accessibilityLable="blabla"
> />

Yes, that sounds right. I was just trying to make the test pass so didn't look into native so feel free to go for it. Or let me know so I can take it later.

TheEhsanSarshar:
Awesome! I'll add that

PR #1290, opened 2023-06-13
fix(component theme): nested theme component now take correct theme a…

TheEhsanSarshar:
Nested component themes ignore their parent after switching light/dark mode.
fixed in this PR


natew:
wont we need to change the this.componentName as well if we do an updateState with a new componentName?

natew:
hm this code looks funky in general, componentName should be coming in from props already, definitely we'll have to figure out a better way to fix

TheEhsanSarshar:
@natew what you think about this way

PR #1291, opened 2023-06-14
sandbox example of bug

kaceycleveland:
Not sure if this is a bug actually or if I am misunderstanding when I should use `styleable`. 

For the most part, I used `styleable` in place of `forwardRef` where I did not have to type out the props as much but that is probably short sighted.

Anywho, the sandbox shows something I am dealing with where certain CSS properties are inconsistent between using in a wrapped component with `styleable` and without it. My actual usage is much more complex and a work around might be to use `forwardRef` instead but figure I would bring this up in case this should be working.

image


natew:
@kaceycleveland i realized we had some correctness issues because i'd left in a `test.only` in a test file and we regressed. this should be fixed now, i went back and cleaned things up and even improved it in a few places. let me know

PR #1339, opened 2023-06-26
fix margin vertical shorthand

benschac:
assuming this a typo, and should be `mv` not `my`

### DIFF
--- a/packages/shorthands/src/index.tsx
+++ b/packages/shorthands/src/index.tsx
@@ -64,7 +64,7 @@ export const shorthands = {
   mr: 'marginRight',
   mt: 'marginTop',
   mx: 'marginHorizontal',
-  my: 'marginVertical',
+  mv: 'marginVertical',
   o: 'opacity',
   ov: 'overflow',
   p: 'padding',


benschac:
* master
  * **PR #1339** <a href="https://app.graphite.dev/github/pr/tamagui/tamagui/1339" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/tamagui/tamagui/1339?utm_source=stack-comment).

la55u:
Given that `mx` also exists, I'm sure it's intentional

TheEhsanSarshar:
`y` in `my` means `y` axis just like `x` in `mx`

benschac:
that makes sense! Curious if adding mv and mh would make sense?

TheEhsanSarshar:
> that makes sense! Curious if adding mv and mh would make sense?

then need to have HStack instead of XStack and so on

benschac:
Ty that makes sense. Closing

Get Outlook for iOS<https://aka.ms/o0ukef>
________________________________

PR #1345, opened 2023-06-27
The ehsan sarshar create accordion component

natew:
cleaning up for merge

@TheEhsanSarshar just did some small cleanups.

Would be good to get final review from everyone @nandorojo @nderscore @dohomi @alitnk


alitnk:
Looks great, @TheEhsanSarshar! Some things that come to my mind after seeing the demo and docs - haven't gone through the code yet:
- Do the accordions work with `Group`? so that we can pass border radius to them as a group of accordions.
- I think the demo would look better if it was smaller or maybe if the triggers had rounded borders.
- Docs: We can probably get rid of the `npm install @tamagui/accordion` section as people using `tamagui` can just import it from `tamagui` without installing anything. Might be misleading and potentially result in unnecessary installs -> version mismatch between `tamagui` and `@tamagui/accordion`
- Maybe can use a better color for borders when focused. it's a bit unsettling/too bright on dark mode.
image

alitnk:
> Tests failing `Failed to create initial commit` maybe you can take a look? It should also use the branch we are on in order to test with this actual branch.

fixed @natew

packages/accordion/src/Accordion.tsx
```diff
@@ -1,491 +1,614 @@
...
+  const [value, setValue] = useControllableState({
+    prop: valueProp,
+    //@ts-ignore
+    defaultProp: defaultValue,
+    onChange: onValueChange,
+  })
+
+  React.useEffect(() => {
+    if (value && control) {
+      control([value])
```
nandorojo, line 164: Can this effect get removed in favor of firing this inside of onChange above? Or wrapping it in setValue?

packages/accordion/src/Accordion.tsx
```diff
@@ -1,491 +1,614 @@
...
+  const [value, setValue] = useControllableState({
+    prop: valueProp,
+    //@ts-ignore
+    defaultProp: defaultValue,
+    onChange: onValueChange,
+  })
+
+  React.useEffect(() => {
+    if (value && control) {
+      control(value)
```
nandorojo, line 227: Same comment as above.

packages/accordion/src/Accordion.tsx
```diff
@@ -1,491 +1,614 @@
...
+  }, [value])
+
+  const handleItemOpen = React.useCallback(
+    (itemValue: string) => setValue((prevValue = []) => [...prevValue, itemValue]),
+    [setValue]
+  )
+
+  const handleItemClose = React.useCallback(
+    (itemValue: string) =>
+      setValue((prevValue = []) => prevValue.filter((value) => value !== itemValue)),
```
nandorojo, line 238: Not a big deal, but I prefer to use `Set<string>` over `Array<string>` when managing state for performance.

packages/animate-presence/src/AnimatePresence.tsx
```diff
@@ -47,7 +47,7 @@ function onlyElements(children: ReactNode): ReactElement<any>[] {
   // We use forEach here instead of map as map mutates the component key by preprending `.$`
   Children.forEach(children, (child, index) => {
     if (isValidElement(child)) {
-      if (!child.key) {
+      if (!child.key && Children.count(children) > 1) {
```
nandorojo, line 50: Why is this change in this PR? What are the implications?

TheEhsanSarshar:
oh, awesome guys! thansk

packages/accordion/src/Accordion.tsx
TheEhsanSarshar: yea you're right, then I think we don't need that useAccordion hook anymore because it's basically doing the job of onChange. maybe we can add some helpful callbacks like `open()` but I think it's better to remove it.

doodle-codes:
Is there a projected date when this component is going to be released? The issue was opened sometime in Jan. and you are telling me it is taking you this long to develop such a simple component?

PR #1353, opened 2023-06-30
fix: more correct types on lucide-icons and helper-icons

alitnk:
individual lucide icons' props were untyped. added `SvgProps` type on icons and fixed the types with generate script and `helper-icons`. this fixes a type issue on takeout icons as well.


natew:
I think I just left SvgProps off IconProps, I fixed here:  f0362cdccb388bcb05f5b4b8d8acf6d398682867

PR #1384, opened 2023-07-04
typo

benschac:
lil typo

PR #1385, opened 2023-07-04
(feat): add debug="borders" prop

benschac:
https://discord.com/channels/909986013848412191/1125240744773423114/1125243555200376923
initial pass of a debug="borders" prop.

Open questions:
- This felt like the simplest way of implementing, not sure if it's the right way of tamagui the library and if I should be looking at some internals of the project. If internals, what files should I be looking at?

- Was going to update all View Components (Text, Spacer, Button, etc), just wanted to start here to make sure I'm on the right path.
- It working on kitchen sink ios, but the entire home screen isn't renedering. Not sure if that's normal or my change.
- https://github.com/tamagui/tamagui/blob/master/packages/components-test/Checkbox.web.test.tsx#L50 is currently failing tests locally, will double check if it's just something locally on my machine
CleanShot 2023-07-04 at 14 47 38@2x

image

### DIFF
--- a/packages/stacks/src/Stacks.tsx
+++ b/packages/stacks/src/Stacks.tsx
@@ -24,11 +24,13 @@ const variants = {
 } as const
 
 export const YStack = styled(Stack, {
+  name: 'YStack',
   flexDirection: 'column',
   variants,
 })
 
 export const XStack = styled(Stack, {
+  name: 'XStack',
   flexDirection: 'row',
   variants,
 })
@@ -36,6 +38,7 @@ export const XStack = styled(Stack, {
 export const ZStack = styled(
   YStack,
   {
+    name: 'ZStack',
     position: 'relative',
   },
   {
--- a/packages/web/src/createComponent.tsx
+++ b/packages/web/src/createComponent.tsx
@@ -599,30 +599,6 @@ export function createComponent<
       keepStyleSSR,
     } as const
 
-    if (process.env.NODE_ENV === 'development' && debugProp === 'borders') {
-      props.borderStyle = 'solid'
-      props.borderWidth = 0.5
-
-      const name = `${
-        componentName ||
-        Component?.displayName ||
-        Component?.name ||
-        '[Unnamed Component]'
-      }`
-
-      switch (name) {
-        case 'XStack':
-          props.borderColor = 'red'
-          break
-        case 'YStack':
-          props.borderColor = 'blue'
-          break
-        default: {
-          props.borderColor = 'green'
-        }
-      }
-    }
-
     const splitStyles = useSplitStyles(
       props,
       staticConfig,
--- a/packages/web/src/createComponent.tsx
+++ b/packages/web/src/createComponent.tsx
@@ -599,6 +599,30 @@ export function createComponent<
       keepStyleSSR,
     } as const
 
+    if (process.env.NODE_ENV === 'development' && debugProp === 'borders') {
+      props.borderStyle = 'solid'
+      props.borderWidth = 0.5
+
+      const name = `${
+        componentName ||
+        Component?.displayName ||
+        Component?.name ||
+        '[Unnamed Component]'
+      }`
+
+      switch (name) {
+        case 'XStack':
+          props.borderColor = 'red'
+          break
+        case 'YStack':
+          props.borderColor = 'blue'
+          break
+        default: {
+          props.borderColor = 'green'
+        }
+      }
+    }
+
     const splitStyles = useSplitStyles(
       props,
       staticConfig,
--- a/packages/web/src/helpers/getSplitStyles.tsx
+++ b/packages/web/src/helpers/getSplitStyles.tsx
@@ -69,6 +69,7 @@ import {
 } from './normalizeValueWithProperty'
 import { getPropMappedFontFamily, propMapper } from './propMapper'
 import { pseudoDescriptors, pseudoPriorities } from './pseudoDescriptors'
+import { updateDebugStyle } from './updateDebugStyle'
 
 const fontFamilyKey = 'fontFamily'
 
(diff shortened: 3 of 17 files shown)


benschac:
* master
  * **PR #1384** <a href="https://app.graphite.dev/github/pr/tamagui/tamagui/1384" target="_blank">Graphite</a> 
    * **PR #1385** <a href="https://app.graphite.dev/github/pr/tamagui/tamagui/1385" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/tamagui/tamagui/1385?utm_source=stack-comment).

apps/kitchen-sink/package.json
```diff
@@ -39,17 +39,17 @@
     "@tamagui/shorthands": "1.38.2",
     "@tamagui/themes": "1.38.2",
     "burnt": "^0.10.0",
-    "expo": "~48.0.15",
+    "expo": "~48.0.18",
```
benschac, line 42: ran kitchen sink, assuming this happens automatically.

apps/kitchen-sink/src/features/home/screen.tsx
```diff
@@ -31,7 +31,15 @@ export function HomeScreen() {
...
         </YGroup>
-        <YStack theme="yellow" bc="$background" p="$3" br="$4" bw={1} boc="$borderColor">
+        <YStack
+          theme="yellow"
+          debug="borders"
+          bc="$background"
+          p="$3"
+          br="$4"
+          bw={1}
+          boc="$borderColor"
```
benschac, line 41: should debug="borders" override border color?

packages/stacks/src/Stacks.tsx
```diff
@@ -23,16 +23,32 @@ const variants = {
...
+      <FrameXStack
+        ref={ref}
+        borderColor="red"
+        borderWidth={0.5}
+        borderStyle="solid"
+        {...props}
+      />
+    )
+  }
+  return <FrameXStack />
```
benschac, line 39: felt like the simplest / straight forward way

TheEhsanSarshar: I guess it's better to put it somewhere here
https://github.com/tamagui/tamagui/blob/master/packages/web/src/createComponent.tsx#L217
so if it's development and prop.debug == borders then merge { borderWidth: 0.5, borderColor: 'red' } with props

natew: yea basically can do it somewhere around there and it will only be one line (+ the if statement). also be sure to put behind a node_env conditional. you could put it right before `useSplitStyles` if you want to be sure it overrides.

natew:
https://twitter.com/reduzio/status/1677375378833408000?s=46&t=5wFlU_OsfjJ0sQPMFbtG0A

Visual render highlights would be cool

packages/web/src/createComponent.tsx
```diff
@@ -471,6 +471,30 @@ export function createComponent<
...
+      switch (name) {
+        case 'XStack':
+          props.borderColor = 'red'
+          break
+        case 'YStack':
+          props.borderColor = 'blue'
+          break
+        default: {
+          props.borderColor = 'green'
+        }
```
benschac, line 494: I'm assuming @natew has some opinions about border color, or if there should be any at all? 

IMO, it's a bit easier to grok what's going on when I know `XStack` is blue. If this is the route we choose, I'd be happy to update the docs too.

benschac:
> https://twitter.com/reduzio/status/1677375378833408000?s=46&t=5wFlU_OsfjJ0sQPMFbtG0A
> 
> Visual render highlights would be cool

That would be cool. I'd possibly be down to add it in a follow up PR

packages/web/src/createComponent.tsx
natew: I think its fine but media query styles will override these, same with pseudo styles, if we want to be really precise we should put this at the end of `getSplitStyles` before it returns the `style` object, and just force `style.borderColor` there, that way it always overrides.

natew:
lets just verify the performance doesn't change from adding `name`, profiling the homepage loading from scratch locally with `yarn site:prod` a few samples and averaging, something like LCP or something

packages/web/types/views/Slot.d.ts
```diff
@@ -5,6 +5,6 @@ interface SlotProps {
 export declare const Slot: import("react").ForwardRefExoticComponent<SlotProps & import("react").RefAttributes<any>>;
 export declare const Slottable: ({ children }: {
     children: ReactNode;
-}) => JSX.Element;
+}) => import("react/jsx-runtime").JSX.Element;
```
benschac, line 8: @natew -- assuming this is normal? I'm just running `watch:build` and `sandbox'

benschac:
> yarn site:prod

I'm assuming there's some kind of `git-crypt` / private package for `@tamagui/studio/api`. I've tried to build a few times now, removed node_modules and lock file and I'm getting.

```
Failed to compile.

./pages/api/studio/procedure.ts
Module not found: Can't resolve '@tamagui/studio/api'
```

apps/kitchen-sink/src/App.web.tsx
```diff
@@ -73,7 +73,7 @@ const SandboxFrame = (props: { children: any }) => {
       />
 
       <Theme name={screenshot ? 'blue' : undefined}>
-        <XStack w="100%" h="100%" fullscreen>
+        <XStack debug="borders" w="100%" h="100%" fullscreen>
```
natew, line 76: lets remove before merge

packages/web/src/helpers/getSplitStyles.tsx
```diff
@@ -1223,14 +1225,36 @@ export const getSplitStyles: StyleSplitter = (
...
+  }
+
   const result: GetStyleResult = {
     space,
     hasMedia,
     fontFamily: styleState.fontFamily,
     viewProps,
     // @ts-expect-error
     style,
-    pseudos,
```
natew, line 1233: !! this should be brought back!

packages/web/types/views/Slot.d.ts
natew: probably fine

packages/web/src/helpers/getSplitStyles.tsx
```diff
@@ -1223,14 +1225,36 @@ export const getSplitStyles: StyleSplitter = (
     }
   }
 
+  if (process.env.NODE_ENV === 'development' && debug === 'borders') {
+    const name = `${
+      componentName || Component?.displayName || Component?.name || '[Unnamed Component]'
+    }`
+
+    switch (name) {
+      case 'XStack':
```
natew, line 1234: really still not a fan of this because we avoid a decent amount of work if theres no theme changing props set, normally not a big deal but again these are super common views... perhaps we can check in some other way? Hard to find something truly simple... hm

What if you added a new helper 

```
updateDebugStyle(XStack, { borderColor: 'blue', borderWidth: 1 })
```

and we rename it to `debug="highlight"`

Then in `web` we can have that function be a no-op in production, otherwise it sets a WeakMap()

I really like this actually, it gives the user control.

Then in tamagui see `setup.js` you can add it there!

alitnk:
> > yarn site:prod
> 
> I'm assuming there's some kind of `git-crypt` / private package for `@tamagui/studio/api`. I've tried to build a few times now, removed node_modules and lock file and I'm getting.
> 
> ```
> Failed to compile.
> 
> ./pages/api/studio/procedure.ts
> Module not found: Can't resolve '@tamagui/studio/api'
> ```

#1644 should fix this

packages/web/src/helpers/getSplitStyles.tsx
benschac: That makes sense. Ty, for the context.

benschac: Yeah, quick visual debug goes a long way. long live

* {
  borders: 1px solid red
}

benschac:
closing this in favor of @natew's debugging solution.

PR #1407, opened 2023-07-10
fix(select): not working when animation is enabled

TheEhsanSarshar:


packages/select/src/SelectImpl.tsx
```diff
@@ -112,7 +112,6 @@ export const SelectInlineImpl = (props: SelectImplProps) => {
 
   const { x, y, strategy, context, refs, update } = useFloating({
     open,
-    onOpenChange: setOpen,
```
TheEhsanSarshar, line 115: after removing this line, everything starts to work and I don't see any adverse side effects.
I think it has to do something with AnimatePresence when `Select.Viewport` has animation and enterStyle

TheEhsanSarshar: yeah, that works, it actually doesn't have any adverse side effect.

natew: Hm then what is setOpen doing? I'd have assumed we need to keep that in sync or else Bad Things happen, but if not maybe we aren't even using it in other places? Sounds off though, we should see where we use `open` in context. I'd need to understand how that is staying in sync if not here.

TheEhsanSarshar: open actually changes, so there is no need to recall it here as well

natew: it does break btw, for example hitting "escape" no longer closes it, floating needs a way to set that state.

also very odd this would break things since if its not changing open it should be a no-op, but if it is its necessary?

PR #1427, opened 2023-07-15
fix copy code in hero

suleymanbariseser:
Copy button in hero component does not work. Content of code block is gathered by query selector. However, this query runs only on the first mount but code block is not rendered in hero at that time becuase it is a collapsible block. We should get the content whenever code block is visible.


natew:
ty

PR #1454, opened 2023-07-20
fix(animations-reanimated) add deprecation warning

nandorojo:


la55u:
IMO a deprecation shouldn't show an error even if it's just a console log. Maybe `console.warning` and a `@deprecated` JSDoc comment so the IDE can pick it up?

nandorojo:
deprecation is an understatement: the reanimated driver crashed our app

la55u:
> deprecation is an understatement: the reanimated driver crashed our app

Oh sorry didn't know that

PR #1459, opened 2023-07-21
fix: add babel plugin details to setup doc

dannyhw:
The docs tell you to add the babel plugin to the babel config but without telling you how. Added in an example from the example app

### DIFF
--- a/apps/site/data/docs/intro/installation.mdx
+++ b/apps/site/data/docs/intro/installation.mdx
@@ -144,7 +144,7 @@ module.exports = {
     [
       "@tamagui/babel-plugin",
       { components: ["tamagui"], config: "./tamagui.config.ts" },
-     ],
+    ],
   ]
 }
 ```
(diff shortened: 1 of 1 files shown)


natew:
We should also add a note though that:

1. Native can't optimize nearly as much due to needing to de-opt whenever it encounters a token value, because RN doesn't support something like CSS variables
2. Because of that, setting up the compiler for native may be more trouble than it's worth, but this is how you can

dannyhw:
@natew I think maybe more rewording is needed then because when I saw this text in the installation docs I got the impression that the babel plugin is required. Though what I'm understanding from what you said is that actually the plugin is strictly optional.

image

perhaps instead of 

> Add @tamagui/babel-plugin to your babel.config.js plugins:

it should say 

add 

> add 'transform-inline-environment-variables' to your babel.config.js plugins

And then a further section that explains how you can add the @tamagui/babel-plugin for further optimization and add those caveats about it not being as valid on native.

natew:
ended up making those changes, so i'll close this, but really appreciate the heads up here

PR #1460, opened 2023-07-21
fix(RadioGroup): RadioGroupItemFrame uses active theme when checked

hackily:
Enabling active theme lets the RadioGroupItemFrame be themed differently than the base component when checked.


natew:
will revisit in v2

PR #1462, opened 2023-07-22
site: Expo guide improvements

LunatiqueCoder:
Hello.

Following multiple discussions in discord, I was getting tired of copy pasting this solution to everyone. I believe it would be easier to just add it to the docs.


natew:
i think we can even leave out the "include" part i'll do that as a follow on

LunatiqueCoder:
It didn’t work for me with expo router other wise. I can prepare a reproduction if needed. Cheers

PR #1480, opened 2023-07-28
Canary

natew:


tobycodes:
I don't know if it is this merge but `npm create tamagui@latest` keeps breaking.

image

PR #1490, opened 2023-08-01
fix: expose possible null value in createStyledContext types

nderscore:
The implementation of `createStyledContext` produces a context that has a default value of `null`.

But the types applied to the context report that it will always have a value.

This PR adjusts the types to more accurately represent the possible value of the context.

I also added some null guards to all the components which were directly accessing a styled context with `React.useContext`

### DIFF
--- a/packages/checkbox/src/Checkbox.tsx
+++ b/packages/checkbox/src/Checkbox.tsx
@@ -32,6 +32,16 @@ export const CheckboxStyledContext = createStyledContext({
   scaleIcon: 1,
 })
 
+const useCheckboxStyledContext = () => {
+  const value = React.useContext(CheckboxStyledContext)
+  if (value === null) {
+    throw new Error(
+      'useCheckboxStyledContext must be used within a CheckboxStyledContext.Provider'
+    )
+  }
+  return value
+}
+
 export type CheckedState = boolean | 'indeterminate'
 
 export function isIndeterminate(checked?: CheckedState): checked is 'indeterminate' {
@@ -140,7 +150,7 @@ const CheckboxIndicator = CheckboxIndicatorFrame.extractable(
         ...indicatorProps
       } = props
       const context = useCheckboxContext(INDICATOR_NAME, __scopeCheckbox)
-      const styledContext = React.useContext(CheckboxStyledContext)
+      const styledContext = useCheckboxStyledContext()
       const iconSize =
         (typeof styledContext.size === 'number'
           ? styledContext.size * 0.65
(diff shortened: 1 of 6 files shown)


packages/checkbox/src/Checkbox.tsx
```diff
@@ -288,8 +298,7 @@ const CheckboxComponent = CheckboxFrame.extractable(
       onChange: onCheckedChange,
     })
 
-    // TODO: this could be null - fix the type
```
nderscore, line 291: I'm hoping this was the expected fix for this comment 🤞

packages/checkbox/src/Checkbox.tsx
```diff
@@ -32,6 +32,16 @@ export const CheckboxStyledContext = createStyledContext({
   scaleIcon: 1,
 })
 
+const useCheckboxStyledContext = () => {
+  const value = React.useContext(CheckboxStyledContext)
+  if (value === null) {
+    throw new Error(
+      'useCheckboxStyledContext must be used within a CheckboxStyledContext.Provider'
```
natew, line 39: we should make these abstracted i think so we dont have all the code duplication

nderscore: How about if `createStyledContext` also created a use hook as part of the `StyledContext`? 

```tsx
const SomeStyledContext = createStyledContext({ foo: true, bar: true })
const useSomeStyledContext = SomeStyledContext.useContext
// ...
const { foo, bar } = useSomeStyledContext()
// OR
const { foo, bar } = SomeStyledContext.useContext()
```

Now I'm wondering if the original props passed into `createStyledContext()` should be returned when the context value is `null`? 🤔 (instead of throwing an error)

```tsx

export function createStyledContext<VariantProps extends Record<string, any>>(
  props: VariantProps
): StyledContext<VariantProps> {
  // ...
  const useHook = () => {
    const value = useContext(StyledContext.context)
    if (value === null) {
      return props
    }
    return value
  }
  // ...
  Context.useContext = useHook

  return Context
}
```

natew:
@nderscore left a comment

packages/checkbox/src/Checkbox.tsx
natew: i think the value would never be null right? i thought react also errored if you use a context that isnt provided?

but if not then yea both seem right, can just `?? props`

nderscore: @natew I just created a reproduction of a case where the styled context stays `null` here:

https://github.com/nderscore/tamagui-styled-context-null

If that's unexpected, maybe it's a bug with boolean props?

If that _is_ the expected behavior though... then maybe the use hook should spread the value of the context into the default props instead:

```tsx
const useHook = () => {
  return {
    ...props,
    ...React.useContext(StyledContext.context)
  }
}
```

natew: i think the default should definitely be there and spread down so thats a good fix for us to make

nderscore:
After spending some more time with the implementation inside `createComponent`, I'm pretty sure this is actually a bug. There should always be a context provider with a value.

I'm closing this PR in favor of the fix in #1501

PR #1495, opened 2023-08-03
sandbox(web): add some changes sandbox (padding, center paragraph & remove &nbsp;)

germanolira:
- Removed an unused import.

- Centered the paragraph and the square related to `useMedia().gtSm`.

- Eliminated the (runs faster) as it caused a break in the word 'faster' (fas-ter).

- Changed the paragraph's positioning from `absolute` to `relative` to keep it centered and prevent overflow.


natew:
this is just a random testing area we play with all the time, it gets messy, but thanks for the pr :)

PR #1501, opened 2023-08-04
fix(core): make styled context handle boolean props

nderscore:
## Background

This is a fix for a bug I encountered when using a styled context which has only boolean props, with `false` as their default value.

The code inside `createComponent` which determines whether a prop is set was using `||` to short-circuit. Depending on the order they're defined in, this swallows `false` values unexpectedly, because `false || undefined` resolves to `undefined`.

When a context prop is `false`, it is being excluded from the context value. When all the context props are `false`, the context provider is never rendered at all.

A reproduction of this bug can be found [here](https://github.com/nderscore/tamagui-styled-context-null)

## Changes

* Replace the short-circuiting OR (`||`) in `createComponent` with nullish-coalescing operators (`??`)
* Merge parent styled context value into overridden props when rendering a nested context provider
  - This way its not possible for a context provider to be missing any props

### DIFF
--- a/packages/web/src/createComponent.tsx
+++ b/packages/web/src/createComponent.tsx
@@ -167,16 +167,17 @@ export function createComponent<
     // order is after default props but before props
     let styledContextProps: Object | undefined
     let overriddenContextProps: Object | undefined
+    let contextValue: Object | null | undefined
     const { context } = staticConfig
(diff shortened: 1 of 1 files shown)


natew:
beautiful, thanks

PR #1508, opened 2023-08-04
update token.mdx 

jocanola:


natew:
not sure why reverted but good catch, made the fix on master

PR #1511, opened 2023-08-06
Fix selectItemIndicator not behaving correctly

TheEhsanSarshar:
selected item lose its selected state after being unmounted ( select closed ) and reset it back to initial value
`isSelected = React.useState(initialValue === value)`
trigger that `emitter` again once Select open


TheEhsanSarshar:
oh we don't need to use `useMemo` my silly code haha
it fix that SelecteItemIndicator issue but I guess it's broken the initial value intention here

PR #1524, opened 2023-08-09
fix(themed): not handling tokens

TheEhsanSarshar:
`<Icon color="$token"/>` was broken because `themed` HOC was only handling theme variables


natew:
themes should fall back to tokens, if not we have a bug in that system, useTheme() should generally fall back to tokens. i'll check that out

natew:
actually i have a test case that proves this should work already, see UseTheme.tsx test case:

CleanShot 2023-08-09 at 13 39 47@2x

CleanShot 2023-08-09 at 13 40 08@2x

Issue #1491, opened 2023-08-02
Font weight for Android are not using the correct font family


idrakimuhamad:
### Current Behavior

Bold text with custom font didn't use the correct font family. It work fine in iOS. The font uses are Lato, using the `@expo-google-fonts/lato`

Left is Android. Right is iOS. You might notice it based on the kerning of the small case "T", and letter "S".

![Screenshot 2023-08-02 at 12 41 05 PM](https://github.com/tamagui/tamagui/assets/1400323/46c82306-4bf8-4a10-aa1f-6756a406154e)

Bottom text uses the correct font family.

### Expected Behavior

It should show the correct font family regardless of the weight as long as set in the `createFont`

This work correctly in previous version before I upgrade v1.39.6.

### Tamagui Version

```markdown
latest
```

### Reproduction

```markdown
https://github.com/idrakimuhamad/repro-tamagui
```

### System Info

```markdown
System:
    OS: macOS 14.0
    CPU: (8) arm64 Apple M1
    Memory: 1.01 GB / 16.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 18.12.0 - ~/.nvm/versions/node/v18.12.0/bin/node
    Yarn: 3.5.0 - ~/.yarn/bin/yarn
    npm: 8.19.2 - ~/.nvm/versions/node/v18.12.0/bin/npm
    Watchman: 2023.05.22.00 - /opt/homebrew/bin/watchman
  Browsers:
    Chrome: 115.0.5790.114
    Firefox: 105.0.3
    Safari: 17.0
```

rranjan14:
Facing same issue. For `fontWeight` above `700`, font family is not being applied and the system default is being used.

natew:
are you using the `face` option? its needed for android: https://tamagui.dev/docs/core/configuration#overview

idrakimuhamad:
> are you using the `face` option? its needed for android: https://tamagui.dev/docs/core/configuration#overview

Of course! If you see the reproduction repo:

![image](https://github.com/tamagui/tamagui/assets/1400323/78a96543-ba0f-4d2e-ac7b-a826c2281db9)

As mentioned, this is a working configuration prior to 1.39.x.

real-marshal:
Same thing, 700 fontWeight doesn't use the correct font, even though it's set in the face key.

alitnk:
[...]

PR #1541, opened 2023-08-16
reproduce #1491 on kitchen-sink

alitnk:
Just the reproduction of the issue on kitchen sink


natew:
issue #1491

alitnk:
Continue: #1558

PR #1554, opened 2023-08-20
docs: fix wrong transform style in compiler media example

DaniHv:
This PR fixes wrong style output in compiler > about docs ["How Tamagui Helps"](https://tamagui.dev/docs/intro/why-a-compiler#how-tamagui-helps), where the given example in "media" tab shows a prop-based (shrink prop) style with transform scale property. Currently the output shows a wrong background style inside a gtSm media query, with this modification it will show the corresponding transform property inside a sm media query, with a tamagui-generated class name (generated using simple-web starter).


natew:
thanks for catching that and helping

PR #1565, opened 2023-08-23
fix(build): enable tsProject parameter to allow integrated monorepos …

jcbdev:
…support (nx)

Integrated monorepos such as NX use tsconfig files to separate projects (and jest) instead of package.json's.  This means that tamagui-build fails to find the correct the tsconfig file and you cannot specify it manually.

Adding this parameter `--ts-project` to the build tool allows you to use tamagui-build inside an nx monorepo but doing the following in the project.json:
```
"build": {
      "executor": "nx:run-commands",
      "options": {
        "command": "tamagui-build --ts-project ./tsconfig.lib.json ",
        "cwd": "libs/ui"
      }
    },
    "watch": {
      "executor": "nx:run-commands",
      "options": {
        "commands": [
          {
            "command": "tamagui-build --ts-project ./tsconfig.lib.json --watch",
            "cwd": "libs/ui"
          }
        ]
      }
    }
```

### DIFF
--- a/packages/build/tamagui-build.js
+++ b/packages/build/tamagui-build.js
@@ -23,8 +23,9 @@ const shouldWatch = process.argv.includes('--watch')
 const declarationToRoot = !!process.argv.includes('--declaration-root')
 const ignoreBaseUrl = process.argv.includes('--ignore-base-url')
 const baseUrlIndex = process.argv.indexOf('--base-url')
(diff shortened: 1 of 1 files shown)


natew:
looks good, thank you

adamkrim:
@jcbdev Do you have an example repository with nx and tamagui working well together ?
I struggle to setup it

PR #1581, opened 2023-08-25
fix(core): `selectable` not working on `<Text>` as expected

### LINKED ISSUE #1536: Creating selectable text
I am trying to create a selectable text component where the user is able to copy/paste/etc.

Consider the following
```js
// import { Text } from "react-native"; // this works!
// import { Text } from "tamagui"; // this doesn't work...

<Text selectable={true}>asdasd</Text>
```
Passing in `selectable` prop does not achieve same functionality when using Tamagui's `Text` component, however it works in React Native's component. Testing on Expo Go Android.

How should I implement selectable text using Tamagui's Text component?

alitnk:
Fixes #1536 Creating selectable text

### DIFF
--- a/packages/web/src/helpers/getSplitStyles.tsx
+++ b/packages/web/src/helpers/getSplitStyles.tsx
@@ -294,6 +294,13 @@ export const getSplitStyles: StyleSplitter = (
         if (keyInit === 'elevationAndroid') continue
       }
 
+      // we have a custom `selectable` variant on tamagui's <Text> component but `selectable` is also a native Text prop
+      // here we make sure the native prop gets through
+      if (keyInit === 'selectable' && isText) {
+        viewProps[keyInit] = valInit
+        continue
+      }
+
       // map userSelect to native prop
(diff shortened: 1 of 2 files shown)


mphill:
FYI, It seems like all text is selectable by default now.  I had to explicitly disable it.

natew:
Oh I thought that was the default but I guess it makes sense for native not to. I’ll make that change.

mphill:
I agree, it's definitely not.  

I noticed this because the `onLongPress` gesture was getting very hard to trigger.  Setting this to false should resolve that issue too so long presses will work on buttons etc  better.

natew:
Hm I don't see us setting selectable on Text at all

mphill:
> Hm I don't see us setting selectable on Text at all

I didn't either when I looked at the source.  I can definitely say that setting selectable to false gets the desired behavior. Can this be the default false?

https://reactnative.dev/docs/text#selectable

The default should be false according to the docs.

PR #1594, opened 2023-08-30
fix(Label): Label is not supposed to have a role

hackily:
According to WAI-ARIA 4, Document conformance requirements for use of ARIA attributes in HTML
https://www.w3.org/TR/html-aria/#docconformance

label role is prohibited and must not have a role, so the role should be removed.


natew:
thank you

PR #1604, opened 2023-09-01
feat(sheet): add option for constant snapPoint sizes in pixels

nderscore:
## Description

Adds a new prop to the `Sheet` component called `snapPointsMode`, which changes the behavior of the `snapPoints` prop. 

The possible values for this prop are:

- `percent` - the default, treat snap points as numeric percentages
- `constant` - treat snap points as constant pixel dimensions
- `fit` - auto-fit to the content's inherent height
- `mixed` - a mixture of multiple types (`numbers` = pixels, `"${number}%"` = percentages, `fit` = fit)

If a snap point value is greater than the height of the parent container, it will be constrained to the height of the parent container.

When in non `percent` modes, an extra View is rendered to measure the height of the parent container in order to get the correct `screenHeight` without being able to do percentage math on the sheet's height.

When an `fit` mode, or when `fit` is one of the mixed snap points, the `frameSize` layout handler moves to the `SheetFrame`.

The names of the `constant` and `pixel` prop values are also aligned with [react-native-ios-modal](https://github.com/dominicstop/react-native-ios-modal/blob/main/src/types/RNIComputable/RNIComputableSize.ts).

### DIFF
--- a/packages/sheet/src/SheetImplementationCustom.tsx
+++ b/packages/sheet/src/SheetImplementationCustom.tsx
@@ -66,6 +66,7 @@ export const SheetImplementationCustom = themeable(
       setFrameSize,
       snapPoints,
       snapPointsMode,
+      hasFit,
       position,
       setPosition,
       scrollBridge,
@@ -89,8 +90,11 @@ export const SheetImplementationCustom = themeable(
     }, [])
 
     const positions = useMemo(
-      () => snapPoints.map((point) => getYPositions(point, screenSize, snapPointsMode)),
-      [screenSize, snapPoints, snapPointsMode]
+      () =>
+        snapPoints.map((point) =>
+          getYPositions(snapPointsMode, point, screenSize, frameSize)
+        ),
+      [screenSize, frameSize, snapPoints, snapPointsMode]
     )
 
     const driver = useAnimationDriver()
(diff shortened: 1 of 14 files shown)


packages/sheet/src/SheetImplementationCustom.tsx
```diff
@@ -314,6 +316,23 @@ export const SheetImplementationCustom = themeable(
...
     )
 
+    const handleMaxContentViewLayout = useCallback(
+      (e: LayoutChangeEvent) => {
+        const next = (() => {
+          let _ = e.nativeEvent?.layout.height
+          if (isWeb && isTouchable && !open) {
+            // temp fix ios bug where it doesn't go below dynamic bottom...
+            _ += 100
+          }
```
nderscore, line 326: Not sure if this is needed - copied it from the other layout handler 😅

nderscore: Removed it. It just pushes the sheet off screen on mobile.

packages/sheet/src/SheetImplementationCustom.tsx
```diff
@@ -87,8 +89,8 @@ export const SheetImplementationCustom = themeable(
     }, [])
 
     const positions = useMemo(
-      () => snapPoints.map((point) => getPercentSize(point, screenSize)),
-      [frameSize, snapPoints]
+      () => snapPoints.map((point) => getYPositions(point, screenSize, snapPointsMode)),
+      [frameSize, screenSize, snapPoints, snapPointsMode]
```
nderscore, line 93: `frameSize` might not be needed in the deps here anymore

PR #1610, opened 2023-09-02
docs(sheet): document snapPointsMode prop

nderscore:
Adding docs for #1604

### DIFF
--- a/apps/site/data/docs/components/sheet/1.59.0.mdx
+++ b/apps/site/data/docs/components/sheet/1.59.0.mdx
@@ -1,0 +1,298 @@
+---
+title: Sheet
+description: A simple sheet component
+name: sheet
+component: Sheet
+package: sheet
+demoName: Sheet
+---
+
+# Sheet
+
+<Description>A bottom sheet that slides up.</Description>
+
+<HeroContainer showAnimationDriverControl>
+  <SheetDemo />
+</HeroContainer>
+
+```tsx hero template=Sheet
+
+```
+
+<Highlights
+  features={[
+    `Lightweight implementation with dragging support.`,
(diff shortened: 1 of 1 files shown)


apps/site/data/docs/components/sheet/1.59.0.mdx
```diff
@@ -0,0 +1,298 @@
...
+```tsx
+<Sheet snapPoints={[85, 50]} /> // 85% and 50%
+```
+
+The behavior of snap points can be changed by setting the `snapPointsMode` prop to any of these values:
+
+- **percent** (default) - Snap points are treated as percentages of the parent container or screen
+- **constant** - Snap points are treated as raw pixel values
+- **fit** - Snap points are ignored and the sheet is constrained to the content height
+- **mixed** - Snap points can be either numbers (pixels), percentages as strings (ex: `"50%"`), or `"fit"` for fit behavior
```
nderscore, line 66: This section is new

apps/site/data/docs/components/sheet/1.59.0.mdx
```diff
@@ -0,0 +1,298 @@
...
+      type: '(number | string)[]',
+      default: `[80]`,
+      description: `Array of values representing different sizes for the sheet to snap to. Values should be ordered most visible to least visible. Use "open" prop for fully closed.`,
+    },
+    {
+      name: 'snapPointsMode',
+      type: '"percent" | "constant" | "fit" | "mixed"',
+      default: '"percent"',
+      description: `Alters the behavior of the 'snapPoints' prop. See docs above for usage information.`
+    },
```
nderscore, line 189: And this changed too

Issue #1272, opened 2023-06-09
npm create tamagui fails if yarn is unavailable


xizhao:
### Current Behavior

```
> yarn install
[tamagui] error installing with yarn
 spawn yarnpkg ENOENT
node:child_process:960
    throw err;
    ^
```

### Tamagui Version

```markdown
latest
```

### Reproduction

```markdown
try it out
```

PR #1635, opened 2023-09-11
Fix for create-tamagui fail to detect proper package manager

Carbowix:
This pull request fixes #[1272](https://github.com/tamagui/tamagui/issues/1272). I also encountered an issue while running the program where it failed when cloning the tamagui master repo. since it spits out "fatal error: too many arguments", I had to double-quote the local directory to handle spaces in paths


natew:
amazing, thank you

PR #1647, opened 2023-09-14
fix(docs): Fix typos

hackily:


natew:
damn quite a few there, thanks for the fix

Issue #1422, opened 2023-07-13
Documentation missing for `Configuring Tokens`


arjanwit:
### Current Behavior

The documentation is missing for the following section in the docs - https://tamagui.dev/docs/core/configuration#configuring-tokens

### Expected Behavior

[### Strict tokens]

In documentation for version 1.38.4
(https://github.com/tamagui/tamagui/blob/v1.38.4/apps/site/data/docs/core/configuration.mdx)

### Tamagui Version

```markdown
1.39.0
```

### Reproduction

```markdown
See above
```

natew:
It's there for me, I think we merged a fix, thanks

PR #1652, opened 2023-09-15
fix(docs): Revert removal of strict token documentation

hackily:
Should resolve https://github.com/tamagui/tamagui/issues/1422


natew:
these docs are out of date but thanks for the catch, i'll fix it with the right docs for the new `settings` options in createTamagui

PR #1654, opened 2023-09-17
Kacey/input adornment

kaceycleveland:
image

Needs some more work and there is a lot of room to expand here but wanted to get an initial PR out to get some feedback on  approach. The overall change here is focus management is controlled by a inner `Input` component and then controls the styling of the container element based on focus and blur. This allows for some generalizing of other input components (i.e. select) down the line where we can adopt similar adornment patterns for those ones. We do this already at my job and it works well where there is a lot of shared logic between `Select` and `Input`.

I have tested a couple things to ensure there is at least no regressions compared to whats on the site:
- Tabbing through and focus management
- Safari tabbing through and focus management
- Placeholders

Is there anything else significant that I should test?

Some things that I am not sure on:
- Button adornment: 
  - Should these share some context with the input component itself? If an `Input` is disabled, should the button be disabled as well?
  - Styling for the button adornments (and regular adornments as well)
- I imagine this can probably be moved to its own package now; safe to do in a future PR or should I include that in this one?
- There is some grouping of components together via `styleable` which works fine for `tamagui` but I am thinking to make it more headless, these can be split out a bit more.

I have not tried or done anything on running the site/docs locally; I would still need to update docs to include docs on adornments and usage. Any thoughts on this?


natew:
This probably need significant reworking:

1. Mapping over children and checking them is an anti-pattern in React, because now you can't abstract them. For example if I put an adornment inside a parent component, it will break, which is a common pattern. For this reason you'd need to have some sort of context based solution.
2. Not a fan of throwing Button in there, as we want to try and keep all the components pretty modular so people can opt-in to as many or few as they want. Instead it should be easy for a user to just grab a Button (or any other component) and somehow wrap it so it works with this.
3. We have the concept of Group which handles these sorts of border radius things, I think we'd want to use that internally.
4. This adds a decent bit of depth to the render tree if you're not actually using adornments, it's now costing a bit more in performance, which is not great. Maybe we'd have it as a whole new thing like `<SegmentedInput />`, or there'd need to be a `<Input acceptsDecorations />` or something.

It's definitely something nice to get in and really appreciate the effort, we may just have to workshop it a bit.

kaceycleveland:
> This probably need significant reworking:
> 
> 1. Mapping over children and checking them is an anti-pattern in React, because now you can't abstract them. For example if I put an adornment inside a parent component, it will break, which is a common pattern. For this reason you'd need to have some sort of context based solution.
> 2. Not a fan of throwing Button in there, as we want to try and keep all the components pretty modular so people can opt-in to as many or few as they want. Instead it should be easy for a user to just grab a Button (or any other component) and somehow wrap it so it works with this.
> 3. We have the concept of Group which handles these sorts of border radius things, I think we'd want to use that internally.
> 4. This adds a decent bit of depth to the render tree if you're not actually using adornments, it's now costing a bit more in performance, which is not great. Maybe we'd have it as a whole new thing like `<SegmentedInput />`, or there'd need to be a `<Input acceptsDecorations />` or something.
> 
> It's definitely something nice to get in and really appreciate the effort, we may just have to workshop it a bit.

I don't mind spending a decent amount of time on this as a lot of it is inspired by implementation we use on our own design system using tamagui core so appreciate all the feedback :)

1. Hmm I will have to poke around with this. I get your reasoning in your example but for our design system it would never make sense to use the component in any other way other then the way its used here so I feel like mapping over the `children` is ok in that sense. With that said though, with an open source design system like `tamagui`, I can get behind the idea of having it more flexible and not requiring the adornments to be limited to one depth level down using some context based thing. Not sure what that will end up looking like but I will experiment.

2. I can take it out for sure. There are some specific stylings needed on buttons when they are an adornment hence why I wanted it as part of the input component (`Input.Start.Button` vs just `Button`). I think if I can figure out a way to do it without mapping over `children`, I can do this as well and just have `Button`.

3. Will look into this, have not used the grouping stuff much since it was first introduced.

4. I like keeping `Input` as a sole component if possible but I do think there are ways to reduce the render tree. Will poke around this as well

kaceycleveland:
@natew I simplified a lot of the logic here based on your feedback and leave a lot more control to the user as well.

`Input` by itself should now have a much smaller footprint on the DOM when not using adornments and now the user has control of where to place the control element of the `Input`:

```tsx
<Input>
<Input.Start>
<Button>Start Button</Button>
</Input.Start>
<Input.Control/>
<Input.End>
<Button>End Button</Button>
</Input.End>
</Input>
```

I am still not quite sure on how to handle providing child elements styles while keeping `Input` specific logic out of their implementations. For example, I need to modify the height and border radius of child elements based on if they are within an `Input` but I don't want to add that logic to the components themselves.

I can always plug in some extra styles via cloning elements and I think I could do it in a generalized way too to make it fairly robust since all components from my understanding would support plugging in `borderRadius` values. Do you think that would be the right approach? It sorta goes back to the idea of mapping through children in a way but not as bad I guess. 

image

packages/tamagui/src/views/Input.tsx
```diff
@@ -72,14 +156,148 @@ export const InputFrame = styled(
...
+
+const InputMain = InputFrame.styleable<InputProps>((propsIn, ref) => {
+  const props = useProps(propsIn)
+  const [isFocused, setIsFocused] = useState(false)
+  const textInputRef = useRef<TextInput>(null)
+  const hasContent = props.children
+
+  const handleOnPress = useMemo(
+    () => composeEventHandlers(props.onPress, () => textInputRef.current?.focus()),
+    []
```
natew, line 276:
```suggestion
    [props.onPress]
```

packages/tamagui/src/views/Input.tsx
```diff
@@ -72,14 +156,148 @@ export const InputFrame = styled(
...
+  )
+  const { onPress: _, ...rest } = props
+
+  useEffect(() => {
+    if (propsIn.disabled) setIsFocused(false)
+  }, [propsIn.disabled])
+
+  if (!hasContent) return <InputControl unstyled={false} {...props} />
```
natew, line 284: seems wrong, why set unstyled={false}? it should be default set through variants?

packages/tamagui/src/views/Input.tsx
```diff
@@ -72,14 +156,148 @@ export const InputFrame = styled(
...
+
+  return (
+    <InputEndAdornmentMain ref={ref} {...props}>
+      {children}
+    </InputEndAdornmentMain>
+  )
+})
+
+const InputMain = InputFrame.styleable<InputProps>((propsIn, ref) => {
+  const props = useProps(propsIn)
```
natew, line 269: this is a part of the reason i worry about adding a useProps + styleable... together they are a bit expensive.

i wonder if we shouldn't leave `Input` alone... and then we can add `Input.Decorated` and have that handle all of this?

packages/tamagui/src/views/Input.tsx
```diff
@@ -72,14 +142,131 @@ export const InputFrame = styled(
...
+
+// const baseStyles = {
+//   height: '100%',
+//   borderTopRightRadius: 0,
+//   borderBottomRightRadius: 0,
+//   borderTopLeftRadius: 0,
+//   borderBottomLeftRadius: 0,
+// }
+
+// const InputStartAdornment = InputStartAdornmentMain.styleable((propsIn, ref) => {
```
kaceycleveland, line 191: @natew I took this logic out but left as a comment because I wanted to see what your thoughts were. I sorta convinced myself this is a bad idea already but essentially the idea was to clone the element and put some additional styling on the adornments such that they fit a bit nicer in the input. Instead, I would opt for the consumer to do the styling themselves.

natew:
@kaceycleveland sorry this has gotten a bit stale, I'd like to come back to it and get it merged. Let me know your thoughts on latest status. I can clean up some stuff.

wallacerenan:
Any news about? that would be very interesting

kaceycleveland:
@wallacerenan 

I can take a look this weekend; I remember I left in a decent place but with all the changes in the past 2 months, will definitely need to revisit.

@natew 

I will update and tag you when its in a good spot for you to take a look again, maybe sometime next week.

natew:
Closing but have a note on v2 branch to try and get this in

PR #1678, opened 2023-09-21
Feat/new scoped context

TheEhsanSarshar:
add scope feature to the createStyledContext 
make the following components to support the scoped context
- Popper
- Popover
- Tooltip


packages/web/src/helpers/createStyledContext.tsx
```diff
@@ -7,38 +7,69 @@ export type StyledContext<Props extends Object = any> = Omit<
...
         ...values,
       }
     }, [objectIdentityKey(values)])
-    return <OGProvider value={value}>{children}</OGProvider>
+    if (scope) {
+      const ScopedContext = scopedContext.get(scope)
+      if (ScopedContext) {
+        return <ScopedContext.Provider value={value}>{children}</ScopedContext.Provider>
+      } else {
+        const NewlyCreatedScopedContext = createContext<VariantProps | undefined>(
```
natew, line 45: isnt this going to create a new context every render??

natew, line 45: this will create a new context every render (!!) meaning huge performance hit, causing every sub component that uses it to re-render on every render of the parent. We need to have this hoisted outside and only created once.

natew: nevermind me again, i see its being set and used based on scope.

PR #1692, opened 2023-09-22
Feat/make some components to use scope context

TheEhsanSarshar:
Make the following components to use the new context scope API

Collection
Collapsible
Toast
Accordion
ToggleGroup
RovingFocusGroup
Tab


DVGY:
What is the use of context scope ?

PR #1698, opened 2023-09-24
fix(switch): avoid needing border width for distance calculation

nderscore:
This PR implements an alternative to using border width in the calculations for moving the Switch component's Thumb.

Instead of measuring the frame directly, an additional view is rendered inside the frame.


natew:
beautiful i knew there had to be a way

PR #1712, opened 2023-09-26
fix(switch): export SwitchContext for headless consumers

nderscore:
Adds a missing export for `SwitchContext` on the `@tamagui/switch` package. 

Previously this was exported from another file.

### DIFF
--- a/packages/switch/src/index.ts
+++ b/packages/switch/src/index.ts
@@ -2,6 +2,7 @@ import { createSwitch } from './createSwitch'
 import { SwitchFrame, SwitchThumb } from './Switch'
 
 export * from './Switch'
+export * from './SwitchContext'
 export * from './createSwitch'
 
 export const Switch = createSwitch({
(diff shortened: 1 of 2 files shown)


natew:
oh nice, btw i think switch adds this context for you automatically when you call createSwitch, but still this is probably useful to export

nderscore:
Oh! Maybe the docs for the headless switch should be updated instead then.

I actually see these warnings when I use the `SwitchContext` context in a custom component.

```tsx
    if (Frame !== DefaultSwitchFrame && Frame.staticConfig.context) {
      console.warn(
        `Warning: createSwitch() needs to control context to pass checked state from Frame to Thumb, any custom context passed will be overridden.`
      )
    }
    if (Thumb !== SwitchThumb && Thumb.staticConfig.context) {
      console.warn(
        `Warning: createSwitch() needs to control context to pass checked state from Frame to Thumb, any custom context passed will be overridden.`
      )
    }
```

PR #1724, opened 2023-09-28
Contribution guide

michaelessiet:
This PR adds a contribution guide that outlines and explains strict guidelines on how to properly contribute to Tamagui.


natew:
I think we want to update the main readme then right? It sort of duplicates some of this content. Probably want to make sure quick start `npm create tamagui@latest` is called out there and then links to contributing and other relevant areas.

PR #1757, opened 2023-10-06
feat(native compiler): support theme value extraction

TheEhsanSarshar:


packages/static/src/extractor/createExtractor.ts
```diff
@@ -1942,12 +1942,24 @@ export function createExtractor(
...
               )
 
-              const outProps = {
+              let outProps = {
                 ...(includeProps ? out.viewProps : {}),
                 ...out.style,
                 ...out.pseudos,
               }
 
+              if (usedThemeKeys.size > 0) {
```
natew, line 1951: this part looks wrong, we need to be sure to keep the order of props because order is important, so we can't change anything in terms of the ordering, so basically shouldn't have any special logic here it should all be part of `attrs` by the time we get down here

natew:
looks good as a start, i think needs some tests, at least 3 or so that test various usages and ensure:

- prop order is preserved
- theme color is what we expect
- avoids collisions with other variables

TheEhsanSarshar:
@natew 
order is correct now and supports the ternary operator correctly as well.
just need to test the rest operator and write some benchmarks

natew:
merged on another branchj

PR #1787, opened 2023-10-17
update create template

benschac:
@natew is going to take this one over


benschac:
@natew or @alitnk - could either of you try running tests on your local machine.

I'm having a hard time groking why this test is failing locally.

benschac:
command runs as expected locally and creates starter as expected.

@natew - what's next step for `starters`/`nextexposolito`?

rm -rf it from monorepo and update the docs?

natew:
@benschac we should have a way for the release script to update the external repo as part of this change for sure. even if it just looks locally at a dir you specify.

natew:
this has been implemented

PR #1802, opened 2023-10-23
fix(core): infer correct prop types for inner function of styleable

nderscore:
This fixes a regression in the types of `styleable()` introduced in [1.74.20](https://github.com/tamagui/tamagui/releases/tag/v1.74.20) by these commits:

- 3db30bf7
- fa19f604

Ever since this version, the inner function of `styleable` has an inferred type of `FunctionComponent<any>`

This PR restores default prop types for the inner function.

---

I tested this out on my local project and all of our `styleable` wrappers seem to be working as expected. We have at least once instance of a double-wrapped `styleable` (`Foo.styleable(...).styleable(...)`) and it seems fine too.

### DIFF
--- a/packages/web/src/types.tsx
+++ b/packages/web/src/types.tsx
@@ -1355,13 +1355,15 @@ export type StyleableOptions = {
 }
 
 export type Styleable<Props, Ref, BaseProps, VariantProps, ParentStaticProperties> = <
-  CustomProps extends Object | void = void,
-  X extends FunctionComponent<any> = FunctionComponent<any>
+  CustomProps extends Object = {},
+  X extends FunctionComponent<
+    Omit<Props, keyof CustomProps> & CustomProps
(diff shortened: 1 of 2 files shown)


packages/web/src/types.tsx
```diff
@@ -1355,13 +1355,15 @@ export type StyleableOptions = {
 }
 
 export type Styleable<Props, Ref, BaseProps, VariantProps, ParentStaticProperties> = <
-  CustomProps extends Object | void = void,
-  X extends FunctionComponent<any> = FunctionComponent<any>
+  CustomProps extends Object = {},
+  X extends FunctionComponent<
+    Omit<Props, keyof CustomProps> & CustomProps
+  > = FunctionComponent<Omit<Props, keyof CustomProps> & CustomProps>
```
nderscore, line 1361: I have a hunch that this generic slot could probably be removed 🤔 Leaving it for now though to avoid accidentally introducing different regressions.

PR #1835, opened 2023-11-05
[bug] invalid props.style key 'borderStyle' for Image fix

ps-george:
image

This style gets inserted here: https://github.com/tamagui/tamagui/blob/f7b64488813ce51ff4468aa8e113f8cd4608441e/packages/web/src/helpers/expandStyles.ts#L26-L30

It's valid for other component types, but not `<Image>`.

### DIFF
--- a/packages/web/src/createComponent.tsx
+++ b/packages/web/src/createComponent.tsx
@@ -612,7 +612,13 @@ export function createComponent<
       elementType,
       debugProp
     )
-
+    if (process.env.TAMAGUI_TARGET === 'native') {
+      if (Component.name === "Image") {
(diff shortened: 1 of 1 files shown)


natew:
this fix should definitely happen at the `Image` component level, but it is quite hard to get right if we use `useProps` we could make it happen, don't remember if we do... Image not supporting borderStyle is super strange on RN's part...

aamirbinabd:
What's the current remedy/workaround for this? We're are in early development phase as of now, should we wait or use a Stack wrapper around Image to handle it?

natew:
Added to my backlog but closing this as I'll land a better fix

PR #1904, opened 2023-11-22
studio: fix - show parent themes in subtheme list

stephenlaughton:


natew:
probably ahve to revisit im likely refactoring that screen next

PR #1909, opened 2023-11-24
Feat/new menu comp

byteab:
### DropDown Menu

- [x] initial code
- [x] ready for web
- [x] ready for native

### Context Menu

- [x] initial code
- [x] ready for web
- [x] ready for native

### Final steps

- [x] Api clean up
- [x] fix some autofocus a11y issues
- [x] Make it support native prop
- [x] documentation
- [ ] LGTM


packages/menu/src/Menu.tsx
```diff
@@ -1,68 +1,1667 @@
...
+      textValue={textValue ?? textContent}
+    >
+      <RovingFocusGroup.Item
+        asChild
+        __scopeRovingFocusGroup={__scopeMenu || MENU_CONTEXT}
+        focusable={!disabled}
+        flexDirection="row"
+        alignItems="center"
+      >
+        <ThemeableStack
```
natew, line 917: We definitely need `unstyled` option, but also probably we should just go headless right away with `createMenu` where we have packages/menu export two things:

1. the default exports are "styled", but accept `unstyled` props
2. we add a new entry `@tamagui/menu/headless` and that points to just the `createMenu` stuff
3. createMenu doesn't rely on `tamagui/core` or `tamagui/web` at all and just passes typed props based on the state
4. we split all the tamagui specific styling/core stuff into a new file that passes them into createMenu for the default styled export

I will fix a lot of small menu styling things, but obviously that should be easy to do from the end-user perspective as well through headless and unstyled.

Also since this file got so huge we probably want to split it up, but I guess that the headless split will be good enough for that.

natew:
Why close? I know it’s old but once we get v2 out we will eventually revisit

byteab:
@natew there is another headless PR for that
https://github.com/tamagui/tamagui/pull/1978

PR #1916, opened 2023-11-27
move props to context

benschac:
move props that gradient depends on into context so the Skeleton API can be composable

```tsx
<Skeleton>
  <Skeleton.Gradient />
</Skeleton>
```

Per the requirements


packages/skeleton/src/skeleton-new.tsx
```diff
@@ -33,6 +34,23 @@ function InnerComponent(props: InnerCompoentProps) {
   }
 }
 
+const AnimatedGradientContext = createContext<{
```
natew, line 37: btw `createStyledContext` could be useful works with context but also works in tamagui

natew (approved):

natew:
closing these unfinished ones you can re-open as a big one

PR #1922, opened 2023-11-28
add inset to web

benschac:
trying to add: https://developer.mozilla.org/en-US/docs/Web/CSS/inset

what file do i need to touch to add the rest of the feature?


benschac:
* master
  * **PR #1922** <a href="https://app.graphite.dev/github/pr/tamagui/tamagui/1922?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/tamagui/tamagui/1922?utm_source=stack-comment).

apps/kitchen-sink/src/Sandbox.tsx
```diff
@@ -3,12 +3,17 @@
...
+import { XStack, YStack } from 'tamagui'
 
 export const Sandbox = () => {
   return (
     <View style={{ width: '100%', height: '100%', padding: 50 }}>
       <>
         <Text color="unset">test me</Text>
+        <YStack pos="relative">
+          {/* @ts-expect-error */}
+          <XStack pos="absolute" margin="$8" inset={5}></XStack>
```
benschac, line 15: @natew - I think I'm close on this one. Trying to get `inset` to pickup spacing values like `$2`, but I'm clearly missing something. Is there anything else I need to configure in this PR?

packages/web/src/types.tsx
```diff
@@ -80,6 +80,12 @@ export type TamaguiComponentPropsBaseBase = {
    */
   id?: string
 
+  /**
+   * The inset CSS property is a shorthand that corresponds to the top, right, bottom, and/or left properties. It has the same multi-value syntax of the margin shorthand.
+   * {@see https://developer.mozilla.org/en-US/docs/Web/CSS/inset}
+   */
+  inset?: Properties['inset']
```
benschac, line 87: This seems correct, just not sure where to pass theme spacing values.

natew:
Inset isn't so simple because native doesn't support it yet right? In that case a simple variant would be the solution, something like:

`variants: { inset: { '...size': {}, ':number': () => {} } })`

But this is something that would be ideally in core, but it's adding a bit of bloat. I guess only native needs to variant so we could leave off some, but still.

They type would then need to be in web/types see things that use SpaceTokens or SizeTokens (also need to decide which it uses, I'd have to reference the area where we decide that and see, somewhere in propMapper or thereabouts).

PR #1947, opened 2023-12-06
Feat/compiler native support dynamic values

TheEhsanSarshar:
- [x] support theme values
- [x] support token values
- [x] support ternary values ( can include any theme or non-theme dynamic values both on the condition and consequence side )
- [x] support ternary on any style prop
- [x] support for single dynamic value ( eg  `width={calculatedWidth}` )

support any of these props
will compile them down to a performant vanilla react native code

```js
 const paddingLeft = 23 / props.metric
 const theme = useTheme()
 const isX = useX()
 const y = 2

<YStack
    background="$background"
    color={theme.color}
    width={"$3"}
    borderColor={isX ? '$green': '#000'}
    paddingLeft={paddingLeft}
    marginRight={y > 2 ? 10 : "$2"}
/>

```

### DIFF
--- a/packages/babel-plugin/src/index.ts
+++ b/packages/babel-plugin/src/index.ts
@@ -147,7 +147,7 @@ export default declare(function snackBabelPlugin(
 
                 assertValidTag(props.node)
                 const stylesExpr = t.arrayExpression([])
-                const forTernary: t.Expression[] = []
+                const expressions: t.Expression[] = []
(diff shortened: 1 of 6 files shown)


natew:
@TheEhsanSarshar I can merge this but I feel like we should add a few tests here to follow on. Maybe a small example too of the "after" of the component you have there ^

PR #1953, opened 2023-12-08
Process the "onScroll" prop in Sheet.ScrollView

sertony:
This assures that we do not skip the "scrollBridge" values processing
when passing this prop into a <Sheet.ScrollView /> overriding it here

Useful when using this ScrollView with lists such as [FlashList](https://github.com/Shopify/flash-list), i.e.
 ```
renderScrollComponent={Sheet.ScrollView}
```


natew:
tyvm

PR #1960, opened 2023-12-12
fix(group): group state change should only trigger children

### LINKED ISSUE #1929: iOS: Custom button - text with pressStyle. onPress doesn't work.
### Current Behavior

Scenario:

1. Only happens on iOS (works as expected on web, haven't checked Android)
2. I'm using a custom button following the guide https://tamagui.dev/docs/guides/how-to-build-a-button#the-solution-createstyledcontext
3. On my custom button text (`const ButtonText = styled(Text, { ... }`) when I apply a `pressStyle` prop, the `onPress` callback on the parent button is not called.

I guess this is somewhat of an edge case given how Tamagui implements pressStyle with Pressable. 

### Expected Behavior

I'd expect this to either be documented in the doc or fixed such that pressStyle would still be applied when pressing, but the parent button `onPress` would be called (perhaps through the context somehow)

### Tamagui Version

```markdown
latest
```

```markdown
iOS
```

### Reproduction

```markdown
No easy repro as this doesn't happen on web. Quick way to repro locally is to follow the custom button guide: https://tamagui.dev/docs/guides/how-to-build-a-button#the-solution-createstyledcontext
And change the custom button text to:

const ButtonText = styled(Text, {
  name: 'ButtonText',
  context: ButtonContext,
  color: '$color',
  userSelect: 'none',

  pressStyle: {
    color: 'red'
  }
  variants: {
    size: {
      '...fontSize': (name, { font }) => ({
        fontSize: font?.size[name],
      }),
    },
  } as const,
})
```
```


### System Info

_No response_

### PR DESCRIPTION
**Pull request by TheEhsanSarshar**
fix #1929 #1937

### CODE REVIEW

File: packages/web/src/createComponent.tsx [hunk 1]
```diff
@@ -398,13 +399,33 @@ export function createComponent<
...
       // when we set state we also set our group state and emit an event for children listening:
       const groupContextState = componentContext.groups.state
       const og = setStateShallow
       setStateShallow = (state) => {
         og(state)
-        componentContext.groups.emit(groupName, {
+        groupHandlers.emit(groupName, {
           pseudo: state,
         })
         // and mutate the current since its concurrent safe (children throw it in useState on mount)
```
TheEhsanSarshar, line 431: I hope I haven't broken anything here, I am not exactly sure about the logic in this block

packages/web/src/createComponent.tsx
```diff
@@ -398,13 +403,28 @@ export function createComponent<
     const groupName = props.group as any as string
     const groupClassName = groupName ? `t_group_${props.group}` : ''
 
+    if (groupName && !stateRef.current.group) {
+      stateRef.current.group = {
```
natew, line 407: i think this will break probably

`<X group="a"><Y group="b"><Z $group-a={{}} /></Y></X>`

because it overwrites the emit/subscribe so now X group emit will get overwritten, then Z wont pick it up

TheEhsanSarshar: that combination is working fine! just tested it on the web.
because inner Providers merge down outer Providers
```js
 // change reference so context value updates
      return {
        ...componentContext.groups,
        // change reference so as we mutate it doesn't affect siblings etc
        state: {
          ...componentContext.groups.state,
          [groupName]: {
            pseudo: initialState,
```

TheEhsanSarshar: Screenshot 2023-12-20 at 18 13 26
in this image `a`, `b`, and `c` are groups

packages/web/src/createComponent.tsx
```diff
@@ -1162,7 +1177,12 @@ export function createComponent<
     if (process.env.NODE_ENV === 'development' && time) time`create-element`
 
     // must override context so siblings don't clobber initial state
+    const groupState = stateRef.current.group
     const subGroupContext = useMemo(() => {
+      if (!groupState) return
+      if (groupState) {
```
natew, line 1183: no need for if check after the early return

packages/web/src/createComponent.tsx
```diff
@@ -901,11 +921,6 @@ export function createComponent<
       if (classNames) classList.push(Object.values(classNames).join(' '))
       if (groupClassName) classList.push(groupClassName)
 
-      // if (fromTheme) {
```
natew, line 904: can remove

packages/web/src/createComponent.tsx
```diff
@@ -340,14 +341,18 @@ export function createComponent<
...
         themeShallow?: boolean
         isListeningToTheme?: boolean
+        group?: {
+          listeners: Set<GroupStateListener>
+          emit: GroupStateListener
+          subscribe: (cb: GroupStateListener) => () => void
+        }
       }
     )
-    stateRef.current ||= {}
```
natew, line 350: technically this avoids making a new object after the first render! i try and make createComponent as fast as possible so may as well keep the tiny optimization

apps/site/protected/_utils/github.ts
```diff
@@ -163,21 +162,7 @@ export const checkForSponsorship = async (
   userToken: string
 ): Promise<GithubAccessStatus> => {
   const orgs = await getOrgs(userToken)
-  // const personalStatus = await isLoginSponsor(login)
```
natew, line 166: remove

PR #1978, opened 2023-12-20
Feat/headless menu

TheEhsanSarshar:
the follow up of this [PR](https://github.com/tamagui/tamagui/pull/1909)
made it to have a separate branch for headless stuff


code/core/web/src/createComponent.tsx
```diff
@@ -1088,6 +1095,8 @@ export function createComponent<
           <span
             className="_dsp_contents"
             {...(isHydrated && events && getWebEvents(events))}
+            // /** we passed these events to the content instead */
+            {...(events && { onFocus: undefined, onBlur: undefined })}
```
natew, line 1099: going to merge this into v2 branch but should i port this change back to v1 too?

code/ui/focus-guard/package.json
```diff
@@ -0,0 +1,37 @@
+{
```
natew, line 1: how is this different from focus-scope?

code/core/web/src/createComponent.tsx
byteab: I think it should be fine in v2 only. it won't have immediate effect to other part of Tamagui

code/ui/focus-guard/package.json
byteab: there are similar behavior but the focus-guard is a simplified version to fit the behaviour of menu and trap the focus by adding to guard span before and after the inner elements

natew:
Closing but may bring this into v2 branch soon

PR #1982, opened 2023-12-21
fix(nested Presence): checkbox nested inside PopoverContent

### LINKED ISSUE #1977: CSS animation driver breaks animating inside a Popover
### Current Behavior

When using an animation inside `<Popover.Content>`, it prevents from exiting the popover once animation has been triggered.

An easy way to reproduce is to add a Checkbox inside a Popover, and animate the checkbox indicator with `enterStyle`.

Then open popover and toggle checkbox.

If you try to close popover before animation is done, it closes properly. If you wait for animation to be done, you can't close anymore.

https://github.com/tamagui/tamagui/assets/674084/9efc09f9-e1c5-495c-8481-7cedaad47f59

### Expected Behavior

Inner animation should have no impact on the popover behavior. Same as when there is no animation at all:

https://github.com/tamagui/tamagui/assets/674084/9a67e7bf-40ac-4df7-b594-8fa1abaaf3fe

### Tamagui Version

```markdown
1.79.12
```

```markdown
web
```

### Reproduction

```markdown
// https://github.com/TimPetricola/tamagui/commit/e13ca28b741ca168b8d903a4c8de5dd3b0e69583

import { Check } from '@tamagui/lucide-icons'
import { Button, Checkbox, Popover, YStack } from 'tamagui'

  return (
    <Popover>
      <Popover.Trigger>
        <Button>Trigger</Button>
      </Popover.Trigger>
      <Popover.Content key="content" borderWidth={1} borderColor="$borderColor">
        <Checkbox size="$4">
          <Checkbox.Indicator enterStyle={{ scale: 0 }} animation="bouncy">
            <Check />
          </Checkbox.Indicator>
        </Checkbox>
      </Popover.Content>
    </Popover>
  )
}
```

### System Info

_No response_

TheEhsanSarshar:
fix #1977

### DIFF
--- a/packages/checkbox/src/Checkbox.tsx
+++ b/packages/checkbox/src/Checkbox.tsx
@@ -405,9 +405,7 @@ const CheckboxComponent = CheckboxFrame.styleable<CheckboxExtraProps>(function C
               size={propsActive.size ?? styledContext?.size ?? '$true'}
               scaleIcon={scaleIcon ?? styledContext?.scaleIcon ?? 1}
             >
-              <Animate type="presence" present={Boolean(checked)}>
-                {propsActive.children}
-              </Animate>
+              {propsActive.children}
             </CheckboxStyledContext.Provider>
           </CheckboxFrame>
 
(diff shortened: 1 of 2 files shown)


TimPetricola (approved):
I confirm it fixes the issue 🙏

natew:
I think we should fix this on the Popover level. We should "reset" presence on the inner content.

So somewhere in Popover, something like this (pseudocode):

```tsx
<AnimatePresence>
   <someotherstuff>
       <AnimatePresenceContext value={null}>
           {props.children}
   ```

TheEhsanSarshar:
fixed now

PR #1986, opened 2023-12-21
fix(tag prop): tag ignored when there is animation prop

### LINKED ISSUE #1985: Using `tag` on an element with `animation` is ignored
### Current Behavior

If you set both `tag` and `animation` props on an element, the `tag` will completely be ignored (tested with both CSS and Moti driver).

### Expected Behavior

The proper `tag` should still be used when animating

### Tamagui Version

```markdown
1.79.12
```

```markdown
web
```

### Reproduction

```tsx
import { View } from '@tamagui/core'

  return (
    // Inspect this in browser, and you'll see that a `div` is rendered instead of a `button`
    <View
      tag="button"
      animation="bouncy"
      height="$4"
      width="$4"
      backgroundColor="$red5"
    />
  )
}
```

### System Info

_No response_

TheEhsanSarshar:
fix #1985

### DIFF
--- a/packages/web/src/createComponent.tsx
+++ b/packages/web/src/createComponent.tsx
@@ -769,6 +769,10 @@ export function createComponent<
     // so the type is pretty loose
     let viewProps = nonTamaguiProps
 
+    if (hasAnimationProp && props.tag && !props.role && !props.accessibilityRole) {
+      viewProps.role = props.tag as any
+    }
+
     if (isHOC && _themeProp) {
       viewProps.theme = _themeProp
     }


natew:
This feels a little weird, how does it fix the tag going away? Doesn't it just set the role? I would imagine with CSS driver the tag shouldn't go away, do you know why?

TheEhsanSarshar:
@natew That's because createComponent uses an AnimatedView and react-native-web only translate role and accessibilityRole to corresponding tags

TheEhsanSarshar:
interesting! with CSS driver tag goes away and `role` also has no effect

natew:
With CSS driver we control the final element completely so `tag` should just work no?

PR #1994, opened 2023-12-22
reset.css reserve width for vertical scrollbars

0xBigBoss:
Found out that I added the reset css to the wrong place and only found out due to me starting a new tamagui project.

Not sure what to do about the @tamagui/web file.

### DIFF
--- a/packages/core/reset.css
+++ b/packages/core/reset.css
@@ -3,7 +3,7 @@ body, #root, #__next {
   display: flex;
   flex-direction: column;
   min-height: 100vh;
-  min-width: 100vw;
+  min-width: calc(100vw - (100vw - 100%)); /* reserve space for scrollbar */
 }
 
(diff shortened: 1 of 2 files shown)


packages/core/reset.css
```diff
@@ -3,7 +3,7 @@ body, #root, #__next {
   display: flex;
   flex-direction: column;
   min-height: 100vh;
-  min-width: 100vw;
+  min-width: calc(100vw - 100%); /* Reserve the width of the scrollbar, even if it's not present */
```
natew, line 6: [SO says](https://stackoverflow.com/a/34884924)  calc(100vw - (100vw - 100%))

0xBigBoss: thx for double checking. dunno how i missed a whole calc. fixed both versions. @natew

PR #2006, opened 2023-12-27
bento - add more date pickers

TheEhsanSarshar:
- month picker
- year picker
- range picker
- multi-select picker


manhhavu:
Unless I'm missing something, @TheEhsanSarshar , I tried to open the .tsx files you added in this PR, they are all in binary format, seems to be very strange?

TheEhsanSarshar:
@manhhavu Those are part of pro-components and will be available for early access some day next week

PR #2014, opened 2023-12-28
fix(button): always respect tag prop

oosawy:
Button should respect tag prop like others do.


natew:
You were missing parens around the conditional, but also I feel like a bit cleaner so I did it here 5a93084f5b3b92d887ae58ab73fb50069759083b

Thanks for the heads up! Next time I'll get your PR merged :/ just didn't want overhead here going back and forth. Much appreciated.

PR #2021, opened 2023-12-31
fix(web): add missing component display names

alitnk:
Fixes https://github.com/tamagui/takeout/issues/179


natew:
Hm this may slow down things just a bit because now it will look up themes on every XStack/View etc, which imo is a bit overkill. But I'll merge this for now, it makes sense and we can check performance and see if it regresses before release.

PR #2065, opened 2024-01-09
fix: broken links in installation doc

bdjessup:


bdjessup:
Just noticed this is a duplicate

PR #2080, opened 2024-01-12
docs: removed content that is not applicable

harshmohite-fradium:
View component has defaultProps that injects flexDirection: 'column'. Hence this content block is not applicable. Check the code below:

```
export const View = createComponent<StackProps, View, StackPropsBase>({
  acceptsClassName: true,
  defaultProps: {stackDefaultStyles},
  validStyles,
})
```


natew:
View by default is column, but View and Stack are now interchangeable and don't set any styles on native (on web they do). So this is correct to remove the warning.

PR #2081, opened 2024-01-12
fix(moti): don't use css vars

### LINKED ISSUE #1520: Bug: Moti animation driver breaks with some complex styles (media queries?)
### Current Behavior

When using the Moti animation driver on a component, sometimes the animation does not occur. 

I haven't narrowed down the specific cause yet, but it seems to be related to combining variants and media query styles together.

The React Native animation driver does not have this issue. I've only tested this on web so far - its possible this doesn't happen on native.

### Expected Behavior

Animation 

### Tamagui Version

```markdown
1.50.3
```

### Reproduction

I've created a reproduction here:

https://github.com/nderscore/tamagui-moti-web-bug

Please see the notes in the readme.

### System Info

_No response_

TheEhsanSarshar:
fix #1520

### DIFF
--- a/packages/animations-moti/src/createAnimations.tsx
+++ b/packages/animations-moti/src/createAnimations.tsx
@@ -25,7 +25,7 @@ export function createAnimations<A extends Record<string, MotiTransition>>(
     Text: Animated.Text,
     isReactNative: true,
     keepStyleSSR: true,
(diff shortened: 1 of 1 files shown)


natew:
This makes it slower on web unfortunately, but correctness worth it for now.

PR #2095, opened 2024-01-16
feat(lucide-icons): upgrade Lucide icons to the latest version

alantoa:
## Why  

Currently, the Tamagui icons version is lagging behind the Lucide icons. Sometimes, when I search for an icon on the Lucide website and try to use it in Tamagui, I find that Tamagui does not have them. So I created a PR for it.

## How 

- Due to the modification of the Lucide icons folder in the new version, I have updated the `lucideIconsDir` location in the `lucide-icons/bin/build.js` file.
- Run `yarn generate` and `yarn build`.
- Then I ran the command `npx @biomejs/biome format --write ./` to format all files in the `lucide-icons` folder, following the `biome.json` config.

## Test  

I have tested the `site` project, and it now has the new icons from Lucide.

|  Before  | After  |
|  ----  | ----  |
| CleanShot 2024-01-16 at 18 13 50@2x  | image |


natew:
Thanks Alan, lgtm.

PR #2104, opened 2024-01-18
feat(select): add focus ring

TheEhsanSarshar:


packages/select/src/SelectItem.tsx
```diff
@@ -189,10 +189,17 @@ export const SelectItem = ListItemFrame.styleable<SelectItemProps>(
...
               hoverTheme: true,
-              focusTheme: true,
               cursor: 'default',
               outlineWidth: 0,
               size,
+              focusStyle: {
+                outlineOffset: -2,
+                outlineWidth: 2,
+                outlineStyle: 'solid',
+                outlineColor: '$color7',
```
natew, line 199: we avoid specific colors for now, but we havent added `outlineColor` to the tamagui themes. We should do that and set it to color7, then use it in all the variants really.

Issue #2118, opened 2024-01-20
Dialog: `trapFocus`  forces element to always be focused


ArthurYidi:
### Current Behavior

`trapFocus` force elements to alway be focused. It should only trap the focus for elements inside the dialog.

### Expected Behavior

Allow elements to focus out, specially important in mobile so the **keyboard goes down**.

https://github.com/tamagui/tamagui/assets/1154353/097bba74-5496-4bf4-b90f-d001b849e649

### Tamagui Version

```markdown
`current`
```

### Platform (Web, iOS, Android)

```markdown
- Web: Desktop and Android Chrome
```

### Reproduction

```markdown
Kitchen Sink: `http://localhost:9000/?demo=Dialog`

- Add `trapFocus={open}` to `Dialog.Content`
- Focus on an element inside a Dialog
- Click on a non-focusable element
- Bug: Focus is still on the element
 
Note: modified default focus colors so it's easier to see.
```
https://github.com/tamagui/tamagui/assets/1154353/5cbcdaee-801e-4f74-93f1-0a2df84f0f05

TheEhsanSarshar:
`Dialog.Content` also accepts `trapFocus` prop. making that prop `true` will fix this issue.
there seems to be missing types.

ArthurYidi:
@TheEhsanSarshar it does not fix the issue see reproduction steps.

TheEhsanSarshar:
emm, sorry maybe didn't understand the issue correctly.
are we going to make the focus to always be trapped inside Dialog?
if yes then making that prop to `true` should fix it.
btw I have a fix, le see if that resolve the issue

ArthurYidi:
@TheEhsanSarshar no I proposed to help tackle that in a separate issue. 

> are we going to make the focus to always be trapped inside Dialog?

no the issue is about allowing the element to un-focus when clicking inside the dialog.

ArthurYidi:
@TheEhsanSarshar also see: https://discord.com/channels/909986013848412191/909986013848412194/1197350336579522712

TheEhsanSarshar:
that unfocus is working fine somehow in the doc.
btw we need to trap focus when it's modal and open, and I think it's better to keep it uncontrolled ? as that will avoid the bug of unclickable content when the user forgets to make the trapFocus false when Dialog is closed!

TheEhsanSarshar:
is that still happening on the [site](https://tamagui.dev/docs/components/dialog/1.0.0)

[...]

PR #2119, opened 2024-01-20
fix(Dialog): reorder `Dismissable` so `focusout` works correctly

ArthurYidi:
Fixes: #2118

### DIFF
--- a/packages/dialog/src/Dialog.tsx
+++ b/packages/dialog/src/Dialog.tsx
@@ -91,7 +91,7 @@ const DialogTriggerFrame = styled(View, {
   name: TRIGGER_NAME,
 })
 
-interface DialogTriggerProps extends StackProps {}
+interface DialogTriggerProps extends StackProps { }
 
 const DialogTrigger = DialogTriggerFrame.styleable(
   (props: ScopedProps<DialogTriggerProps>, forwardedRef) => {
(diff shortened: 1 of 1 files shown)


TheEhsanSarshar:
Thanks for the PR!
it seems focus out handler is useful, as also mentioned in the Radix
https://github.com/radix-ui/primitives/blob/main/packages/react/focus-scope/src/FocusScope.tsx#L91

ArthurYidi:
@TheEhsanSarshar yeah! Making some changes https://discord.com/channels/909986013848412191/994937898908516362/1198416664128147546

packages/dialog/src/Dialog.tsx
```diff
@@ -584,28 +585,29 @@ const DialogContentImpl = React.forwardRef<TamaguiElement, DialogContentImplProp
...
-            ref={composedRefs}
-            onDismiss={() => context.onOpenChange(false)}
+            onUnmountAutoFocus={onCloseAutoFocus}
           >
+
             {contents}
-          </Dismissable>
-        </FocusScope>
+          </FocusScope>
+        </Dismissable>
```
ArthurYidi, line 610: Switches the order of `FocusScope` <-> `Dismissable`. That way `FocuScope`'s container is the dialog.

packages/dialog/src/Dialog.tsx
```diff
@@ -460,6 +460,7 @@ const DialogContentModal = React.forwardRef<TamaguiElement, DialogContentTypePro
         onFocusOutside={composeEventHandlers(props.onFocusOutside, (event) =>
           event.preventDefault()
         )}
+        outlineStyle='none'
```
ArthurYidi, line 463: disable `outline` for the overall dialog (same approach as Adobe's spectrum aria)

natew:
Thanks, tested it out and seems fine, but I do remember we've switched orders on these two before at some point. Couldn't find it, will test this out a bit more before releasing.

natew:
Merged on another PR

PR #2121, opened 2024-01-21
fix(dialog): broken animation

TheEhsanSarshar:


natew:
This needed an animatableDefaults fix i put into master as it should handle that automatically. We should add a test for that actually if you get a chance to be sure it sets x=0 y=0 by default if you don't so we don't regress there again.

PR #2122, opened 2024-01-21
feat(theme names): short css theme names initial changes

byteab:


packages/site-config/src/tamagui.config.ts
```diff
@@ -211,11 +211,18 @@ const monoFont = createGenericFont(
   }
 )
 
+let themeKeysOnly = Object.fromEntries(
+  Object.entries(themesIn).map(([key], index) => [key, index])
```
natew, line 215: Can't we just use the index of the array and just use Object.keys()?

natew: Ohh i see its for type compat

packages/web/src/createTamagui.ts
```diff
@@ -84,9 +84,7 @@ export function createTamagui<Conf extends CreateTamaguiProps>(
       )
 
   if (configIn.themes) {
-    const noThemes = areThemesJustNames
-      ? false
-      : Object.keys(configIn.themes).length === 0
+    const noThemes = areThemesJustNames ? true : Object.keys(configIn.themes).length === 0
```
natew, line 87: How hard is backward compat?

PR #2129, opened 2024-01-22
fix(site:tint-glow): get stuck at the bottom fast scroll

TheEhsanSarshar:


natew:
had to do a few more things to get it right

PR #2144, opened 2024-01-24
fix(select): export SelectValueExtraProps interface

nderscore:
Fix for this TypeScript error when inferring types from the `Select` component.

```
error TS4023: Exported variable 'MySelect' has or is using name 'SelectValueExtraProps' from external module "/path/to/node_modules/@tamagui/select/types/Select" but cannot be named.
```


packages/select/types/Select.d.ts
```diff
@@ -331,5 +331,4 @@ export declare const Select: ((props: ScopedProps<SelectProps>) => JSX.Element)
         }>> & React.RefAttributes<import("react-native").ScrollView>>;
     };
 };
-export {};
```
nderscore, line 334: Not sure why this line disappeared after running `yarn build:types` 🤔

PR #2152, opened 2024-01-27
feat(themes): add v3 themes with much lighter payload

natew:


natew:
i did some work here, lets just document this + fix tests and we're good to go.

natew:
Btw does the old behavior of having the css be longer names and you pass an empty object still work? We should have that still work until next major.

PR #2173, opened 2024-01-31
Keep handle dragging for Sheet on Web if scrolled

### LINKED ISSUE #2162: Sheet Handle doesn't drag if scrolled at all
### Current Behavior

Given a `Sheet` with a `Sheet.ScrollView` inside it, if the inner `ScrollView` is dragged at all, then it `Handle` won't drag

### Expected Behavior

Ideally the `Handle` will continue to drag no matter what

I isolated this behaviour to these lines: https://github.com/tamagui/tamagui/blob/master/packages/sheet/src/SheetImplementationCustom.tsx#L277-L280

Removing those lines creates the desired behaviour for me. But I of course believe those lines exist for a reason. Maybe this should be configurable?

### Tamagui Version

```markdown
1.88.13
```

```markdown
Web - Chrome
```

### Reproduction

```markdown
Using the code from the Sheet docs, this should reproduce it (I just added a Sheet.ScrollView):


      <Sheet
        forceRemoveScrollEnabled={open}
        modal={modal}
        open={open}
        onOpenChange={setOpen}
        snapPoints={snapPoints}
        snapPointsMode={snapPointsMode}
        dismissOnSnapToBottom
        position={position}
        onPositionChange={setPosition}
        zIndex={100_000}
        animation="medium"
      >
        <Sheet.Overlay
          animation="lazy"
          enterStyle={{ opacity: 0 }}
          exitStyle={{ opacity: 0 }}
        />
        <Sheet.Handle />
        <Sheet.Frame padding="$4" justifyContent="center" alignItems="center" space="$5">
          <Sheet.ScrollView>
            <Button size="$6" circular icon={ChevronDown} onPress={() => setOpen(false)} />
            <Input width={200} />
            {modal && isPercent && (
              <>
                <InnerSheet open={innerOpen} onOpenChange={setInnerOpen} />
                <Button
                  size="$6"
                  circular
                  icon={ChevronUp}
                  onPress={() => setInnerOpen(true)}
                />
              </>
            )}
          </Sheet.ScrollView>
        </Sheet.Frame>
      </Sheet>


Note I've manually patched out the lines I mentioned above, and you can see that behaviour live on our site: 
https://guild.host/guild and then click the "Donate" button in a small viewport
```

### System Info

_No response_

tazsingh:
Fixes #2162 

As mentioned in that issue, I'm not sure why this logic is here

However, since I'm currently only working on web, I assume that perhaps it's there for mobile reasons?

Thus, sending this PR to hopefully help identify its purpose and at the least help narrow the impact of this behaviour

As always, let me know how I can help move things forward and thanks for an awesome library!

### DIFF
--- a/packages/sheet/src/SheetImplementationCustom.tsx
+++ b/packages/sheet/src/SheetImplementationCustom.tsx
@@ -274,7 +274,7 @@ export const SheetImplementationCustom = themeable(
         const isDraggingUp = dy < 0
         // we can treat near top instead of exactly to avoid trouble with springs
         const isNearTop = scrollBridge.paneY - 5 <= scrollBridge.paneMinY
-        if (isScrolled) {
+        if (isScrolled && !isWeb) {
           previouslyScrolling = true
           return false
         }


natew:
Thanks for the heads up, looked at it this morning and I think this should do: f6714594f1477f14406ea6d7e28165ae0cf13d24

Just filters for the actual handle element.

tazsingh:
Perfect, thanks so much @natew 🙌

PR #2196, opened 2024-02-03
Import package, not types

skenmy:
This resolves a problem with using Reanimated animations - the compiler is unable to resolve the `/types` import correctly.


natew:
No idea why but something keeps changing this, either VSCode or our lint fix.

PR #2210, opened 2024-02-05
fix(createComponent): avoid passing themeInverse to DOM

TheEhsanSarshar:


natew:
Idk where my last comment went but `delete` is relatively expensive for performance last I checked, and also is unsafe code, i think we can add this to skipProps no?

PR #2211, opened 2024-02-05
fix(core): only resolve tag on web

TheEhsanSarshar:


natew:
It seems like the inner logic is just wrong, but maybe I'm not understanding, can you document it?

PR #2234, opened 2024-02-08
feat: Added native Progress implementation

andrew-levy:
- Added native progress component written with SwiftUI
- Added to kitchen sink for testing
- Supports `max` and `value` props (I think it also supports all tamagui props but it applies it to the wrapper not the actual swiftui view... need to test this more)

### DIFF
--- a/packages/progress/src/Progress.tsx
+++ b/packages/progress/src/Progress.tsx
@@ -1,28 +1,35 @@
 // forked from Radix UI
 // https://github.com/radix-ui/primitives/blob/main/packages/react/progress/src/Progress.tsx
 
-import type { GetProps } from '@tamagui/core'
-import { getVariableValue, styled } from '@tamagui/core'
-import type { Scope } from '@tamagui/create-context'
-import { createContextScope } from '@tamagui/create-context'
-import { getSize } from '@tamagui/get-token'
-import { withStaticProperties } from '@tamagui/helpers'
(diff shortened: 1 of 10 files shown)


packages/select/src/Select.tsx
```diff
@@ -440,6 +441,10 @@ export const Select = withStaticProperties(
       []
     )
 
+    if (props.native && isIos) {
+      return <NativeSelect {...props} />
+    }
+
```
andrew-levy, line 447: Likely can't do this naive approach. Will probably need a setup function since expo modules core is needed.

packages/progress/src/Progress.tsx
```diff
@@ -173,6 +179,10 @@ const Progress = withStaticProperties(
     const valueLabel = isNumber(value) ? getValueLabel(value, max) : undefined
     const [width, setWidth] = React.useState(0)
 
+    if (props.native && isIos) {
+      return <NativeProgress {...props} />
+    }
+
```
andrew-levy, line 185: Likely can't do this naive approach. Will probably need a setup function since expo modules core is needed.

natew:
I think we'd basically want to do it how we do Sheet now but even a bit further since we have the native package ourselves

- `@tamagui/progress` exports some sort of setupNative function
- `@tamagui/progress-native` probably is the actual combined progress + native
- `tamagui/progress` can export `@tamagui/progress-native`
- new package `@tamagui/kit` that is basically `tamagui` but all the +native versions

PR #2241, opened 2024-02-10
Move return null of Popover.Content not visible one level up

dohomi:
This PR would change the render mechanism of `<Popover/>`, return null will be moved one level up meaning that the portal `DIV/VIEW` will not be rendered if not visible. 

This aims to cleanup the created DOM elements and hopefully also solves some issues with trap focus if too many popovers would be rendered on a page.


natew:
Seems to work and should be faster for sure

Issue #2223, opened 2024-02-07
Input inside Dialog doesn't drop focus as expected


chrisaugustsson:
### Current Behavior

Using blur() on Input inside a Dialog causes the text inside the input to be highlighted. Doing the same outside doenst cause this.

### Expected Behavior

Using blur() on the ref should drop focus.

### Tamagui Version

```markdown
1.89.11
```

### Platform (Web, iOS, Android)

```markdown
Web
```

### Reproduction

```markdown
https://codesandbox.io/p/sandbox/long-butterfly-2zxj54?file=%2Fpackages%2Fapp%2Ffeatures%2Fhome%2Fscreen.tsx%3A37%2C32
```

natew:
You have to instead focus another element within Dialog. Dialog has a focus guard that traps focus within it, so blurring instead is forced back to the first focused action

natew:
You can also disable the focus trap behavior! Lmk if that works.

chrisaugustsson:
> You can also disable the focus trap behavior! Lmk if that works.

That did work, however, there doesn't seem to be a way to set trapFocus to false from the Dialog.Content. If you take a look into DialogContentImpl, trapFocus is set to context.open, so there is no way to set it to false.

In FocusScope, the handleFocusOut function selects the last focused element if the target isn't in the container, but wouldn't it be possible to set to focus to the container itself? It works for my issue, but I'm not sure if it breaks anything else.

chrisaugustsson:
If my suggestion would be a valid solution, I can make a PR for it.

natew:
Looks like maybe two things here a bugfix for passing trapFocus. 

The second part is harder. Do we want to make trapFocus take another argument to prefer the container? But what if we want the first time to focus the input but then all blurs to do the container? I’d think what we could do is add some special logic to detect a blur called inside and have it automatically return the focus to the container rather than the first input in just that special case. Happy to accept a PR.

chrisaugustsson:
I have a possible solution, that might be a bit hacky. However, I'm not allowed to create a PR?

In short the fix looks like this:
 `    function handleFocusIn(event: FocusEvent) {
      if (focusScope.paused || !container) return
      const target = event.target as HTMLElement | null
      if (container.contains(target)) {
        // Set container as lastFocusedElement to prevent inputs
        // to be refocused on blur events
        target?.addEventListener('blur', () => {
          lastFocusedElementRef.current = container
        })
        lastFocusedElementRef.current = target
[...]

PR #2242, opened 2024-02-10
Adds event listener on target for blur events

chrisaugustsson:
Related to https://github.com/tamagui/tamagui/issues/2223

Using input inside FocusScope with trapped focus, the input gets reselected on blur events.

This PR adds a eventlistener for blur events, and sets focus to the container if blur events are triggered.

### DIFF
--- a/packages/focus-scope/src/FocusScope.tsx
+++ b/packages/focus-scope/src/FocusScope.tsx
@@ -70,6 +70,11 @@ export function useFocusScope(
       if (focusScope.paused || !container) return
       const target = event.target as HTMLElement | null
       if (container.contains(target)) {
+        // Set container as lastFocusedElement to prevent inputs
+        // to be refocused on blur events
+        target?.addEventListener('blur', () => {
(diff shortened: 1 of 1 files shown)


natew:
We'd have to do this differently, but the idea is right. You need to handle/dispose that event listener.

chrisaugustsson:
> We'd have to do this differently, but the idea is right. You need to handle/dispose that event listener.

What do you think about using a abort controller to remove the event listener?

natew:
this looks much better

packages/focus-scope/src/FocusScope.tsx
```diff
@@ -66,26 +66,37 @@ export function useFocusScope(
   React.useEffect(() => {
     if (!enabled) return
     if (!trapped) return
+    const controller = new AbortController;
```
natew, line 69:
```suggestion
    const controller = new AbortController();
```

??

natew:
merged in another pr, ty

PR #2258, opened 2024-02-12
Add selection color support

nandorojo:
Missing:

- [ ] Token support, not just theme
- [x] Autocomplete


natew:
Theme falls back to token automatically, so this should be fine, but I think using `acceptToken` would be the next step. We can help finish this one.

natew:
This should fix it a bit better https://github.com/tamagui/tamagui/commit/62216ace4f5fce693c23f85e01836639a5507b05

PR #2261, opened 2024-02-13
feat(core:compiler): use stableStyle for ternary

TheEhsanSarshar:


packages/static-tests/tests/__snapshots__/babel.native.test.tsx.snap
```diff
@@ -73,12 +84,18 @@ const _sheet = __ReactNativeStyleSheet.create({
     }]
   }
 });
+const __withStableStyle = require('@tamagui/core')._withStableStyle;
 const __ReactNativeView = require('react-native').View;
 const __ReactNativeText = require('react-native').Text;
 import { YStack } from 'tamagui';
 export function Test(isLoading) {
-  return <__ReactNativeView style={[_sheet[\\"0\\"], isLoading ? _sheet[\\"1\\"] : _sheet[\\"2\\"], _sheet[\\"3\\"]]} />;
```
natew, line 80: isnt it a bit slower tho now because it wraps a component in between? rather than just using directly? may not be a big speed up in that case

TheEhsanSarshar: I think having that array and passing that over the bridge on re-render is much slower than having a HOC

packages/static-tests/tests/__snapshots__/babel.native.test.tsx.snap
```diff
@@ -190,12 +207,18 @@ const _sheet = __ReactNativeStyleSheet.create({
     \\"marginBottom\\": 12
   }
 });
+const __withStableStyle = require('@tamagui/core')._withStableStyle;
 const __ReactNativeView = require('react-native').View;
 const __ReactNativeText = require('react-native').Text;
 import { Stack } from 'tamagui';
 export function Test(props) {
-  return <__ReactNativeView style={props === 123 ? _sheet[\\"0\\"] : _sheet[\\"1\\"]} />;
-}"
```
TheEhsanSarshar, line 198: for this one need to keep it as before, using wrapper is useless here

PR #2265, opened 2024-02-13
feat(core): add web only filter, backdrop, mix-blend-mode

ehxxn:
add the following props support ( web only )
`filter`
`backdropFilter`
`mixMlendMode`


packages/web/src/interfaces/RNExclusiveTypes.ts
```diff
@@ -19,12 +24,15 @@ export interface RNExtraProps {
...
+export interface RNViewProps
+  extends GestureResponderHandlers,
+    RNExtraProps,
+    WebOnlyProps {
   rel?: any
   download?: any
 }
 
-export interface RNTextProps extends RNExtraProps {
+export interface RNTextProps extends RNExtraProps, WebOnlyProps {
```
natew, line 35: i think this is wrong, should only be part of style props

packages/web/src/interfaces/RNExclusiveTypes.ts
```diff
@@ -2,6 +2,11 @@ import type { GestureResponderHandlers, LayoutChangeEvent } from 'react-native'
 
 type OnLayout = ((event: LayoutChangeEvent) => void) | undefined
 
+export interface WebOnlyProps {
+  filter?: React.CSSProperties['filter']
+  backdropFilter?: React.CSSProperties['backdropFilter']
+  mixBlendMode?: React.CSSProperties['mixBlendMode']
+}
```
natew, line 9: i think we have a web only style props interface somewhere in web/types

packages/helpers/src/validStyleProps.ts
```diff
@@ -149,6 +149,11 @@ export const stylePropsView = {
   justifyContent: true,
   left: true,
   margin: true,
+  ...(isWeb && {
```
natew, line 152: there's a place for this already just below

natew:
we need to update the docs for these too in styles.mdx. if we're updating these, we should be more thorough and catalog which are important and at least plausibly likely to be implemented by react native add a few more.

TheEhsanSarshar:
> we need to update the docs for these too in styles.mdx. if we're updating these, we should be more thorough and catalog which are important and at least plausibly likely to be implemented by react native add a few more.

we can also have these in react-native, there is a [library](https://github.com/iyegoroff/react-native-color-matrix-image-filters#supported-filters) for that. but for now we can skip it

TheEhsanSarshar:
@natew Added 30 more style props
I think it's better to have a separate PR to make current Apis compatible with web and deprecate old ones.
and keep this PR just adding new web styles

packages/web/src/types.tsx
```diff
@@ -1320,6 +1316,204 @@ interface ExtraStyleProps {
     | 'bottom'
     | TwoValueTransformOrigin
     | `${TwoValueTransformOrigin} ${Px}`
+  /**
+   * @platform: web
```
natew, line 1320: Can we make these say like the others ` Web-only style property. Will be omitted on native.`

packages/web/src/constants/webToNativeProps.native.tsx
```diff
@@ -0,0 +1,88 @@
...
+      contain: 'contain',
+      cover: 'cover',
+    }
+
+    let resizeMode = resizeModeMap[val] || 'cover'
+
+    return [['resizeMode', resizeMode]]
+  },
+  verticalAlign: (val: any) => {
+    const data = {
```
natew, line 16: hoist this out of the function to be faster

packages/web/src/constants/webToNativeProps.native.tsx
```diff
@@ -0,0 +1,88 @@
+export const webToNativeDynamicExpansion = {
+  objectFit: (val: any) => {
+    const resizeModeMap = {
```
natew, line 3: hoist this out of the function to be faster

apps/site/components/HeaderMenu.tsx
```diff
@@ -1,7 +1,8 @@
 import { ThemeTintAlt, useTint } from '@tamagui/logo'
 import { Menu } from '@tamagui/lucide-icons'
 import * as React from 'react'
-import type { ThemeName } from 'tamagui'
+import type {
```
natew, line 4: formatting

packages/web/types/types.d.ts
```diff
@@ -738,6 +734,204 @@ interface ExtraStyleProps {
...
+    paddingBlockEnd?: SpaceTokens | number;
+    paddingInline?: SpaceTokens | number;
+    paddingInlineStart?: SpaceTokens | number;
+    paddingInlineEnd?: SpaceTokens | number;
+    objectFit?: Properties['objectFit'];
+    verticalAlign?: Properties['verticalAlign'];
+    inset?: SpaceTokens | number;
+    insetBlock?: SpaceTokens | number;
+    insetBlockStart?: SpaceTokens | number;
+    insetBlockEnd?: SpaceTokens | number;
```
natew, line 925: are these supported on native?

TheEhsanSarshar: yeah, they are mapped to native alternatives.
like `paddingInline` to `paddingHorizontal`

Issue #2271, opened 2024-02-15
Lack of support focus:visible pseudo-class


arekkubaczkowski:
### Current Behavior

Currently, there is no distinction between the focus and focus:visible pseudo-classes, resulting in limitations when styling components for keyboard-only interactions.

The primary disparity between these two pseudo-classes lies in their application: focus is applied to elements regardless of whether they are selected by mouse or keyboard, whereas focus:visible specifically targets keyboard interactions. This differentiation is crucial for seamlessly styling buttons, inputs, etc., as in most scenarios, you wouldn't want focus styles applied to elements selected by mouse.

I would be very appreciative if you could add such functionality.

### Tamagui Version

```markdown
"tamagui": "1.89.20",
```

### Platform (Web, iOS, Android)

```markdown
Web
```

### Reproduction

```markdown
-
```

PR #2278, opened 2024-02-15
feat(core): add focus-visible support

TheEhsanSarshar:
fix #814 #2271


packages/web/src/helpers/pseudoDescriptors.ts
```diff
@@ -21,10 +26,16 @@ export const pseudoPriorities = {
   hover: 1,
   press: 2,
   focus: 3,
+  focusVisible: 3,
```
arekkubaczkowski, line 29: Just to make sure, should it be the same priority as `focus` is?

TheEhsanSarshar: yeah I think so. because focusStyle can also do the job of focusVisible for some components, so need to rely on order

arekkubaczkowski:
@TheEhsanSarshar thanks for this PR, is there any ETA when it will be released?

TheEhsanSarshar:
@arekkubaczkowski We will merge it someday this weekend

arekkubaczkowski:
I understand you are very busy but I would be really really appreciated for releasing it 🙏

TheEhsanSarshar:
@arekkubaczkowski yeah sorry for the delay, I'll ask the team to review it today if possible

packages/web/src/createComponent.tsx
```diff
@@ -987,14 +992,26 @@ export function createComponent<
               }),
             ...(attachFocus && {
               onFocus: (e) => {
-                setStateShallow({
-                  focus: true,
-                })
+                if (pseudos?.focusVisibleStyle) {
+                  setTimeout(() => {
```
natew, line 996: @TheEhsanSarshar why the settimeout here?

packages/web/src/createComponent.tsx
```diff
@@ -987,14 +992,26 @@ export function createComponent<
...
+                } else {
+                  setStateShallow({
+                    focus: true,
+                    focusVisible: false,
+                  })
+                }
                 onFocus?.(e)
               },
               onBlur: (e) => {
+                isKeyboardEvent.current = true
```
natew, line 1011: this could be triggered by a non-keyboard action no? clicking away?

packages/web/src/createComponent.tsx
```diff
@@ -909,11 +910,14 @@ export function createComponent<
 
     if (process.env.NODE_ENV === 'development' && time) time`events-setup`
 
+    const isKeyboardEvent = useRef(true)
```
natew, line 913: we should never add a useRef to component, hooks == expensive and this whole area is very sensitive. we have a single useRef we use for a bunch of state above

packages/web/src/helpers/pseudoDescriptors.ts
natew: so if both are defined it ends up being the order that determines it?

natew:
Will get this merged soon seems like we need to resolve just a few last things @TheEhsanSarshar, also I'd like to see at least a couple tests.

packages/web/src/helpers/pseudoDescriptors.ts
TheEhsanSarshar: yeah, order will determines it.
but the focusVisible will only be triggered by keyboard. so it will be ignored when mouse

packages/web/src/createComponent.tsx
TheEhsanSarshar: to avoid race condition
like when onFocus is triggered it can be focused by mouse, and `onPressIn` will make that flag `false`, right now there is no way to understand if it's keyboard focus or pointer focus on web, so we used this workaround. which should work fine.

packages/web/src/createComponent.tsx
TheEhsanSarshar: that's basically a reset

packages/web/src/createComponent.tsx
TheEhsanSarshar: ok thanks, I'll fix that

arekkubaczkowski:
@natew @TheEhsanSarshar focusVisibleStyle is not working along with animation prop. 
Possibly related to: https://github.com/tamagui/tamagui/issues/2299

```

const CheckboxSelectorFrame = styled(View, {
  tag: 'button',
  accessibilityRole: 'button',
  name: CHECKBOX_NAME,
  width: 20,
  height: 20,
  backgroundColor: '$color-bg-neutral-inverse',
  animation: 'instant',
  animateOnly: ['backgroundColor'],
  focusVisibleStyle: {
    outlineColor: 'red',
    outlineWidth: 4,
    outlineStyle: 'solid',
  },
})
```

PR #2286, opened 2024-02-17
poc: Fix Basic Vite Support

bnussman:
## Description 📝 

This is a POC that shows how `@tamagui/config/v3` is broken for Vite

The main two issues with basic Vite support are
- `process.env.TAMAGUI_BAIL_AFTER_SCANNING_X_CSS_RULES` makes Vite dev mode not work
- `@tamagui/config/v3`  exports react native code I think?

## Before
Screenshot from 2024-02-17 12-03-31

## After
[Screencast from 02-17-2024 11:41:30 AM.webm](https://github.com/tamagui/tamagui/assets/6440455/60e67f2d-b854-41b0-acaf-570c760106c0)


apps/vite/vite.config.ts
```diff
@@ -0,0 +1,10 @@
+import { defineConfig } from 'vite'
+import react from '@vitejs/plugin-react'
+
+// https://vitejs.dev/config/
+export default defineConfig({
+  define: {
+    'process.env.TAMAGUI_BAIL_AFTER_SCANNING_X_CSS_RULES': false,
```
bnussman, line 7: Somethig is going to have to be done with `process.env.TAMAGUI_BAIL_AFTER_SCANNING_X_CSS_RULES` in `insetStyleRule.tsx`. This won't work in the browser

I tried removing it but that package would break when it was rebuilt.

packages/config/src/v3.ts
bnussman, line 1: This solved the main problem

apps/vite/README.md
bnussman, line 1: I created this `vite` app just for testing.

apps/vite/vite.config.ts
natew: I'm not sure what to do about import.meta.env. Webpack doesn't support it:

https://github.com/webpack/webpack/issues/15833

So you're sort of stuck. It's actually a bit wild that we have the two most popular bundlers set up in a way that you can't really access env without some really weird hacks.

I think I've done something like `typeof import.meta !== 'undefined' ? import.meta.env.X : process.env.X` before.

packages/config/src/v3.ts
natew: The react plugin?

bnussman: My goal was to see what it would take to tamagui without any plugins. 

These changes in this file allowed me to do so.

natew:
Merged on another branch :)

PR #2306, opened 2024-02-20
fix shorthand for bc/borderColor

antoniopresto:
I think `bc` stands for "borderColor', right?
It stopped working for me after updated takeout today, maybe after tamagui@0.18.20


natew:
You'd want `@shorthands/v2` which has the updated values.

antoniopresto:
@natew do you know why it stopped working after updating to tamagui@1.89.20?

natew:
It was always bc => backgroundColor before, it never should have been bc => borderColor until v2. Not sure what could have happened there

PR #2319, opened 2024-02-22
Feat/moti support component config

byteab:
know issues
for some reason `translateX: { type: 'spring', overshootClamping: true }` doesn't work, but if I use `transform` instead it works.
@nandorojo probably an issue in moti `useMotify()`


nandorojo:
Oh sorry thought this was a PR on the moti repo

nandorojo:
But yeah I can fix this upstream

TheEhsanSarshar:
thanks!
yea I'll improve the code, ok I'll keep it translateX then

nandorojo:
For shorthands, what if you just did it before the `useMotify` step. This way the `transition` has a stable reference and does less work at render time.

```ts
createAnimations(animations) {
  for (const animationKey in animations) {
    const shorthands = { x: 'translateX', y: 'translateY' }

    for (const shorthand in shorthands) {
      if (animations[animationKey][shorthand]) {
        const styleName = shorthands[shorthand]
        animations[animationKey][styleName] = animations[animationKey][shorthand]
     }
    }
  }

  return { ... }
}
```

packages/animations-moti/src/createAnimations.tsx
```diff
@@ -172,9 +168,32 @@ export function createAnimations<A extends Record<string, MotiTransition>>(
...
-        ? props.animation[0]
+      const isAnimationArray = Array.isArray(props.animation)
+      const animationKey = isAnimationArray ? props.animation[0] : props.animation
+
+      const animationObject = isAnimationArray
+        ? props.animation
+            .filter((x) => typeof x !== 'string')
+            .reduce((a, b) => ({ ...a, ...b }))
         : props.animation
+      const normalizedAnimationObject =
```
natew, line 179: lets use a for loop and a single object, should be quite a bit more performance sensitive

natew: same for the animationObject above probably we can do everything in one loop

TheEhsanSarshar: thank! updated

TheEhsanSarshar:
@nandorojo for now we only have `x, y` so I guess it's fine to do it in a render loop?

PR #2320, opened 2024-02-22
[WIP] Specify package dependencies

antoniopresto:
Depending on the type of monorepo, undeclared dependencies break the build.
- [ ]  remove `./scripts/sync-deps/` folder.


scripts/sync-deps/sync-deps.ts
```diff
@@ -0,0 +1,106 @@
+#!/usr/bin/env bun-safe
```
natew, line 1: hm a couple things - we have some scripts already for grabbing all the packages and looping over them. i appreciate the effort but dont want two ways of doing things.

also we should read the current version from the root of the package.json rather than hardcode.

antoniopresto: Nice! I forgot to mention that it was a WIP/bootstrap, I intended to remove the script - the 2 packages I created basically for myself to use (no guarantees 😅)

Anyway, do you think this PR makes sense? I wouldn't mind closing it if there's another way to solve it :)

natew: What is the purpose of the changes? Bun or Nx compat or something?

antoniopresto: pnpm monorepo was throwing some package not found errors. (I solved it with pnpm custom hooks, to install missing dependencies…)

natew: Makes sense. I can accept some dep correctness stuff. I’ll have to check it more in depth. Lmk when it’s ready.

natew:
Going to close this out! Thanks for bringing it up though I'll have to revisit soon

PR #2329, opened 2024-02-23
fix(core): onLayout correct value when css scale

TheEhsanSarshar:


packages/core/src/helpers/getRect.tsx
```diff
@@ -6,3 +6,12 @@ export const getRect = (node: HTMLElement) => {
   const { x, y, top, left } = rect
   return { x, y, width: node.offsetWidth, height: node.offsetHeight, top, left }
 }
+
+export const getOffsetRect = (node: HTMLElement) => {
+  return {
```
natew, line 11: adding a note that reading offsetWidth triggers reflows in browsers so were paused on this for now

natew:
Superseded by new implementation using ResizeObserver

PR #2334, opened 2024-02-26
Updated import name

kasparas-ban:
Currently if you copy/paste the code that's in the installation instructions, you'll get a Typescript error because of the import name mismatch. This PR fixes that.


natew:
thank you!

PR #2372, opened 2024-03-02
Set hover and focus when disabled

chrisaugustsson:
If a button click is causing the button to be disabled, the button gets stuck in focus and hover state. Beacause the button is disabled it is not possible for the button to loose focus or hover state since all events are turned of.

### DIFF
--- a/packages/web/src/createComponent.tsx
+++ b/packages/web/src/createComponent.tsx
@@ -492,6 +492,10 @@ export function createComponent<
 
     const hasTextAncestor = !!(isWeb && isText ? componentContext.inText : false)
     const isDisabled = props.disabled ?? props.accessibilityState?.disabled
+    if (isDisabled) {
+      state.hover = false
+      state.focus = false
+    }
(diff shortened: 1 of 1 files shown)


natew:
Landed this but in a different way here, 3775156f47044974c70ae60020a1cd83907f5a6a

But really appreciate the PR!

chrisaugustsson:
> Landed this but in a different way here, [3775156](https://github.com/tamagui/tamagui/commit/3775156f47044974c70ae60020a1cd83907f5a6a)
> 
> But really appreciate the PR!

That fix doesn't seem to solve the problem? `mergeIfNotShallowEqual` never runs when disabled is set to true. 
There's already a check for disabled on line 427, that updates the disabled state. Here you could simply replace the current state with the initial state and it will reset the component:      ` setState({ ...initialState, disabled });`

chrisaugustsson:
If you want I can open up a new PR for this
@natew

PR #2395, opened 2024-03-06
[various-ui-fixes]: remove inverse switch for light theme on signup s…

bidah:
bento: normalize Choose File button to be same style and color as others
bento: normalize gap between columns on forms. Applied to SignUpTwoSide.tsx
bento: normalize spaceing between labels and inputs on SignUpTwoSide.tsx
bento: better overall spacing arrangement for Checkboxes group on SignUpTwoSide.tsx
bento: copywrite for label matching placeholder
bento: normalize width of buttons using token for width for SignUpScreen.tsx an SignUpTwoSide.tsx
bento: normalize ShortEmailPassword.tsx buttons border radius
bento: margin x to layout on mobile web
[various-ui-fixes]: style: fix: for ShortEmailPassword on mobile web
[various-ui-fixes]: normalize form email password with the same FormCard wrapper that brings in the faded border
[various-ui-fixes]: Checkbox label with pb 0 to match other form items spacing
[various-ui-fixes]: normalize extra sign in options for SignInScreen
[various-ui-fixes]: remove inverse switch for light theme on signup screen


natew:
Just one small thing `minWidth={'100%'}` actually this could de-opt the compiler (i need to check, we can fix easily and should) but also looks worse than just `minWidth="100%"`

bidah:
Ok. will disable it in the IDE then

PR #2397, opened 2024-03-07
feat(web): Support css variable prefixing via environment variable using TAMAGUI_CSS_VARIABLE_PREFIX

Mimameid:
This change allows prefixing generated css variable names to avoid collisions with third party/other stylesheets. 

- Adds a new constant `CSS_VARIBLE_PREFIX` (defaults to `''`) and environment variable `TAMAGUI_CSS_VARIABLE_PREFIX` to set a custom prefix

Info: Tried to do it via a property on tamagui config, figured that it branches a lot and there is a "before config exists" and "after config exists" phase. Decided to do it via an env variable as it simplifies things by a lot.


packages/tamagui/types/views/Input.d.ts
```diff
@@ -168,7 +168,7 @@ export declare function useInputProps(props: InputProps, ref: any): {
     componentName?: string | undefined;
     tabIndex?: string | number | undefined;
     disableOptimization?: boolean | undefined;
-    forceStyle?: "hover" | "press" | "focus" | undefined;
+    forceStyle?: "hover" | "press" | "focus" | "focusVisible" | undefined;
```
Mimameid, line 171: This one came in when I was merging and had `watch` running. Maybe a slip through from a previous change?

packages/web/src/createVariable.ts
```diff
@@ -26,13 +27,19 @@ export const createVariable = <A extends string | number | Variable = any>(
   skipHash = false
 ): Variable<A> => {
   if (!skipHash && isVariable(props)) return props
+
   const { key, name, val } = props
+  const _prefix = process.env.TAMAGUI_CSS_VARIABLE_PREFIX || CSS_VARIABLE_PREFIX
```
natew, line 32: can you move this into CSS_VARIABLE_PREFIX?

natew, line 32: no need to `_prefix` just `prefix`

packages/web/src/createVariable.ts
```diff
@@ -90,5 +97,6 @@ export const createCSSVariable = (nameProp: string, includeVar = true) => {
     }
   }
   const name = simpleHash(nameProp, 60)
-  return includeVar ? `var(--${name})` : name
+  const _prefix = process.env.TAMAGUI_CSS_VARIABLE_PREFIX || CSS_VARIABLE_PREFIX
+  return includeVar ? `var(--${_prefix}${name})` : name
```
natew, line 101: can we unify this logic into a helper since its done twice?

natew:
lgtm i'm going to wait to merge until we are ready for another minor release shouldn't be too long

Issue #2300, opened 2024-02-19
Wrong typing for color tokens in variant functions


arekkubaczkowski:
### Current Behavior

There is a wrong typing for color tokens inside of the variant function:

Create a following styled component:
```
const CheckboxSelectorFrame = styled(View, {
  variants: {
    error: {
      true: (_, { props, tokens }) => ({
        borderColor: tokens.color['$color-border-destructive-bold-hover'], // this works but it's not consistent with TS
      }),
      false: {
          backgroundColor: '$color-bg-destructive-default', // this works perfectly, why can't the above scenario work the same way?
      },
    },
  } as const,
});
``` 

<img width="744" alt="Screenshot 2024-02-19 at 11 20 29" src="https://github.com/tamagui/tamagui/assets/18654447/e83eb748-f783-44dc-ab0c-bcfefd88c678">

TypeScript suggests the color tokens under the tokens.color prop without the '$' sign, which appears to be incorrect as that object contains the color keys with the '$' sign included. Adding the '$' sign fixes it. However, TypeScript still doesn't complain for some reason, which means that you can provide whatever key you want and TypeScript won't throw an error.

Question:
Why the function variant cannot work the same way as inline variants, so you can pass token string without accessing from  any additional object?

### Expected Behavior

Ideally if tokens would be available directly as a string values instead of getting it from the object that comes from the upper function

### Tamagui Version

```markdown
"tamagui": "1.89.20"
```

### Platform (Web, iOS, Android)

```markdown
Web/iOS/Android
```

### Reproduction

```markdown
-
```

arekkubaczkowski:
@TheEhsanSarshar any update on this by chance? Can I help you somehow to get that fixed?

TheEhsanSarshar:
@arekkubaczkowski this PR https://github.com/tamagui/tamagui/pull/2398 should fix it

arekkubaczkowski:
@TheEhsanSarshar btw. is there any way to get font tokens of out react context similarly how getToken is doing it? as getToken has not included fonts for some reason

TheEhsanSarshar:
there is a `getFontSize()` function
```tsx
import { getFontSize } from '@tamagui/font-size'
```

arekkubaczkowski:
@TheEhsanSarshar nice, but what about the rest of properties like weight, lineheight?  is it accessible somehow?

arekkubaczkowski:
if not, I can give a try to implement it based on getFontSize helper. Let me know

[...]

PR #2398, opened 2024-03-07
fix(core:types): variants tokens

TheEhsanSarshar:


arekkubaczkowski:
@natew  @TheEhsanSarshar I assume that wasn't correct approach, however can we expect other/better solution?

arekkubaczkowski:
By the way, guys, you're doing a really great job, and I truly appreciate it! Someday, I'm going to delve into the implementation details so I can help make this library even better 🤞

TheEhsanSarshar:
oh, awesome that would be great! thanks ❤️
btw some changes to fix issue in landed, let me know if that fixed this issue

arekkubaczkowski:
@TheEhsanSarshar I am still driving this issue, in which version this was supposed to be fixed?

natew:
I did merge a different fix, can you get me a reproduction for what you're seeing?

arekkubaczkowski:
@natew 

https://github.com/tamagui/tamagui/issues/2300

There is an issue reported already

arekkubaczkowski:
In the latest version tokens are properly prefixed already but unfortunately suggesting still doesn't for for inline way.

@natew 
Screenshot 2024-03-13 at 15 48 52

PR #2404, opened 2024-03-08
fix(web): Fix adding extra style elements for each `updateFont` call with same name argument

Mimameid:
Updating fonts (calling `updateFont`) in runtime would call `insertFont` which again inserts a new style element each time. This change ensures `insertFont` adds a single new style element given the same name parameter.

### DIFF
--- a/packages/web/src/insertFont.ts
+++ b/packages/web/src/insertFont.ts
@@ -10,36 +10,17 @@ import type { GenericFont } from './types'
 /**
  * Runtime dynamic insert font
  */
-export function insertFont<A extends GenericFont>(
+function insertFont<A extends GenericFont>(
   name: string,
   fontIn: A
-): DeepVariableObject<A> {
-  const styleElement = document.createElement('style')
-  return processFont(name, fontIn, styleElement)
-}
-
-/**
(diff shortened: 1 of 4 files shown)


packages/web/src/insertFont.ts
```diff
@@ -18,10 +19,12 @@ export function insertFont<A extends GenericFont>(
...
     const fontVars = registerFontVariables(parsed)
-    const style = document.createElement('style')
-    style.innerText = `:root .font_${name} {${fontVars.join(';')}}`
-    style.setAttribute('data-tamagui-font', name)
-    document.head.appendChild(style)
+    const styleElement: HTMLStyleElement =
+      document.querySelector(`style[${FONT_DATA_ATTRIBUTE_NAME}="${name}"]`) ||
+      document.createElement('style')
+    styleElement.innerText = `:root .font_${name} {${fontVars.join(';')}}`
+    styleElement.setAttribute(FONT_DATA_ATTRIBUTE_NAME, name)
```
Mimameid, line 26: Could ensure that the attribute is not set anew if the style element existed. Not sure what the value here is though.

PR #2406, opened 2024-03-08
fix(core): better way handle focus visible

TheEhsanSarshar:


packages/web/src/createComponent.tsx
```diff
@@ -334,7 +334,7 @@ export function createComponent<
     // [animated, inversed]
     // HOOK
     const stateRef = useRef<TamaguiComponentStateRef>({
-      handleFocusVisible: true,
+      handleFocusVisible: false,
```
natew, line 337: oh btw this object shouldn't be here, another tiny performance win just to leave it out and only set it when needed

packages/web/src/createComponent.tsx
```diff
@@ -936,13 +936,26 @@ export function createComponent<
       )
     const needsPressState = Boolean(groupName || runtimeHoverStyle)
 
+    if (isWeb) {
```
natew, line 939: hm we really can't afford to add hooks for performance (remember this runs on every single component which is a ton), can we join this with the main useEffect that does a bunch of other things? seems like it could fit there.

TheEhsanSarshar: in that hook we are subscribing to group state. so not sure if it's possible to put it there.
maybe adding one more guard here can help` if(isWeb && focusVisibleRuntimeStyle)`

natew: I think we can or else we can use an external listener and just add/remove from a map

TheEhsanSarshar: used global event handler for keep tracking the keyboard interaction

PR #2416, opened 2024-03-09
bento: style: UsersTable on mobile web

bidah:


natew:
with the new release we should support media queries in contentContainerStyle - but actually i'd almost recommend making an inner wrapping view to wrap the content instead of doing it there.

Also useGroupMedia looks like an anti-pattern generally whne you can use tokens, its not SSR safe.

bidah:
no merge conflicts now @natew . Also removed the changes to `_Showcase` for this PR since they where non related. I will bring up a new PR to apply the changes you mentioned to have a better Showcase component for mobile.

PR #2418, opened 2024-03-09
bento: style: better styling for basic table and with pagination on desktop

bidah:
- [x] header with rounded top corners to align with rounded borders theme
- [x] chevron icons to the right, smaller and grayed out
- [x] hover for table rows


natew:
Styling can keep being improved here imo, but good start. Text in the headers seems too big, borders maybe too strong.

PR #2419, opened 2024-03-09
fix(Button): not animating correctly on web

TheEhsanSarshar:


natew:
Can you give me some context - link to the issue?

TheEhsanSarshar:
@natew basically `useProp()` normalize transforms and that will break animation

PR #2424, opened 2024-03-11
chore(dialog): remove cryptic console warn

gijosso:
I do not feel like it is supposed to warn a message, feel free to change the content of the warning if necessary


TheEhsanSarshar:
Thanks!

PR #2427, opened 2024-03-11
fix(themeInverse): delay issue on native

TheEhsanSarshar:


natew:
we may need more logic here, to avoid the re-render?

PR #2434, opened 2024-03-13
feat(popover): imperative close

arekkubaczkowski:
Allow to close popover imperatively. It's useful when you don't want to take over the control over `open` value and want to close it manually at some point.

### DIFF
--- a/packages/popover/src/Popover.tsx
+++ b/packages/popover/src/Popover.tsx
@@ -563,6 +563,7 @@ type Rect = {
 
 export type Popover = {
   anchorTo: (rect: Rect) => void
+  close: () => void
 }
 
 export const Popover = withStaticProperties(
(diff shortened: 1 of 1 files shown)


arekkubaczkowski:
@natew @TheEhsanSarshar would be grateful for review and release

natew:
We'd probably want to just expose setOpen right? Close is good too but I think both work. Maybe toggle is good too.

arekkubaczkowski:
@natew thanks, fixed

arekkubaczkowski:
@natew, is there a smart way to manage popover behavior triggered by text input focus? I'm aware that technically, we can take control of the 'open' state and toggle it on blur/focus events, but this approach comes with trade-offs. For instance, we'd need to manually handle situations where the user interacts with content inside the popover to prevent it from closing unexpectedly.

TheEhsanSarshar:
I think having both `close` and `open` is better probably. to have more control ?

natew:
Yes, we should have open/close/toggle!

arekkubaczkowski:
> I think even setOpen/open/close/toggle! May as well since they are all very short.

good point, done!

TheEhsanSarshar:
lgtm, Thanks!

PR #2438, opened 2024-03-13
fix(site): moved guides to intro/installation tabs

SamuraiF0x:
- intro/introduction: centered "Quick start" copy button
- intro/installation: added tabs (WideTabs) for in depth guides
- core/introduction: fixed segmented text


natew:
Added some vercel comments

PR #2443, opened 2024-03-14
feat(ListItem): add new composition api

byteab:
- [x] implementation
- [x] doc


natew:
a few things: 

1. your variants were missing `as const`, fixed one
2. we should have `unstyled` prop for everything that passes through context
3. the Button does this but i think its done differently
4. if were doing this change, we should also do the headless conversion!

TheEhsanSarshar:
ok thanks, I'll fix those

TheEhsanSarshar:
@natew docs will be available with another PR

apps/site/data/docs/components/list-item/1.92.0.mdx
```diff
@@ -0,0 +1,137 @@
...
+ListItem is already installed in `tamagui`, or you can install it independently:
+
+```bash
+npm install @tamagui/list-item
+```
+
+## Anatomy
+
+```tsx
+import { Card } from 'tamagui' // or '@tamagui/card'
```
bidah: bad import

PR #2464, opened 2024-03-21
feat(adapt): ios/android disctinction

arekkubaczkowski:
I find it very useful to have the ability to distinguish between iOS and Android separately.
Technically it can be achieved by `isIos` and `isAndroid` but JSX declarative way is way much more readable and consistent.


arekkubaczkowski:
cc @natew @TheEhsanSarshar

jakubdrozdek (approved):

natew:
Looks good, I'll wait to merge until we're ready to do a minor release.

Issue #1707, opened 2023-09-25
Icons not inheriting group styles


jbean96:
### Current Behavior

Icons don't inherit group styles `$group-X-hover` like other components do. 

### Expected Behavior

Icons behave the same way as other components and can inherit the group styles.

I understand that this likely would have to do with the implementation of `themed`. Ours is very much derived from Tamaguis so I figured I'd verify that it was a Tamagui issue.

### Tamagui Version

```markdown
1.71.0
```

### Platform (Web, iOS, Android)

```markdown
Web
```

### Reproduction

```markdown
- Clone [this repo](https://github.com/jbean96/tamagui-icon-groups)
- Setup:
  - `yarn`
  - `yarn watch:build`
  - `yarn sandbox`
- Visit `localhost:9000`
- Hover over box in the center, observe that the icon doesn't turn white like the text even though they are configured the same.
```

### System Info

```markdown
System:
    OS: macOS 13.5.1
    CPU: (10) arm64 Apple M1 Pro
    Memory: 247.50 MB / 32.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 19.6.0 - ~/.nvm/versions/node/v19.6.0/bin/node
    Yarn: 3.5.0 - /opt/homebrew/bin/yarn
    npm: 9.4.0 - ~/.nvm/versions/node/v19.6.0/bin/npm
    Watchman: 2023.06.12.00 - /opt/homebrew/bin/watchman
  Browsers:
    Chrome: 116.0.5845.187
    Firefox: 114.0.1
    Safari: 16.6
  npmPackages:
    @biomejs/biome: ^1.2.2 => 1.2.2
    @manypkg/cli: ^0.19.1 => 0.19.2
    @nderscore/tamagui-typescript-plugin: ^0.3.1 => 0.3.1
    @vercel/og: ^0.5.8 => 0.5.8
    check-dependency-version-consistency: ^3.0.3 => 3.0.3
    import-sort-style-module: ^6.0.0 => 6.0.0
    npm-run-all: ^4.1.5 => 4.1.5
    p-map: ^5.5.0 => 5.5.0
    prettier: ^2.8.1 => 2.8.8
    pretty-quick: ^3.1.2 => 3.1.3
    prompts: 2.1.0 => 2.1.0
    turbo: ^1.10.13 => 1.10.13
    typescript: ^5.2.2 => 5.2.2
    ultra-runner: ^3.10.5 => 3.10.5
```

[...]

PR #2472, opened 2024-03-25
feat(core): add group support to usePropsAndStyle

### LINKED ISSUE #1945: Bug: Not possible to subscribe to group pseudo state without trickery (WEB)
### Current Behavior

When using the "group" props feature of tamagui, it's not possible to directly subscribe to group pseudo state changes on the web platform.

Initially, I tried to naively subscribe to the `ComponentContext` from tamagui like this:

```tsx
const { groups } = useContext(ComponentContext.context)
const pseudoState = groups.groupName?.pseudoState
```

But changes to the pseduo state did not trigger re-renders for consumers of this context.

---

After looking a little deeper into the implementation of the group prop feature, I noticed there is a `subscribe` method on the `groups` property of the component context, so I wrote a [custom hook to subscribe to changes based on group name](https://github.com/nderscore/tamagui-starter-free/blob/bug-group-pseudo-subscribe/packages/ui/src/useGroupPseudoState.ts).

Doing this doesn't work on the web though, because there the values of these state changes are optimistically ignored under normal circumstances, when `runtimeHoverStyle`/`runtimePressStyle`/etc are not true here:

https://github.com/tamagui/tamagui/blob/v1.79.6/packages/web/src/createComponent.tsx#L944-L959

To work around this, it's possible to pass empty functions for the `onPress` or `onHoverIn` props, which bypasses the conditions that prevent the values of pseudo-state updates from being set.

---

After resolving that issue, I ran into the issue described by #1937 - subscriptions are not managed independently for each instance of a group, so a hover on one group triggers changes on all groups.

### Expected Behavior

Pseudo state updates should not be hidden on the web platform when group subscribers exist.

Subscribers to group context changes should only receive updates from their parent component, not from any component that implements that group name.

### Tamagui Version

```markdown
1.79.6 and prior
```

```markdown
Web
```

### Reproduction

I've created a minimal reproduction on a fork of the starter template. The branch name is `bug-group-pseudo-subscribe`:

https://github.com/nderscore/tamagui-starter-free/tree/bug-group-pseudo-subscribe

### System Info

_No response_

TheEhsanSarshar:
fix #1945 #1707 
- [x] support pseudo and media subscription
- [x] support CSS rules not runtime
- [x] make sure it won't break anything, need a final test

Some Notes:
if you want to use CSS and no runtime, then you need to pass `{ noClassNames: !isWeb }` as the second parameter of `usePropsAndStyles` if you don't pass it then you need to disable className on the root group container. `    disableClassName` and sometimes it's necessary especially when CSS solution doesn't work like for lucid icons that need to have access to `color` prop directly and won't work with className


packages/web/src/createComponent.tsx
```diff
@@ -1628,3 +1578,84 @@ function getMediaState(
...
+  state: TamaguiComponentState
+  unPress?: Function
+}) => {
+  useEffect(() => {
+    if (disabled) {
+      return
+    }
+
+    if (shouldEnter) {
+      setStateShallow({ unmounted: false })
```
natew, line 1607: this is going to cause an extra render for no reason in groups no?

natew: mounts are only for ssr matching for animations, but maybe we need that too for useProps not sure actually i'd have to look at it

PR #2483, opened 2024-03-28
Feat/use web a11y deprecate react native

TheEhsanSarshar:
Need to merge with other deprecation PRs


natew:
btw we shouldn't remove the code, just deprecate it, once we do v2 we will remove it

TheEhsanSarshar:
@natew because react native supports `aria-*` as well, so those codes are not necessary anymore

natew:
Yea but we can’t just remove code without bumping major version. So we deprecate first but leave it in, then in v2 we remove it.

PR #2484, opened 2024-03-28
Feat/make boolean only group

TheEhsanSarshar:
- [x] implementation
- [x] test on web, css and no css
- [x] test on native
- [x] add doc   
- [x] fix some issues

Example

```ts
<View group>
     <View  $group-hover={{ background: 'red' }}/>
</View>
```

### DIFF
--- a/apps/site/data/docs/intro/styles.mdx
+++ b/apps/site/data/docs/intro/styles.mdx
@@ -88,12 +88,22 @@ type ViewStyle = WithStates & {
   $sm?: WithStates
 
   // add group queries
+  $group-hover?: WithStates
+  $group-focus?: WithStates
+  $group-press?: WithStates
+
+  // add group + container queries
+  $group-sm-hover?: WithStates
+  $group-sm-focus?: WithStates
+  $group-sm-press?: WithStates
+
+  // add named group queries
   $group-tabs?: WithStates
   $group-tabs-hover?: WithStates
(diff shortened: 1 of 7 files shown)


apps/site/data/docs/intro/styles.mdx
```diff
@@ -251,6 +276,26 @@ Now the Text will be white, but only when the View matches the media query
 `sm`. This uses the same media query breakpoints you defined in
 `createTamagui({ media })`. You can see it also works with pseudo styles!
 
+For more advanced usecases, you can use named groups with container queries:
```
natew, line 279:
```suggestion
For more advanced use cases, you can use named groups with container queries:
```

apps/site/data/docs/intro/styles.mdx
```diff
@@ -210,14 +220,29 @@ set to the matching `header` value.
...
-  <Text $group-card-hover={{ color: 'white' }} />
+<View group>
+  <Text $group-hover={{ color: 'white' }} />
 </View>
 ```
 
 Now only when the parent View is hovered the Text will turn white. The allowed
 psuedo modifiers are `hover` (web only), `press`, and `focus`.
 
+For more advanced usecases you can use named groups
```
nandorojo, line 231: I think you should put a heading above this line that says `### Named Groups`

PR #2487, opened 2024-03-29
feat: create web api compatible image

ehxxn:
- [x] new web compatible image implementation
- [x] fix some type issues
- [x] doc


packages/webimage/src/Image.tsx
```diff
@@ -0,0 +1,99 @@
+import { isWeb } from '@tamagui/constants'
+import type { StackProps } from '@tamagui/core'
+import { View, styled } from '@tamagui/core'
+import { forwardRef } from 'react'
+import { Image as RNImage } from 'react-native'
```
natew, line 5: it would be really cool if we only exported a way to configure your own underlying image library, that way not only will it not depend on react-native, but also you can add in support for expo-image or other libraries on native.

we could export a `setupImage` function and they could then do something like:

```tsx
setupImage(props => <Image source={{ uri: props.src }} />)
```

And we could export a separate setup function from `@tamagui/webimage/react-native` that makes it so you can just do `setup(Image)` and it works. And then eventually have `@tamagui/webimage-expo` that has its own setup. Something like that

KittichoteKamalapirat:
Does this PR solve the issue of image not renderring on Next.js? It does render correctly on mobile though.

In the `packages/app` directory, I have the following

```tsx
import myImage from '../../assets/my-image.png'

<Image
            source={{
              uri: myImage,
              width: 200,
              height: 50,
            }}
          />
```

PR #2498, opened 2024-04-02
feat(input): create core input component

### LINKED ISSUE #2495: <Input /> element doesn't pass through the name prop.
### Current Behavior

Hi i've tried both Input from tamagui and TextInput from react-native to pass the name prop through but in both they don't show up or get passed in. This causes FormActions from Next.js to not work because Input elements without `name` don't get registered in form-actions, only way to do so is for name props to be present.

### Expected Behavior

Expected behavior is for the passed in name prop to be passed into and rendered in the HTML.

### Tamagui Version

```markdown
I was on `latest`
```

```markdown
Web
```

### Reproduction

```markdown
import {Input} from 'tamagui';

export const Example = () =>  <Input name="test" />;
```

### System Info

_No response_

TheEhsanSarshar:
- [x] web compatible native Input
- [x] web compatible web input
- [x] doc
- [x] fix font not applying bug

fix #2495


apps/site/components/demos.tsx
```diff
@@ -82,6 +82,11 @@ export const InputsDemo = lazyDemo(
   // @ts-ignore
   import('@tamagui/demos/demo/InputsDemo').then((x) => x.InputsDemo)
 )
+
+export const NewInputsDemo = lazyDemo(
+  // @ts-ignore
+  import('@tamagui/demos/demo/NewInputsDemo').then((x) => x.NewInputsDemo)
```
natew, line 88: realizing these lazyDemo wont be lazy i think because the import isnt in a function and called later right? doesn't it fetch on calling the import? we should change that

packages/helpers/src/validStyleProps.ts
```diff
@@ -181,6 +181,8 @@ export const stylePropsView = {
   ...tokenCategories.radius,
   ...stylePropsTransform,
   ...stylePropsUnitless,
+  lineHeight: true,
```
natew, line 184: these already exist elsewhere right, im trying not to duplicate anything

natew: oh you deleted them from stylePropsFont? why?

packages/input/src/Input.native.tsx
```diff
@@ -0,0 +1,175 @@
+import React, { useEffect } from 'react'
+import { registerFocusable } from '@tamagui/focusable'
+import { styled, useComposedRefs } from '@tamagui/core'
+import { TextInput, NativeSyntheticEvent, TextInputChangeEventData } from 'react-native'
+import { InputProps } from './type'
+import { defaultStyles, inputSizeVariant } from './shared'
+const INPUT_NAME = 'Input'
+const StyledInput = styled(
```
natew, line 8: this should be shared between web and native no?

packages/input/src/Input.tsx
```diff
@@ -0,0 +1,188 @@
...
+      },
+    } as any)
+  })
+
+  useEffect(() => {
+    if (onSelectionChange) {
+      ref.current?.addEventListener('selectionchange', _onSelectionChange)
+    }
+    return () => {
+      if (onSelectionChange) {
```
natew, line 128: there should just be one if that returns its own dispsoe since this is simple

packages/input/src/Input.tsx
```diff
@@ -0,0 +1,188 @@
...
+
+function usePseudoColorApplier(
+  ref: React.RefObject<HTMLInputElement>,
+  placeholderColor?: ColorTokens,
+  selectionColor?: ColorTokens
+) {
+  const theme = useTheme()
+  useEffect(() => {
+    if (placeholderColor) {
+      ref.current?.style.setProperty(
```
natew, line 173: we can't just pass this to style prop declaratively?

packages/input/src/type.ts
```diff
@@ -0,0 +1,58 @@
+import { ColorTokens, StackProps, TamaguiComponentPropsBase } from '@tamagui/web/types'
```
natew, line 1: never import from /types (also lint type import)

packages/input/src/type.ts
natew, line 1: lets name it types.ts for consistency

packages/web/reset.css
natew, line 1: undo formatting

packages/core/reset.css
natew, line 1: lets no do formatting

packages/core/reset.css
```diff
@@ -15,24 +17,53 @@ a {
...
 
-a.is_Button, .is_Anchor:has(.is_Button) {
+a.is_Button,
+.is_Anchor:has(.is_Button) {
   text-decoration: none;
 }
+
+input::selection,
+textarea::selection {
+  background-color: var(--selectionBackground) !important;
```
natew, line 63: this isnt the right reset we can't assume they have certain tokens on their themes, do we need this?

packages/helpers/src/validStyleProps.ts
TheEhsanSarshar: oh because I was trying set that on a View but I think now I set `validStyles` explicitly on the View I can remove these now

packages/core/reset.css
TheEhsanSarshar: that's something that we will set in Input and TextArea via javascript
but yeah `reset` is not the best place to have that

packages/input/src/type.ts
TheEhsanSarshar: oh right, thanks

PR #2514, opened 2024-04-05
docs(site): next js guide updated

SamuraiF0x:
- fixed next js guide


apps/site/components/ProductButton.tsx
```diff
@@ -0,0 +1,112 @@
...
+
+import { BentoIcon } from "./BentoIcon";
+import { BentoLogo } from "./BentoLogo";
+import { NextLink } from "./NextLink";
+import { TakeoutIcon } from "./TakeoutIcon";
+import { TakeoutLogo } from "./TakeoutLogo";
+
+const productSettings = (
+	product: string
+): { ProductLogo: React.ReactNode; ProductIcon: React.ReactNode; colorOffset: number } => {
```
natew, line 14: make sure you have `biome` extension and are formatting files using it

natew:
I don't like these cards here at all, I don't think we need a call out here.

PR #2520, opened 2024-04-08
feat(create-tamagui): deprecate some props and move inside settings

byteab:
- [x] implement add deprecation migrations
- [x] Make sure `onlyAllowShorhands` type are working properly
- [x] test  

### List of deprecated props

- disableRootThemeClass,
- disableSSR,
- maxDarkLightNesting,
- mediaQueryDefaultActive,
- themeClassNameOnRoot,
- shouldAddPrefersColorThemes,
- cssStyleSeparator
- onlyAllowShorthands ( still a tiny issue )
- selectionStyles
- defaultFont


apps/site/data/docs/core/configuration.mdx
```diff
@@ -207,11 +207,10 @@ The `createTamagui` function receives a configuration object:
   components initial values. These merge downwards, so
   `styled(Text, { name: 'Paragraph' })` will get any defaultProps set for
   `Text`.
-- `onlyAllowShorthands` when set to true will ensure that if you define
-  shorthands, they override existing long-form style props.
 
 <Notice>
-  On Android you need to set the `face` option in `createFont` or else fonts won't pick up different weights, due to a React Native restriction.
+  On Android you need to set the `face` option in `createFont` or else fonts
```
bidah, line 212: when reading documentation I have seen that some paragraphs get cut in the middle before end. It seems that could be because of this formatting? this change shouldn't have happend right? just the same text but with a different formatting on the diff. It could be related.

packages/web/src/types.tsx
```diff
@@ -1141,8 +1233,10 @@ export type WithThemeValues<T extends object> = {
 
 export type NarrowShorthands = Narrow<Shorthands>
 export type Longhands = NarrowShorthands[keyof NarrowShorthands]
-
-type OnlyAllowShorthands = TamaguiConfig['onlyAllowShorthands']
+// check if the settings value is not provided then fallback to reading from TamaguiConfig
+type OnlyAllowShorthands = TamaguiSettings['onlyAllowShorthands'] extends boolean
+  ? TamaguiSettings['onlyAllowShorthands']
+  : TamaguiConfig['onlyAllowShorthands']
```
TheEhsanSarshar, line 1239: This fallback for some reason doesn't work

bidah (approved):

apps/site/data/docs/core/configuration.mdx
TheEhsanSarshar: yeah I had prettier enabled, I think prettier did that, Biome doesn't seem to support mdx

TheEhsanSarshar: ok fixed! thanks

natew:
Oh for the paragraphs that's probably a regression because we changed the default text white-space to behave in-line with RN. I thought Paragraph should set it back correctly though, but either way a fix would be to not have "pre" behavior on text as I think markdown isn't supposed to be sensitive to formatting.

PR #2533, opened 2024-04-15
fix(Button): Button.Text name

### LINKED ISSUE #2507: Button Press Style Inconsistency with Nested Text Component
### Current Behavior

When the button component is pressed, it scales down as expected. However, pressing the text within the button causes only the text to scale down, not the entire button. This inconsistent behavior seems to occur due to pressStyle not being applied to the button when the nested text is pressed.

Workaround:
A temporary workaround is to set textProps={{ disabled: true }}, which seems to fix the issue.

https://github.com/tamagui/tamagui/assets/22500853/00bdf24e-d634-48b8-96f5-383ff1e51955

### Expected Behavior

The pressStyle should be applied to the entire button component, including the nested text (even when the text inside of the button get pressed), ensuring consistent press effects.

### Tamagui Version

```markdown
1.94.1
```

```markdown
iOS
```

### Reproduction

```markdown
1- Create a button component with a nested text component.
2- Apply a pressStyle transformation to the button.
3- Observe the behavior when pressing the button directly versus pressing the text within the button.

<Button
onPress={() => onSignin(STRATEGIES.oauth_apple)}
w="$7/8"
mt="$2.5"
h="$4.5"
bg="black"
color="white"
icon={Apple}
>				{t('screens:signin.signinWithApple')}
</Button>

Tamagui config: 
defaultProps: {
 Button: {
  animation: '100ms',
  elevation: '$size.15',
  pressStyle: {
  elevation: '$size.7',
  scale: 0.95,
  },
 },
},
```

### System Info

_No response_

TheEhsanSarshar:
fix #2507

### DIFF
--- a/packages/button/src/Button.tsx
+++ b/packages/button/src/Button.tsx
@@ -152,7 +152,7 @@ const ButtonFrame = styled(ThemeableStack, {
 })
 
 const ButtonText = styled(SizableText, {
-  name: 'Button',
+  name: 'ButtonText',
   context: ButtonContext,
 
   variants: {


TheEhsanSarshar:
there is a tradeoff of having same theme for both component or make default props to work properly!
as `defaultProps` is not a common thing and will be deprecated in near future, we can keep it as it is

PR #2537, opened 2024-04-15
refactor(styled): move name to third argument

byteab:


natew:
We'd need to update docs in a couple places i think, and also we should move all internal usage at the same time!

TheEhsanSarshar:
> We'd need to update docs in a couple places i think, and also we should move all internal usage at the same time!

ohh right! 😀

TheEhsanSarshar:
all internal component and doc are ready

PR #2540, opened 2024-04-16
doc(components): document need for PortalProvider

ehxxn:


natew:
Lets make this a bit friendlier. Lets move it to an install section (if they don't have it lets make it consistent).

No need for any Notice I think just text "In order to use this component independently of `tamagui`, you'll need to add a provider to the root of your app:" and then a code block of the import  + usage example.

PR #2550, opened 2024-04-17
Don't error silently in dev mode for inserting styles

nandorojo:
Errors should throw in dev and (sometimes) be silent in prod, never the opposite.


natew:
Have a fix in https://github.com/tamagui/tamagui/commit/df14f54db75ca0b204e6a4e8d85e0f35d07c0c6d

Issue #2099, opened 2024-01-17
Dialog sheet adoption on native with snapPointsMode="fit"


Maciej-Siara:
### Current Behavior

The problem with Dialog with Sheet adaption on native. When snapPointsMode="fit" is setted on Sheet root component, the Adapt.Contents is rendering in a wrong way:
There is some extra bottom padding added on the bottom of the sheet on Android device. When I am trying to inspect that to check if it is some extra padding which I added by mistake I cannot even inspect that space by clicking on it.
Moreover closing animation is some kind of laggy - Sheet jumps up and then is closing.
Also Sheet content is disappearing while sheet is closing but I assume that is desirable behaviour of the Adapt.Contents

https://github.com/tamagui/tamagui/assets/108800119/fb3c9206-499d-4518-a6a8-7b90cd360757

### Expected Behavior

When snapPointsMode="fit" is given to Sheet component it should work and behave in the same way as snapPointsMode="percent" - No flickering of animations and no extra bottom space on android

https://github.com/tamagui/tamagui/assets/108800119/0e16aa38-dbd4-4c59-b18c-fcdf3fc10bb7

### Tamagui Version

```markdown
v1.86.4
```

### Platform (Web, iOS, Android)

```markdown
iOS, Android
```

### Reproduction

```markdown
https://github.com/Maciej-Siara/tamagui-issue
```

### System Info

```markdown
System:
    OS: macOS 14.2.1
    CPU: (8) arm64 Apple M2
    Memory: 90.94 MB / 16.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 20.5.0 - ~/.nvm/versions/node/v20.5.0/bin/node
    Yarn: 4.0.2 - /usr/local/bin/yarn
    npm: 10.2.5 - ~/.nvm/versions/node/v20.5.0/bin/npm
  Browsers:
    Chrome: 120.0.6099.234
    Firefox: 112.0.1
    Safari: 17.2.1
  npmPackages:
    @babel/runtime: ^7.18.9 => 7.21.0 
    @manypkg/cli: ^0.19.1 => 0.19.2 
    check-dependency-version-consistency: ^3.0.3 => 3.0.3 
    eslint: ^8.21.0 => 8.39.0 
    node-gyp: ^9.3.1 => 9.3.1 
    prettier: ^2.7.1 => 2.8.8 
    turbo: ^1.10.3 => 1.10.3 
    typescript: ^5.3.3 => 5.3.3
[...]

PR #2563, opened 2024-04-21
fix(sheet): trigger close action when sheet is fully closed

ehxxn:
call setOpen(false) when Sheet is fully closed!
fix junky animation in this issue #2099


arekkubaczkowski:
@ehxxn can we count on getting it released? We are having animation issues in the prod app

ehxxn:
@arekkubaczkowski This isn't a perfect fix, I should close this PR
I'll try to find a proper fix over the weekend. or if someone from Tamagui team wants to fix it
as a workaround, you can use setTimeout

```jsx
onOpenChange={(open) => {
     if (!open) {
         setTimeout(() => setOpen(false), 300)
     } else {
        setOpen(true)
     }
}}
```

Issue #2625, opened 2024-05-06
Scroll behavior in Select with long lists


chrisaugustsson:
### Current Behavior

In a longer list of items in the Select, the selected item is not shown when opening the list. Its does not scroll to the selected item.

### Expected Behavior

It should show the selected item

### Tamagui Version

```markdown
Latest
```

### Platform (Web, iOS, Android)

```markdown
Web
```

### Reproduction

If you open the dialog here; https://tamagui.dev/ui/dialog/1.0.0 and shirnk the screen size so you dont have much room at the top. Then select one of the items further down, like Jackfruit or Grape, the content wont scroll down to the selected item.

In my image I have selected Grape and when I open the Select it looks like this:
![2024-05-06 11_53_06-(14) Discord _ #general _ tamagui](https://github.com/tamagui/tamagui/assets/31002572/e1955c51-1772-4800-97b7-329485f3087f)

PR #2629, opened 2024-05-07
Fix select scroll behaviour

chrisaugustsson:
Related to #2625

useListNavigation tries to scroll the item into view when the list opens, witch collides with the useInnerOffset, resulting in items further down the list won't end up in the correct position.

The removal of onScroll event makes the down arrow appear when the list opens, even if you are at the end of the list.


natew:
Thanks for that, great contribution. Any interest in doing some contract work ongoing lmk you can DM me on discord foobully or email me.

PR #2641, opened 2024-05-13
feat(core): add reanimated animation driver

ehxxn:


natew:
btw why did you add this one? moti does seem to be doing well for us

PR #2688, opened 2024-06-01
chore: Add missing import for React in ThemeProvider

Code-Parth:
error at `<Theme>`

```
'React' refers to a UMD global, but the current file is a module. Consider adding an import instead.ts(2686)
```

```tsx
return (
    <Theme
      className={props.className}
      name={props.defaultTheme}
      // if root class disabled, force class here
      forceClassName={!props.disableRootThemeClass}
      // @ts-expect-error
      _isRoot
    >
      {props.children}
    </Theme>
  )
```


natew:
this entirely depends on your tsconfig i believe, we don't see this error internally. how do you trigger the error?

Code-Parth:
> this entirely depends on your tsconfig i believe, we don't see this error internally. how do you trigger the error?

in the facing of this issue https://github.com/tamagui/tamagui/pull/2689. I tackle with this error and solve via importing a react

PR #2703, opened 2024-06-06
Pass Spinner props to ActivityIndicator

tazsingh:
[React Native Web appears to wrap the `ActivityIndicator` in a `View`, which is expecting a `style` prop to display as intended](https://github.com/necolas/react-native-web/blob/master/packages/react-native-web/src/exports/ActivityIndicator/index.js#L56-L77)

This PR passes `stackProps` to `ActivityIndicator` as well. I wasn't sure if it's more optimal to just pass `style` 🤷

Happy to adapt, let me know! Thanks

### DIFF
--- a/packages/tamagui/src/views/Spinner.tsx
+++ b/packages/tamagui/src/views/Spinner.tsx
@@ -23,7 +23,7 @@ export const Spinner: React.ForwardRefExoticComponent<
       }
       return (
         <YStack ref={ref} {...stackProps}>
-          <ActivityIndicator size={size} color={color} />
+          <ActivityIndicator {...stackProps} size={size} color={color} />
         </YStack>
       )
     }),


natew:
I'd almost rather just remove the react-native-web dep on web and do something nicer, our props wont map cleanly to theirs right? We'd have to do something weird like useStyle() etc etc. Likely easier for now to just fork this component yourself and we can revisit. Never did love the way we just wrap Spinner, Tamagui usually wants to have its own implementation and then `native` boolean props to go to native stuff, so we should implement our own eventually.

tazsingh:
Yeah from a Tamagui perspective, that makes sense 💯 

For my current needs, I can simply re-wrap `ActivityIndicator` on our end. It's relatively straightforward to essentially copy/paste/adapt the Tamagui `Spinner.tsx` into what we need. So no rush on doing something different here!

I'll close this PR as there's a plan to resolve the underlying issue eventually 👍

PR #2714, opened 2024-06-09
RFC - feat(web): allow DOM elements

mfazekas:
The goal of the PR is to get something like
```jsx
let StyledDiv = styled('div', {
  backgroundColor: 'violet',
  variants: {
    variant: {
      red: {
        backgroundColor: 'red',
      },
      gray: {
        backgroundColor: 'gray',
      }
    }
  },
  defaultVariants: {
    variant: 'gray'
  }
} as const);
```

working:

See 
https://codesandbox.io/p/sandbox/cranky-resonance-l779n8?file=%2Fsrc%2FApp.tsx%3A27%2C34

Note the backgroundColor is violet with 1.100.3, but it should be 'gray' because of the defaultVariants.

image

There is 2 change: 
1.) add string to StylableComponent
2.) acceptsClassName is true for string components. 

Neither of those are complete changes:
- typescript `InferStyledProps` generic type needs to be implemented for string components
- acceptsClassNames is not really for all string components - custom elements like 'my-element' doesn't accept classNames they access class . See https://react.dev/reference/react-dom/components#custom-html-elements

.

### DIFF
--- a/packages/core-test/getSplitStyles.web.test.tsx
+++ b/packages/core-test/getSplitStyles.web.test.tsx
@@ -104,6 +104,35 @@ describe('getSplitStyles', () => {
     expect(styles.rulesToInsert[0].property).toEqual(`boxShadow`)
   })
 
+  test(`backgroundColor in defaultVariant for div`, () => {
+    const CustomDiv = styled('div', {
+      backgroundColor: 'white',
+      variants: {
+        variant: {
+          red: {
+            backgroundColor: 'red',
+          },
+          green: {
+            backgroundColor: 'green',
+          },
+          gray: {
+            backgroundColor: 'gray',
+          }
+        }
+      },
+      defaultVariants: {
+        variant: 'gray'
+      }
+    } as const)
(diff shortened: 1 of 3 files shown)


ehxxn:
I think you should consider supporting other tags as well if you want to work on this
and also make sure they work on native too
probably the better way could be mapping them to Tamagui components like this

`input` => `Input`
`div` => `View` with tag="div"
`img` => `Image`
`nav` => `View` with tag="nav"
and so on

mfazekas:
> I think you should consider supporting other tags as well if you want to work on this and also make sure they work on native too probably the better way could be mapping them to Tamagui components like this
> 
> `input` => `Input` `div` => `View` with tag="div" `img` => `Image` `nav` => `View` with tag="nav" and so on

FWIW I've added a more complete version of the PR as #2793. 

I don't see the point of adding such mapping, as they can write `styled(Input, ...` if they want cross platform behaviour. Sometimes you want web only code, and `styled('div', ...` etc. should address that.

PR #2723, opened 2024-06-13
Support `size` + `transformOrigin` in `Popper`

nandorojo:


natew:
Thanks for this contribution\! I've incorporated these changes into #2734 with a few adjustments (added `isWeb` guard for the transformOrigin middleware, removed redundant `floating.update()` call, etc). Appreciate the work on this\!

PR #2793, opened 2024-07-15
feat(web): styled('div', ...) support

mfazekas:
Changes:
- treats `keyof JSX.IntrinsicElements` (aka "div" and co as)  StylableComponent
- sets acceptsClassName for intrinsic elements that supports classNames
- implement GetNonStyledProps for `JSX.IntrinsicElements` to return correct props for intrinsic elements (aka "div")

See:
https://react.dev/reference/react-dom/components#custom-html-elements


natew:
Closing as we discussed a better way to do in discord, ty for the effort here though

PR #2796, opened 2024-07-15
add demo button to takeout sidebar

bidah:
visual ref:
image


bidah:
update visual ref @natew 
CleanShot 2024-07-22 at 15 46 08

Issue #2773, opened 2024-07-06
1.101.4 web w/o acceptsClassName: hoverStyle not working (regression)


mfazekas:
### Current Behavior

```tsx
const Button = styled("div", {
   ...
   hoverStyle: { backgroundColor: "lightgray", },
);
```

hoverStyle doesn't apply on 1.101.4 or later, tried up to 1.101.7.

https://github.com/tamagui/tamagui/assets/52435/c6ed48f4-b0b2-478d-8b23-6a93dedbce5d

### Expected Behavior

hoverStyle to apply as before. 

https://github.com/tamagui/tamagui/assets/52435/bf897484-6fca-4083-bbb6-b363fb1fcfba

### Tamagui Version

```markdown
1.101.4 and later (tested up to 1.101.7)
```

### Platform (Web, iOS, Android)

```markdown
Web
```

### Reproduction

```markdown
https://codesandbox.io/p/sandbox/amazing-leaf-zfszt4?file=%2Fsrc%2FApp.tsx%3A18%2C4
```

If I use `debug="verbose"` on 1.101.4 I do see:
```js
events {
    "events": null,
    "attachHover": false,
    "attachPress": false
}
```

on 1.101.3 I do see:
```js
events {
    "events": {},
    "attachHover": true,
    "attachPress": false
}
```

I can work around the issue by adding `acceptsClassName: true` 

### System Info

```markdown
n/a
```

natew:
`styled('div')` is not supported by Tamagui, can you update the repro and make sure you're using the vanilla packages?

mfazekas:
@natew updated the repo I could reproduce with a simple custom component. 

```jsx
function MyComponent({ style, onMouseEnter, onMouseLeave }) {
  console.log(" => style", style, onMouseEnter, onMouseLeave);
  return (
    <div
      style={{ width: 20, height: 20, ...style }}
      onMouseEnter={onMouseEnter}
      onMouseLeave={onMouseLeave}
    ></div>
  );
}

const MyStyledComponent = styled(MyComponent, {
  name: "MyStyledComponent",

  backgroundColor: "red",
  hoverStyle: {
    backgroundColor: "blue",
  },
});
```

This worked `1.101.3`, onMouseEnter and onMouseLeave was passed to the component and style was changed on hover.

On `1.104.2` onMouseEnter, onMouseLeave are not passed to the component.

mfazekas:
Looking at the code:

[...]

PR #2815, opened 2024-07-26
fix(web): avoid classes if acceptsClassName is not set

mfazekas:
Fixes: #2773 

Before refactor:

https://github.com/tamagui/tamagui/blob/148e67ac5976110e9a405318a1db5265878ec69c/packages/web/src/createComponent.tsx#L287-L296

We now missing the !staticConfig.acceptsClassName.

This pr attempts to restore that condition

Used this component to verify that works on sandbox:

```tsx
function MyComponent({ style, onMouseEnter, onMouseLeave }) {
  console.log(" => style", style, onMouseEnter, onMouseLeave);
  return (
    <div
      style={{ width: 20, height: 20, ...style }}
      onMouseEnter={onMouseEnter}
      onMouseLeave={onMouseLeave}
    ></div>
  );
}

const MyStyledComponent = styled(MyComponent, {
  name: "MyStyledComponent",

  backgroundColor: "red",
  hoverStyle: {
    backgroundColor: "blue",
  },
});

export default function StyledHTMLTags() {
  return (<MyStyledComponent debug={"verbose"} />);
}
```

### DIFF
--- a/code/core/web/src/createComponent.tsx
+++ b/code/core/web/src/createComponent.tsx
@@ -299,7 +299,7 @@ export const useComponentState = (
     const isDisabledManually =
       disableClassName && !isServer && didHydrateOnce && state.unmounted === true
 
-    if (isAnimatedAndHydrated || isDisabledManually) {
+    if (isAnimatedAndHydrated || isDisabledManually || !staticConfig.acceptsClassName) {
       shouldAvoidClasses = true
 
       // debug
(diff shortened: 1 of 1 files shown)


natew:
Hm so this logic actually we force all components to be css on first render on client on web, so we can do proper SSR, i'll have to tweak this logic and see if its not actually correct.

Issue #2739, opened 2024-06-22
Sheet with moveOnKeyboardChange animation delay in iOS


ex-git:
### Current Behavior

When a sheet with input triggered keyboard, the move to avoid keyboard delay like 1 sec in iOS, same delay also happen when keyboard is dismissed. In Android it is smother.

### Expected Behavior

Sheet move with minimum delay when keyboard is shown or dismissed

### Tamagui Version

```markdown
v1.101.0
```

### Platform (Web, iOS, Android)

```markdown
iOS
```

### Reproduction

```markdown
Just using the sample code from https://tamagui.dev/ui/sheet#sheet-1 with moveOnKeyboardChange
```

### System Info

```markdown
System:
    OS: Linux 6.1 Debian GNU/Linux 12 (bookworm) 12 (bookworm)
    CPU: (20) x64 12th Gen Intel(R) Core(TM) i7-12700K
    Memory: 2.85 GB / 31.03 GB
    Container: Yes
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 20.13.1 - ~/.nvm/versions/node/v20.13.1/bin/node
    Yarn: 4.1.0 - ~/.nvm/versions/node/v20.13.1/bin/yarn
    npm: 10.5.2 - ~/.nvm/versions/node/v20.13.1/bin/npm
    bun: 1.1.9 - ~/.bun/bin/bun
  npmPackages:
    @babel/runtime: ^7.24.6 => 7.24.7 
    @manypkg/cli: ^0.21.4 => 0.21.4 
    check-dependency-version-consistency: ^4.1.0 => 4.1.0 
    eslint: ^9.3.0 => 9.5.0 
    immer: ^10.1.1 => 10.1.1 
    node-gyp: ^10.1.0 => 10.1.0 
    prettier: ^3.2.5 => 3.3.2 
    turbo: ^1.13.3 => 1.13.4 
    typescript: ^5.4.5 => 5.5.2 
    use-immer: ^0.10.0 => 0.10.0
```

giacomocerquone:
Agree on iOS being painfully slow for the user and thank you for opening this issue.

Right now exactly because of this, as a workaround, I'm using a Sheet with a `snapPointsMode="constant"` and then manually setting listeners on the keyboard as well as setting a dynamic snapPoints value.

This is because apparently when you use a state to set the sheet heights' value through the `snapPoints` prop, the changes are tracked way faster both on iOS and android.

[...]

PR #2819, opened 2024-07-27
fix(sheet): use willShow instead of didShow on iOS when moveOnKeyboardChange is true

giacomocerquone:
working on #2739 

the implementation where I am at is still not ok but things are improving.
There are a couple of issues:

1. a race condition that I'll focus on later on
2. a missed fade out animation on the last part that's clearly visible in the recording above

I'm still figuring out how the animation was composed, so bear with me. Any suggestion is welcomed of course

EDIT: Also, willHide and willShow are not available on android so at this point the implementation should be platform based


giacomocerquone:
I hope I got a nice working solution which satisfies the stakeholders here.

The logic was the following (some variables names are fake to make this simpler).

1. The sheet is naturally at the top of the screen and then pushed below screen with translateY when hidden
2. When the sheet is opened, for mode=fit for example, it just does `screen.height - sheetContent.height`. 
3. After keyboard appearance (`didShow`), the code interprets `animatedNumber.getValue()`, which is the variable responsible for animating the sheet's position throughout its lifetime, as the sheets' height and does: `animatedNumber.getValue() - keyboard.height` to lift the sheet up.

Now to the problem:

1. didShow on iOS fires too late... but really late 😅 
2. Using willShow would lead to race conditions if the sheet is closed and reopened quickly because  animatedNumber.getValue() may be accessed when the sheet is not fully opened yet, take a look:

https://github.com/user-attachments/assets/a33660fd-dab5-4e99-ad49-29e08034a467

3. So the solution I found (which I hope to be resilient enough) is to reuse: `isHidden || position === -1 ? screenSize : positions[position]` to obtain the correct sheet's height independent of its animated position while continuing using willShow (iOS-only since on android there is no such event).
Take a look:

https://github.com/user-attachments/assets/69d6146a-234b-4fea-8a88-7450d63b5af8

natew:
Very nice, I wonder if we can't switch to a linear animation rather than a spring as well in the future for only when its making that transition, that way it won't have that little delay/bounce which feels a bit off. But we can follow on with that, going to check the code here now.

code/ui/sheet/src/SheetImplementationCustom.tsx
```diff
@@ -380,24 +380,31 @@ export const SheetImplementationCustom = themeable(
...
+            type: 'spring',
+            ...animationConfig
+          }
         )
       })
       const keyboardDidHideListener = Keyboard.addListener('keyboardDidHide', () => {
         if (sizeBeforeKeyboard.current === null) return
-        animatedNumber.setValue(sizeBeforeKeyboard.current)
+        animatedNumber.setValue(sizeBeforeKeyboard.current, {
+          type: 'spring',
```
natew, line 397: if they are using a non spring animation will this break?

giacomocerquone: I'm sorry, but I think I'm lacking the context here.
They, who are you referring to? The users or the developers using the Sheet component?

I put there a spring animation only to uniform all the sheet's animations' types, but I didn't think it could have the power to break something in some cases

giacomocerquone:
> Very nice, I wonder if we can't switch to a linear animation rather than a spring as well in the future for only when its making that transition, that way it won't have that little delay/bounce which feels a bit off. But we can follow on with that, going to check the code here now.

I totally agree and thought the same to be honest. But I couldn't find any linear animation with VS Code code completion. Do we have any already in the project?

natew:
I think if you do type: 'timing' at least for reanimated and react native animated that would work, but i'd need to test it out

giacomocerquone:
> I think if you do type: 'timing' at least for reanimated and react native animated that would work, but i'd need to test it out

Yeah I actually remember trying it but didn't know what duration to put.
I'm testing with 250 and it seems very nice even though I don't like putting arbitrary values like this. Any advice?


https://github.com/user-attachments/assets/aa712db4-a514-4fe6-9088-9d5c7056b3c1

code/ui/sheet/src/SheetImplementationCustom.tsx
giacomocerquone: Oh now I understand what you meant. Basically if from the outside a different animationConfig that doesn't play well with the animation's type spring, it will probably break.

The thing is though that I just took it from here: https://github.com/tamagui/tamagui/blob/7c3feea1bf6b5039a7dfb3cd1e0e5a2fb5d2aa6d/code/ui/sheet/src/SheetImplementationCustom.tsx#L186-L198

does this also need to be checked?

natew:
Probably just a bit longer, 300-350 maybe but yea it looks pretty good.

PR #2839, opened 2024-08-08
Fix the accordion losing height

zzorba:
As reported on Discord (https://discord.com/channels/909986013848412191/909986013848412194/1242282248754040883) and in our own experience, tapping the Accordion rapidly causes the content to be rendered, but the height to not expand correctly.

An inspection of the code seemed to show that the laid out height of the children was being overridden by the dynamic 'open' height, causing the state to be lost after toggling.

This fixes it to make the height state only contain the laid out height, leaving the animation 'open' height as a purely derived value.


natew:
@jonsherrard looks good but needs a more thorough review, curious if we need that internal height still, but haven't looked a the source.

PR #2840, opened 2024-08-08
fix: external link error

RSS1102:
When I try to access `npm create tamagui@latest` *Check out the source of the templates* , get a 404 error occurred. I believe the new template link should be this one.


natew:
thank you

PR #2858, opened 2024-08-16
fix: allow custom color type for input

CorentinAndre:
I had issue setting the color of the Input component exported from `@tamagui/input`.
The typing for the color prop comes from `DetailedInputProps['style']` which is the basic HTML color type. It's not completely true since it accepts custom theme values coming from tamagui. 

What I've done is simply exclude the color property from `DetailedInputProps['style']` to use the one coming from `TextProps` from `@tamagui/web`.


natew:
thank you

PR #2859, opened 2024-08-16
fix: add missing deps in @tamagui/input

CorentinAndre:
@tamagui/focusable & @tamagui/core are required as deps from @tamagui/input. Importing it without one of those would crash the app.


natew:
@CorentinAndre thanks! if you can update to new version, sorry for churn, if not i'll do it later this weekend

PR #2860, opened 2024-08-16
Fix typo in Select docs

SchnTgaiSpock:
The docs mention using `SheetContents` or `Select.SheetContents` to insert content into the Adapt, but it should be `Adapt.Contents` or `Select.Adapt.Contents`


natew:
good catch, thanks

PR #2867, opened 2024-08-21
fix: design system doc links

victoralmeidadev:
Actually the links on design system documentation page are pointing to a invalid address on repo.

How to reproduce:

- Access https://tamagui.dev/docs/guides/design-systems
- Click on "create-tamagui is itself a well structured monorepo." or "More complete, the tamagui package itself." line.


natew:
ty!

PR #2876, opened 2024-08-25
Pass props to inner View of Accordion.HeightAnimator

tazsingh:
I'm using the `Accordion` component and found that I need to style the inner `View` for the `HeightAnimator`, as its position is `absolute`.

This PR primarily serves as a proposal for one way to do it. I'd be happy to adapt this as directed or for another approach to be implemented. Let me know how I can help!

Thanks again! ❤️

### DIFF
--- a/code/ui/accordion/src/Accordion.tsx
+++ b/code/ui/accordion/src/Accordion.tsx
@@ -8,7 +8,7 @@ import type { H3 } from '@tamagui/text'
 import { H1 } from '@tamagui/text'
 import { useControllableState } from '@tamagui/use-controllable-state'
 import { useDirection } from '@tamagui/use-direction'
-import type { GetProps, GetRef, Stack, TamaguiElement } from '@tamagui/web'
+import type { GetProps, GetRef, Stack, TamaguiElement, ViewProps } from '@tamagui/web'
 import { View, useEvent } from '@tamagui/web'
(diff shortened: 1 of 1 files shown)


byteab:
Wondering if we can do that in a `composition` way

btw do you have any specific use case for that ?

tazsingh:
Thanks for the quick response!

I found I needed it to pass `width='100%'` on this page in a mobile viewport: https://guild.host/gitnation/events . You can see that if I remove it here, the child does not fill the width.

Happy to either adapt my app or do this another way in Tamagui. Let me know how I can help!
image

byteab:
I think alongside position absolute, we should also make the inner View always be 100% of the outer View,  probably by setting `left, right, top, bottom` to 0, or `inset={0}` should also work
that should fix the issue, as the inner View will always follow the dimension of the outer View

would you like to try that and let me know?

tazsingh:
Yes agreed that it should be 100% of the outer `View`, however the purpose of that inner `View` is basically to measure the `height` via the `onLayout` to then animate.

Thus, setting `top` or `bottom` would interfere with that. Between settng `left` & `right` to `0` or setting `width` to `'100%'`, is there a preference?

Let me know and I'll update the PR with that!

byteab:
ahh!! yeah I see, that makes sense
so in this case `100%` is better I think

PR #2894, opened 2024-09-03
fix: use default font from tamagui config, if provided, before using '$body'

wkoutre:
## Current Behavior

Currently, if you don't use the `body` key in `fonts` passed into `createTamagui`, and then try to use `Button` from `tamagui` without setting its `fontFamily` prop, `tamagui` will throw an error:

```
TypeError: Cannot read property 'size' of undefined
```

This is because `getFontSizeVariable` and `getFontSizeToken` fallback to using `'$body'` if no `font` prop is provided (in this case, it would be via `Button`'s `fontFamily` prop).

## Expected Behavior

The `config.settings.defaultFont` value, if present, is used as a default fallback value when using `Button` from `tamagui`.

## Platform (Web, iOS, Android)

iOS, Android

## Tamagui Version
1.108.4

## Example

```
// tamaguiConfig.ts

import { createTamagui } from 'tamagui';

const tamaguiConfig = createTamagui({
  ...
  fonts: {
    primary: returnOfCreateFont
  },
  settings: {
    ...,
    defaultFont: 'primary',
  },
});

type TamaguiConfig = typeof tamaguiConfig;

declare module 'tamagui' {
  // overrides TamaguiCustomConfig so your custom types
  // work everywhere you import `tamagui`

  // eslint-disable-next-line @typescript-eslint/no-empty-interface
  interface TamaguiCustomConfig extends TamaguiConfig {}
}

export default tamaguiConfig;
```

```
// MyButton.tsx

import { Button } from 'tamagui';

const MyButton = ({ children, ...props }: MyButtonProps) => (
  <Button {...props}>{children}</Button>
);
```

### DIFF
--- a/code/core/font-size/src/getFontSize.ts
+++ b/code/core/font-size/src/getFontSize.ts
@@ -1,17 +1,16 @@
 import type { FontSizeTokens, FontTokens, TamaguiInternalConfig } from '@tamagui/core'
 import { getConfig, isVariable } from '@tamagui/core'
+import { getSetting } from "@tamagui/web/types/config"
 
 type GetFontSizeOpts = {
   relativeSize?: number
   font?: FontTokens
 }
 
-const getFontKey = (conf: TamaguiInternalConfig, opts?: GetFontSizeOpts) => {
+const getFontKey = (opts?: GetFontSizeOpts) => {
   const FALLBACK_FONT_KEY = '$body'
 
-  const tokenizedDefaultFont = conf.settings?.defaultFont ? `$${conf.settings.defaultFont}` : undefined;
-
-  return opts?.font || tokenizedDefaultFont || FALLBACK_FONT_KEY
+  return opts?.font || getSetting("defaultFont") ? `$${getSetting("defaultFont")}` : undefined || FALLBACK_FONT_KEY
 }
 
 export const getFontSize = (
@@ -34,7 +33,7 @@ export const getFontSizeVariable = (
     return inSize
   }
   const conf = getConfig()
-  return conf.fontsParsed[getFontKey(conf, opts)].size[token]
+  return conf.fontsParsed[getFontKey(opts)].size[token]
 }
 
 export const getFontSizeToken = (
@@ -47,7 +46,7 @@ export const getFontSizeToken = (
   // backwards compat
   const relativeSize = opts?.relativeSize || 0
   const conf = getConfig()
-  const fontSize = conf.fontsParsed[getFontKey(conf, opts)].size
+  const fontSize = conf.fontsParsed[getFontKey(opts)].size
   const size =
     (inSize === '$true' && !('$true' in fontSize) ? '$4' : inSize) ??
     ('$true' in fontSize ? '$true' : '$4')
--- a/code/core/font-size/src/getFontSize.ts
+++ b/code/core/font-size/src/getFontSize.ts
@@ -7,9 +7,9 @@ type GetFontSizeOpts = {
   font?: FontTokens
 }
 
-const getFontKey = (opts?: GetFontSizeOpts) => {
-  const FALLBACK_FONT_KEY = '$body'
+const FALLBACK_FONT_KEY = '$body'
 
+const getFontKey = (opts?: GetFontSizeOpts) => {
   return opts?.font || getSetting("defaultFont") ? `$${getSetting("defaultFont")}` : undefined || FALLBACK_FONT_KEY
 }
 
--- a/code/core/font-size/src/getFontSize.ts
+++ b/code/core/font-size/src/getFontSize.ts
@@ -9,7 +9,9 @@ type GetFontSizeOpts = {
 const getFontKey = (conf: TamaguiInternalConfig, opts?: GetFontSizeOpts) => {
   const FALLBACK_FONT_KEY = '$body'
 
-  return opts?.font || conf.settings?.defaultFont || FALLBACK_FONT_KEY
+  const tokenizedDefaultFont = conf.settings?.defaultFont ? `$${conf.settings.defaultFont}` : undefined;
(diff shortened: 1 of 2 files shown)


code/core/font-size/src/getFontSize.ts
```diff
@@ -1,11 +1,19 @@
...
 
 type GetFontSizeOpts = {
   relativeSize?: number
   font?: FontTokens
 }
 
+const getFontKey = (conf: TamaguiInternalConfig, opts?: GetFontSizeOpts) => {
+  const FALLBACK_FONT_KEY = '$body'
+
+  const tokenizedDefaultFont = conf.settings?.defaultFont ? `$${conf.settings.defaultFont}` : undefined;
```
natew, line 12: this should be inlined in the || tokenizedDefaultFont line so it avoids running if opts?.font is set, also FALLBACK_FONT_KEY should be hoisted above

code/core/web/src/helpers/propMapper.ts
```diff
@@ -418,7 +418,7 @@ export const getTokenForKey = (
         case 'lineHeight':
         case 'letterSpacing':
         case 'fontWeight': {
-          const defaultFont = conf.defaultFont || '$body'
+          const defaultFont = conf.defaultFont || (conf.settings?.defaultFont ? `$${conf.settings.defaultFont}` : undefined) || '$body'
```
natew, line 421: instead of this we should use `getSetting` helper, also in the other file, it should pick up the setting properly.

wkoutre: Sounds good. Also, despite the return type of `getSetting`. I'm still guarding for `getSetting("defaultFont")` potentially being `undefined` as I'm fairly certain `conf!.settings["defaultFont"] || conf["defaultFont"]`  could still actually be `undefined`. If not, let me know.

code/core/font-size/src/getFontSize.ts
```diff
@@ -1,11 +1,18 @@
-import type { FontSizeTokens, FontTokens } from '@tamagui/core'
+import type { FontSizeTokens, FontTokens, TamaguiInternalConfig } from '@tamagui/core'
 import { getConfig, isVariable } from '@tamagui/core'
+import { getSetting } from "@tamagui/web/types/config"
```
natew, line 3: this should import from the above line `@tamagui/core`

wkoutre: When trying to do this:

image

The only auto-detected import is from `'@tamagui/web/types/config'`. Let me know how you'd like to proceed.

code/core/font-size/src/getFontSize.ts
```diff
@@ -1,11 +1,18 @@
...
 
 type GetFontSizeOpts = {
   relativeSize?: number
   font?: FontTokens
 }
 
+const getFontKey = (opts?: GetFontSizeOpts) => {
+  const FALLBACK_FONT_KEY = '$body'
+
+  return opts?.font || getSetting("defaultFont") ? `$${getSetting("defaultFont")}` : undefined || FALLBACK_FONT_KEY
```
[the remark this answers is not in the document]
wkoutre: Sure, if the concern is memory allocation each time this function is run, we could create and store the `defaultFont` with `$` when `createTamagui` is run. In this case, I agree that you're better able to handle that given your understanding of the project.

We could also implement memoization, as a temporary workaround, to hold reference to the result. An example, non-Generic implementation:

```ts
const memoize = (fn: GetFontSize) => {
  const cache = new Map<string, string>();

  const returnFn: GetFontSize = (maybeFontKey) => {
    if (maybeFontKey !== undefined && cache.has(maybeFontKey)) {
      return cache.get(maybeFontKey)!;
    }

    const result = fn(maybeFontKey);

    if (maybeFontKey !== undefined) {
      cache.set(maybeFontKey, result);
    }
    
    return result;
  }

  return returnFn;
};
```

However, my opinion is that these considerations may be over-optimizations at this point in time.

Consider - at a glance, in this file - `getFontSizeToken` and `const sizeTokens = Object.keys(fontSize)`. I understand it's from 3 years ago, and so it likely wasn't and hasn't been a consideration since; but, to me, the same rationale applies: The keys of `fontSize` should be stored when `createTamagui` is run + a helper fn should subsequently be added and used to reference these keys.

Just my take! Plenty of room for improvement, everywhere + I understand the thought process when dealing w/ new contributions (let's get it right, now) vs legacy code (it works; let's improve when measured + deemed appropriate). But, this is a bug impacting 10+ projects for me (my company uses `tamagui` as a base for all new projects) and I'd like to get it fixed!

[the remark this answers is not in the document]
wkoutre: Could it be as straightforward as:
1. adding `defaultFontWithPrefix` to `GenericTamaguiSettings`
2. adding `defaultFontWithPrefix: defaultFont,` to [createTamagui](https://github.com/tamagui/tamagui/blob/master/code/core/web/src/createTamagui.ts)
3. Amending this PR's changes with something like

```ts
return opts?.font || getSetting("defaultFontWithPrefix");
```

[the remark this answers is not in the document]
natew: Yea lets have it store into setting and then use getSetting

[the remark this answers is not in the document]
natew: Actually better just store onto the Parsed config and use that

natew:
ended up just making this change myself since its a good fix, hope you understand. pushing it in a second with the new release.

wkoutre:
> ended up just making this change myself since its a good fix, hope you understand. pushing it in a second with the new release.

Yeah dude! Of course + no worries. Thanks for taking care of it 🤝🏼

PR #2901, opened 2024-09-06
feat(theme): Update color scheme toggle functionality

jonsherrard:
- Replaced `HydrateTheme` and `UserThemeProvider` with `useColorScheme` and `SchemeProvider` in `_layout.tsx`
- Added `ToggleThemeButton` component to handle theme toggling in `ToggleThemeButton.tsx`
- Updated `Header` component to use `ToggleThemeButton` instead of `ThemeToggle` in `Header.tsx`
- Updated `package.json` to include `@vxrn/color-scheme` dependency

Implementation reason: This change introduces a more flexible and user-friendly way to toggle between light, dark, and system color schemes, enhancing the user experience by allowing dynamic theme adjustments.

### DIFF
--- a/code/tamagui.dev/package.json
+++ b/code/tamagui.dev/package.json
@@ -37,6 +37,7 @@
     "@tamagui/one-theme": "1.110.2",
     "@tamagui/use-store": "1.110.2",
     "@vercel/og": "^0.5.8",
+    "@vxrn/color-scheme": "^1.1.210",
     "color2k": "^2.0.2",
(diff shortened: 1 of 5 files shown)


code/tamagui.dev/package.json
```diff
@@ -37,6 +37,7 @@
     "@tamagui/one-theme": "1.110.2",
```
natew, line 37: we can remove this dep

PR #2904, opened 2024-09-06
fix(ui): change cursor default to inherit in ListItemText

sensecodevalue:
The Text components inside the ListItem were previously set to cursor: default, causing an issue where even if the ListItem had cursor: pointer, the cursor would still display as default. This behavior prevented the intended cursor style from being inherited from the parent.

To resolve this, I’ve changed cursor: default to cursor: inherit in ListItemText, allowing the Text components to inherit the cursor style from their parent elements. Additionally, I’ve set cursor: default as the default style for ListItemFrame to maintain consistency where needed.

| AS-IS | TO-BE |
|----------|----------|
| 스크린샷 2024-09-07 오전 2 57 31 | 스크린샷 2024-09-07 오전 2 58 17 |

### DIFF
--- a/code/ui/list-item/src/ListItem.tsx
+++ b/code/ui/list-item/src/ListItem.tsx
@@ -79,6 +79,7 @@ export const ListItemFrame = styled(ThemeableStack, {
         overflow: 'hidden',
         flexDirection: 'row',
         backgroundColor: '$background',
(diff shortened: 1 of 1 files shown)


yujiseok:
ㅋㅋㅋㅋ굿

PR #2907, opened 2024-09-08
feat(core): fallback to dataSet for Animated Views

byteab:
**Animated** views will ignore `data-*` attributes
so we need to fallback to `dataSet`


natew:
merged this with some changes

Issue #2764, opened 2024-07-02
TextArea with useId causes duplicate ID


NerdCowboy:
### Current Behavior

When using the id from React's `useId` on the `Textarea` component, it triggers a duplicate ID error.
![IMG_D0A7EEC8E1AC-1](https://github.com/tamagui/tamagui/assets/757960/c9ecd340-3118-4ea8-a7c5-053bde08bb11)

It's worth noting this only seems to happen with the TextArea component.

i.e.
``` ts
export function Screen() {
  const id = useId()

  return (
    <TextArea id={id} placeholder="Duplicate ID here" />
  )
}
```

### Expected Behavior

To be able to use TextArea without a duplicate ID error

### Tamagui Version

```markdown
1.101.5
```

### Platform (Web, iOS, Android)

```markdown
only tested iOS
```

### Reproduction

```markdown
https://github.com/NerdCowboy/tamagui-textarea-bug
```

### System Info

```markdown
System:
    OS: macOS 14.5
    CPU: (16) x64 Intel(R) Core(TM) i9-9980HK CPU @ 2.40GHz
    Memory: 5.36 GB / 64.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 15.0.0 - ~/.nvm/versions/node/v15.0.0/bin/node
    npm: 7.24.2 - ~/.nvm/versions/node/v15.0.0/bin/npm
    Watchman: 2024.05.06.00 - /usr/local/bin/watchman
  Browsers:
    Brave Browser: 110.1.48.171
    Chrome: 126.0.6478.127
    Safari: 17.5
  npmPackages:
    @babel/runtime: ^7.24.6 => 7.24.6
    @manypkg/cli: ^0.21.4 => 0.21.4
    check-dependency-version-consistency: ^4.1.0 => 4.1.0
    eslint: ^9.3.0 => 9.4.0
    node-gyp: ^10.1.0 => 10.1.0
    prettier: ^3.2.5 => 3.3.0
    turbo: ^1.13.3 => 1.13.4
    typescript: ^5.4.5 => 5.4.5
```

Cauafernandes:
Has anyone resolved this problem?

I'm experiencing the same error.

dBianchii:
Same for me

dBianchii:
I created a PR: https://github.com/tamagui/tamagui/pull/2917

PR #2596, opened 2024-04-28
fix(Input): remove duplicate focus registration

TheEhsanSarshar:
fix #2592

PR #2917, opened 2024-09-13
fix(TextArea): remove duplicate focus registration

dBianchii:
Fixes https://github.com/tamagui/tamagui/issues/2764

so, I just copied the same that fixed the bug for the Input component. It was merged some time ago.
https://github.com/tamagui/tamagui/pull/2596


natew:
Hm ok so useFocusable already does this makes sense, thanks.

Issue #2230, opened 2024-02-08
[Expo][Vercel] Unable to resolve module /vercel/path0/.tamagui/css/...


iamursky:
### Current Behavior

When deploying an expo app to Vercel from the local machine using vercel cli (`vercel --prod`), everything is ok. But when Vercel builds the project via CI/CD on a commit, it fails with the following error:

```
➤ YN0000: │ my-app@workspace:. STDOUT     ➡ [tamagui] built config and components (5809ms):
--
13:46:39.461 | ➤ YN0000: │ my-app@workspace:. STDOUT
13:46:39.461 | ➤ YN0000: │ my-app@workspace:. STDOUT         Config     ./.tamagui/tamagui.config.cjs
13:46:39.461 | ➤ YN0000: │ my-app@workspace:. STDOUT         Components ./.tamagui/tamagui-components.config.cjs
13:46:39.461 | ➤ YN0000: │ my-app@workspace:. STDOUT
13:46:44.953 | ➤ YN0000: │ my-app@workspace:. STDOUT     ➡ [tamagui] output css: ./tamagui-web.css
13:46:44.954 | ➤ YN0000: │ my-app@workspace:. STDOUT
13:46:45.374 | ➤ YN0000: │ my-app@workspace:. STDOUT ✔ Check that packages match versions required by installed Expo SDK
13:46:52.443 | ➤ YN0000: │ my-app@workspace:. STDOUT ✔ Check for legacy global CLI installed locally
13:46:52.905 | ➤ YN0000: │ my-app@workspace:. STDOUT ✔ Check that native modules do not use incompatible support packages
13:46:54.907 | ➤ YN0000: │ my-app@workspace:. STDOUT ✔ Check that native modules use compatible support package versions for installed Expo SDK
13:46:54.914 | ➤ YN0000: │ my-app@workspace:. STDOUT
13:46:54.915 | ➤ YN0000: │ my-app@workspace:. STDOUT Didn't find any issues with the project!
13:46:54.939 | ➤ YN0000: └ Completed in 42s 172ms
13:46:55.120 | ➤ YN0000: · Done in 1m 24s
13:46:56.223 | env: load .env
13:46:56.224 | env: export EXPO_PUBLIC_APPCHECK_SITE_KEY
13:46:57.120 | Starting Metro Bundler
13:46:57.959 |  
13:46:57.960 | ➡ [tamagui] built config and components (1215ms):
13:46:57.960 |  
[...]

PR #2920, opened 2024-09-16
fix(metro-plugin): inline css instead of injecting a file

mrjackdavis:
Firstly, I'm not strongly held to this PR. It's more of a proposal.

I hit https://github.com/tamagui/tamagui/issues/2230 and inferred that writing files to disk isn't supported in metro-transformers.

This just inlines what would've been written to disk.

I have confirmed it correctly bundles for web, but I've not checked native yet

Fixes https://github.com/tamagui/tamagui/issues/2230

### DIFF
--- a/code/compiler/metro-plugin/src/transformer.ts
+++ b/code/compiler/metro-plugin/src/transformer.ts
@@ -1,4 +1,3 @@
-import { outputFile, pathExists } from 'fs-extra'
 import type {
   JsTransformerConfig,
   JsTransformOptions,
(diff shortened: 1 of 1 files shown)


natew:
Couldn't we just write the file sync? This is going to be terrible for performance, having all your CSS actually in JS.

mrjackdavis:
@natew thanks for the quick response

I'm not versed in metro or Tamagui's build systems. I won't pretend to know stuff I don't lol

The changed I made here was based on [this comment](https://github.com/tamagui/tamagui/issues/2230#issuecomment-2235718548),

> I will note that I bumped this all the way from 400 to 10000 ms (10 full seconds) and it did nothing to help. I do not think this is as simple as a basic timing error / classical race condition based on my investigation. I suspect it might be related to filesystem caching issues with how metro-file-map tracks and caches filesystem trees between each Metro transformer worker, but I am not knowledgeable enough about this to easily confirm that hypothesis.

If metro does in fact track the file system in memory, writing to disk will not work unless there is some way to tell metro about the changes.

In lieu of better options and needing a fix, inlining the CSS was the best I could come up with at the time.

It'd be preferable to notify the build system of the new files, but I couldn't find any documentation on the matter.

Without that though, perhaps there are other optimisations we can make to make inline CSS performance manageable? A thought; does metro end up concatenating all the files as it compiles them to bytecode for hermes or js for web?

And finally, looks like you might've merged a rendition of this PR elsewhere? Yay thanks! Feel free to close this if you'd like, I'm not attached

natew:
thank you for the help here, very huge 🙏 should be resolved in latest release, its not ideal but better than breaking.

PR #2940, opened 2024-09-25
Fix issue with cache being compared before input string is trimmed

carlpontus:
Fix issue with cache being compared before input string is trimmed.

This caused styles not being applied if long names in theme was used.


natew:
ahhh this was confusing me, thanks for that. fixed it slightly differently but gave you credit on 9ccf3605a79b5408eda3aeebb6825585133d89cc

PR #2943, opened 2024-09-25
fixes to button blog post

benschac:
general fixes to inconsistencies.

### DIFF
--- a/code/tamagui.dev/data/docs/guides/how-to-build-a-button.mdx
+++ b/code/tamagui.dev/data/docs/guides/how-to-build-a-button.mdx
@@ -101,7 +101,14 @@ export const ButtonFrame = styled(View, {
       },
     },
   } as const,
-
+  // defaultVariants take priority over values passed higher up in the component tree.
+  // example:
+  // <Button.Props size="$lg">
+  //   <Button>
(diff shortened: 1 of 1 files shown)


code/tamagui.dev/data/docs/guides/how-to-build-a-button.mdx
```diff
@@ -101,7 +101,14 @@ export const ButtonFrame = styled(View, {
       },
     },
   } as const,
-
+  // defaultVariants take priority over values passed higher up in the component tree.
```
benschac, line 104: @natew not really sure how to articulate this, or where the best place to put this content is. But, being a bit clearer on prop order / president / which prop wins would be helpful in building the mental model.

natew: we probably should fix this to be more intuitive somehow, i'd have expected this to win 🤔 agree?

benschac: yeah, that makes sense. Feel free to take this over and or merge it. I don't think pedantic changes are worth additional cycles back-and-forth. Generally just trying to call out / help make the concept clearer.

PR #2873, opened 2024-08-23
fix(sheet): fix sheet max content size on translucent android system ui

LiZhequ:
close #2799

PR #2952, opened 2024-09-29
perf(sheet): avoid unnecessary rendering on keyboard chanage

LiZhequ:
In https://github.com/tamagui/tamagui/pull/2873, I changed `Dimensions.get('window').height` to `Dimensions.get('screen').height` to fix the issue of incorrect sheet maxContentSize on Android. Since the screen height doesn’t change when the keyboard appears, it is no longer necessary to listen for keyboard changes here. Additionally, as the sheet component re-renders when the keyboard changes—including all sheet components in the page stack—removing `useKeyboardVisible` improves performance, especially in complex forms with many sheet components.


natew:
I think we'll need to adjust this for web to use window but native use screen, but i can follow with that.

PR #2966, opened 2024-10-04
fix: Remove the sideEffects note from tutorial

roninjin10:
## Side effect fix

The sideEffect note is not true. It's only true for library code not the application code.

Below is from [webpack docs](https://webpack.js.org/configuration/optimization/) but this applies to vite and all other bundlers as well.

image

I kept sideEffect: true in the package.json though because it's still a nice practice since sometimes users may move to a monorepo setup in which subpackages it is true and this allows the copy pasted package.json to be set up by default. There is no downside to explicitly stating it

## Other issues found

I found other minor issues in getting started that I'm unsure were worth fixing or addressing:

- if you use npm you must do `--force` or set legacy deps to true because of some mismatched legacy deps. Since project is in beta I think this is fine
- There is no tsconfig setup which will cause ts to be mad we are not importing react based on the default ts configuration. This seems very minor though
- Using pnpm I could not get the app to start. It's common for a dev to not listen to the docs and use whatever package manager they prefer. If this wasn't beta I would suggest explicitly calling out devs can use yarn
- If user is using pnpm, implicit dependencies such as vite must be installed too. The getting started guide shoud either include vite in the package.json or update one to add vite as a peer dependency.

### DIFF
--- a/code/ooo/data/docs/installation.mdx
+++ b/code/ooo/data/docs/installation.mdx
@@ -42,9 +42,6 @@ Let's begin with our package.json:
 }
 ```
 
-Note the `sideEffects` field - for production builds, this will enable much better tree shaking.
-
-
 It is strongly recommended to pin the `one` version for production stability:
 
 ```bash


natew:
You know I noticed the side effects actually did make a difference for some reason when building the server side and api stuff. Otherwise I’d end up with a lot of stuff in each endpoint.

It’s possible the monorepo I had caused this but I’ll set up a test case and see what I find.

natew:
One sets up tsconfig for you when you start the dev server as well with a decent default.

natew:
Thanks for these though. I’ll get a lot of these rolled in

natew:
Thank you for all this feedback, I believe I've addressed it all: 9621f83e5c3cf050d48def514fc39f2633f37e5a e6c94c495a6e0671fd58da654a1f07d35db64254 (and a couple on the one side).

PR #2967, opened 2024-10-04
Fix capitalization of GitHub

michaelmior:
There are a number of places where GitHub is incorrectly capitalized as Github. This corrects those errors.


natew:
thank you

PR #2984, opened 2024-10-12
fix(toast): ios should repropagate context when the native portal is not used

### LINKED ISSUE #2998: Cannot read property 'viewports' of undefined when use portal toast on ios
### Current Behavior

On iOS, when process.env.TAMAGUI_USE_NATIVE_PORTAL = 'false' is set, if the toast is configured with portalToRoot, it will fail to properly retrieve the context and throw an error.
image

### Expected Behavior

Render the toast correctly without throwing an error.

### Tamagui Version

```markdown
1.114.2
```

```markdown
iOS
```

### Reproduction

```markdown
process.env.TAMAGUI_USE_NATIVE_PORTAL = 'false' 
```
```
<ToastViewport portalToRoot />
```
```


### System Info

_No response_

LiZhequ:
iOS may not be using the native portal, for example, when `process.env.TAMAGUI_USE_NATIVE_PORTAL === 'false'`. Therefore, iOS also needs to repropagate  the context.
https://github.com/tamagui/tamagui/blob/a05d829ab766f7734bc97007a00b520ac29e092a/code/ui/portal/src/Portal.native.tsx#L31-L39

close #2998


natew:
@LiZhequ if you have an email wanted to see if you're interested in some part-time work

PR #2991, opened 2024-10-15
add docker test

natew:


jonsherrard:
Have the tests running in a docker instance using `yarn test:docker` and a new image in `test/Dockerfile.test`

They are failing at the minute but would be fixed by https://github.com/tamagui/tamagui/pull/2989

One key thing to note, I have remove .git from the .dockerignore, as we're using git to pull the tests into docker - to simulate Github Actions.

Is the Dockerfile in the root used for deploying `ooo` and `tamagui.dev` sites?

PR #2992, opened 2024-10-15
fix(static): allow to load .node files

franzmoro:
e.g. canvas.node files


jonsherrard:
Hi @franzmoro coudl you point me to an example file I could test this with?

PR #3004, opened 2024-10-19
Update theme-builder guide createTheme params

canadaduane:
Fix theme-builder guide by including missing createTheme hard-coded mapper function params. Also remove mentions of masks, a concept which was previously removed from the guide for the sake of simplicity.


natew:
Ah yea thanks, I'm trying to get a better theme builder setup for us eventually.

PR #3097, opened 2024-12-15
Add :focus-within support  

byteab:
- [x] web support
- [x] native support
- [x] CSS only support
- [x] fix `focusVisible` ( add missing doc, and fix CSS only mode )
- [x] `$group-` support
- [x] document and demo


natew:
@byteab how close is this one?

byteab:
> @byteab how close is this one?

just need to be added to doc

byteab:
@natew it should be ready now

code/core/web/src/createComponent.tsx
```diff
@@ -688,6 +689,11 @@ export function createComponent<
 
     const unPress = () => setStateShallow({ press: false, pressIn: false })
 
+    if (propsIn.focusWithinStyle) {
+      componentContext.onWithinFocus = () => setStateShallow({ focusWithin: true })
```
natew, line 693:
```suggestion
      componentContext.onWithinFocus ||= () => setStateShallow({ focusWithin: true })
```

natew: for both

code/core/web/src/helpers/pseudoDescriptors.ts
```diff
@@ -20,6 +20,11 @@ export const pseudoDescriptorsBase = {
     name: 'focus',
     priority: 3,
   },
+  focusWithinStyle: {
```
natew, line 23: what happens when you have both visible and within? they collide right?

byteab: yeah both have same `specificity` so should respect order

PR #3101, opened 2024-12-18
Fix `fontVariant` style

TimPetricola:
Currently, when passing `fontVariant` to a `Text` component, nothing happens (on web it adds `font-variant` attribute in the node).

It happens because this style property is not whitelisted as a valid style prop.

Adding it to the whitelist resolves the issue


rlods:
Amazing - Thanks!

PR #3109, opened 2025-01-01
feat(core): add rem support for native

byteab:
Add CSS `rem` to work on native too

### DIFF
--- a/code/core/web/src/types.tsx
+++ b/code/core/web/src/types.tsx
@@ -461,7 +461,7 @@ export interface SpacerStyleProps
   extends Omit<StackStyleBase, keyof SpacerUniqueProps>,
     SpacerUniqueProps {}
 
-export type SpacerProps = WithThemeShorthandsPseudosMedia<SpacerStyleProps>
+export type SpacerProps = WithThemeShorthandsPseudosMediaRem<SpacerStyleProps>
 
 type AllowedValueSettingBase =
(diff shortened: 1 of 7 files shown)


guillempuche:
This is fantastic news!
You could add a section in docs for using REM in different CSS properties and how it fits to the device's zoom/font scale.

natew:
@poteboy or @baronha - lets add some tests and docs and then this is good to merge.

code/core/web/src/types.tsx
```diff
@@ -1153,6 +1153,10 @@ export type GetTokenFontKeysFor<
     | 'color',
 > = keyof TamaguiConfig['fonts']['body'][A]
 
+export type WithRem<T extends object> = {
+  [K in keyof T]: Extract<T[K], number> extends never ? T[K] : `${number}rem` | T[K]
```
natew, line 1157: this strategy seems a bit wrong - we can probably just type the original ViewStyle to have `rem`. only reason is type performance is really sensitive and doing maps like this in this specific area is costly. i think at least.

guillempuche:
Do you have something in mind, @natew ?

IMO, the most important feature Tamagui is missing is rem/em/calc. I want to build fluid designs https://utopia.fyi/, very needed in this age with so many screen sizes, platforms...

The breakpoints are the past.

PR #3134, opened 2025-01-14
🐛 Fix: Adapt Context

baronha:
fix(adapt): safely call context methods to prevent potential undefined errors

Screenshot 2025-01-14 at 13 28 27


natew:
I fixed this presumably here https://github.com/tamagui/tamagui/commit/05b74b41a80fd302cb2ddc76391c7189840f6c59, but will merge since it could help in the future.

PR #3157, opened 2025-01-22
invalid input value check

alantoa:
CleanShot 2025-01-23 at 05 17 48@2x


alantoa:
CleanShot 2025-01-23 at 05 21 32@2x

another issue is happening here; actually, this error has already been returned, but I don't know why I didn't see any toast to show this error.

natew:
Yea need to make that flow now, I had an error from the api saying you must purchase, but haven't tested it yet.

PR #3184, opened 2025-01-31
Allow type prop on new Input.tsx

dohomi:
Allow the use of `time` or any other input related types. Currently there is a map and I am unsure why this is not allowing all input type variants.


dohomi:
yes agree I made the changes

natew:
@dohomi just curious are the types right here? can we improve them alongside this?

PR #3254, opened 2025-02-18
fix(core) merging transform of style prop, so they don't replaced by parent transforms.

### LINKED ISSUE #3223: Applying a transform like `scale` to SliderThumb breaks the positioning of the thumb
### Current Behavior

I want to add a pressStyle to the SliderThumb to make it larger when pressed

```JS
pressStyle: {
    scale: 1.2
}
```

but as soon as I implement the pressStyle the transforms (transformX & transformY) that are applied by (I guess) the Slider Wrapper are overitten. Without the transformX and transformY the Slider looks broken

Image

Is there any way to merge the transforms?

### Expected Behavior

Scaling (or applying any other transformation) the Thumb does not effect the actual position on the slider

### Tamagui Version

```markdown
^1.123.17
```

```markdown
Web, iOS, Android
```

### Reproduction

```markdown
Just use the Slider example from the docs https://tamagui.dev/ui/slider and apply a `pressStyle={{ scale: 1.5 }}` to the Slider.Thumb.


import type { SliderProps } from 'tamagui'
import { Slider, XStack } from 'tamagui'

export function SliderDemo() {
  return (
    <XStack height={200} alignItems="center" gap="$8">
      <SimpleSlider height={200} orientation="vertical" />
      <SimpleSlider width={200} />
    </XStack>
  )
}

function SimpleSlider({ children, ...props }: SliderProps) {
  return (
    <Slider defaultValue={[50]} max={100} step={1} {...props}>
      <Slider.Track>
        <Slider.TrackActive />
      </Slider.Track>
      <Slider.Thumb size="$2" index={0} circular pressStyle={{ scale: 1.5 }} />
      {children}
    </Slider>
  )
}
```

### System Info

```markdown

```

zarin-maruf:
Closes #3223

## current behavior

if you use styled(SliderThumb, {...}) and then pass Pseudo-transform styles those will override the component transform style. (IOS, Android, web)

## New behavior 

it will merge those transforms with component transform styles


natew:
Would be actually more helpful to set up a failing test case, because these changes are incredibly tricky and require a lot of care. With a proper failing test case I can also save a lot of time fixing it.

You can add it to the kitchen-sink, you'll see how it works there, use-cases.

natew:
If you want to verify with me first on what exactly you want to work too, it's possible that we don't want to merge things that way so we can save time and clarify.

zarin-maruf:
> Would be actually more helpful to set up a failing test case, because these changes are incredibly tricky and require a lot of care. With a proper failing test case I can also save a lot of time fixing it.
> 
> You can add it to the kitchen-sink, you'll see how it works there, use-cases.

Thanks for the feedback! That makes sense. I'll add a failing test case in the kitchen-sink to demonstrate the issue clearly. This should help ensure the fix is well-tested and reliable. I'll update the PR soon!

zarin-maruf:
> If you want to verify with me first on what exactly you want to work too, it's possible that we don't want to merge things that way so we can save time and clarify.

Got it! I appreciate the guidance. Before making further changes, I'd love to clarify the approach to ensure it's aligned with what you have in mind. Could you share any specific concerns or preferred direction for handling this?

natew:
It's more I don't understand fully what's breaking or not working as expected, a test case will help show that.

zarin-maruf:
@natew, I’ve added a test case but couldn’t fully complete the test because the function doesn’t return the actual translateX and translateY values for transform.

Here’s the current function:

```
export async function getStyles(locator: Locator) {
  return await locator.evaluate((el) => {
    return window.getComputedStyle(el)
  })
}
```

I tried the following alternative, which worked and allowed me to see the overridden transform values:

```
export async function getStyles(locator: Locator) {
  return await locator.evaluate((el) => {
    return el.style
  })
}
```

However, I’m unsure whether this should be added to the test utils. Let me know your thoughts on the best approach!

natew:
I think not doing Switch is important for the test case, it should be very minimal and only use plain `View`. That will help test it more easily and isoalte the logic. I think what you are saying is you want pressStyle scale to merge into the parent transforms properly, which it should do definitely, and I think does in simple case. So that will help me see maybe there's some specific case here we missed.

zarin-maruf:
The test case is now complete. I appreciate your valuable feedback on the PR and would love to hear your thoughts on the updated approach!

natew:
I meant that you shouldn't have `Switch` at all in the test - just plain View. Lets isolate the issue down to the simplest form.

zarin-maruf:
Apologies for the misunderstanding. Just to confirm, are you suggesting replacing `XStack` in the test case with a plain `View`?

natew:
I’m saying the test case should not import or use Slider at all. It should only be done with plain View. That way I can see exactly what’s going wrong without all the complexity of Slider which has a lot of stuff going on.

zarin-maruf:
@natew My mistake—I finally found the root cause. The issue occurs when wrapping a styled component with `React.memo `and then passing it to another styled() function, which prevents it from getting the parent’s static config. I’ve also updated the test case. Let me know your thoughts on the test and code changes.

natew:
Ah yes, you can't wrap with memo that will break the styleable() logic. That makes sense, and I see the error now in SliderThumb and have a fix going out.

That was super helpful to have you reduce it down! I'll close as the fix should be out here in a minute, if it doesn't work lmk we can re-open.

PR #3271, opened 2025-02-24
fix(slider): prevent slider-active-track to capture event instead of slider-track. so it gives different locationY

zarin-maruf:
## Current behavior

When you tap on the slider track in the vertical slider, it doesn't move properly. (Android, IOS)

https://github.com/user-attachments/assets/bfb5f178-1d0a-44f8-adbd-9cdd4e687eeb

## New Behavior

When you tap on the slider track, it moves correctly.  

https://github.com/user-attachments/assets/c2928cfa-9636-4caf-bc25-aef1d5cc91e5


zarin-maruf:
The root problem was that slider-track-active captures the events, so locationY gets calculated based on slider-tracK-active. which is wrong.

natew:
thank you

PR #3281, opened 2025-03-01
cleanup pro helpers and remove old code

natew:


natew:
@poteboy the code around this whole area is a mess, I dont think we even need the claim system etc, but also want to avoid a lot of work here if its working. but these changes seemed at least fine, perhaps we can clean up some more.

poteboy (approved):
Double-checked and confirmed that this change does not have any impact. I agree we have a lot to clean up.

PR #3282, opened 2025-03-04
AI Theme DB & OG Image

poteboy:


code/tamagui.dev/app/(site)/theme/[...subpath].tsx
```diff
@@ -26,36 +26,79 @@ import {
...
+        }
+        openGraph={{
+          images: [
+            {
+              url: `https://tamagui.dev/api/theme/open-graph?id=${data?.id || '0'}`,
+            },
+          ],
+        }}
+      />
+      {typeof window !== 'undefined' && <ThemePage data={data} />}
```
Comment (poteboy) [inline code/tamagui.dev/app/(site)/theme/[...subpath].tsx:78]: We could load it dynamically to reduce the bundle size, but the initial load is faster with synchronous loading if we load the entire component after rendering.

PR #3283, opened 2025-03-04
Fix one-time payment subscription handling and add migration script

poteboy:
## Issue
When users purchase a Pro plan with disableAutoRenew=true, the system creates a one-time payment invoice instead of a subscription. This causes two problems:
1. The webhook handler doesn't process invoice.paid events (only handles checkout.session.completed)
2. As a result, one-time payment users don't get their subscription records in the database

## Changes
1. Add webhook handler for one-time payments
     - Added invoice.paid event handler in webhook+api.ts
     - The handler creates subscription records for one-time payment invoices
     - Sets subscription status to active with a one-year validity period
2. Add migration script
     - Created scripts/migrate-one-time-payments.mjs to:
     - Find all paid invoices for Pro plan one-time payments
     - Create missing subscription records for existing customers
     - Add corresponding subscription items

### DIFF
--- a/code/tamagui.dev/app/api/stripe/webhook+api.ts
+++ b/code/tamagui.dev/app/api/stripe/webhook+api.ts
@@ -15,6 +15,7 @@ import {
 } from '~/features/auth/supabaseAdmin'
 import { sendProductRenewalEmail } from '~/features/email/helpers'
 import { stripe } from '~/features/stripe/stripe'
+import { supabaseAdmin } from '~/features/auth/supabaseAdmin'
 
 const endpointSecret = process.env.STRIPE_SIGNING_SIGNATURE_SECRET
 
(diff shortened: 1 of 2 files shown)


code/tamagui.dev/app/api/stripe/webhook+api.ts
```diff
@@ -124,9 +134,11 @@ export default apiRoute(async (req) => {
       }
 
       case 'checkout.session.completed': {
-        await addRenewalSubscription(event.data.object as Stripe.Checkout.Session, {
-          toltReferral,
-        })
+        const options = toltReferral ? { toltReferral } : undefined
+        await addRenewalSubscription(
+          event.data.object as Stripe.Checkout.Session,
```
poteboy, line 139: fixed a type error

PR #3298, opened 2025-03-08
fix(core): deleted color from tokenCats array because it is not existed in config v3 and v4.

### LINKED ISSUE #3177: ToggleGroup.Item throws error
### Current Behavior

After update to latest Tamagui version `ToggleGroup.Item` stopped working.

Minimal reproduction:
```jsx
        <ToggleGroup type="single">
          <ToggleGroup.Item value="a">
            <ChevronDown />
          </ToggleGroup.Item>
        </ToggleGroup>
```

```
Unhandled Runtime Error
TypeError: Cannot use 'in' operator to search for '' in undefined

Call Stack
getVariantDefinition
```

### Expected Behavior

Should render

### Tamagui Version

```markdown
1.123.3
```

```markdown
web
```

### Reproduction

https://codesandbox.io/p/devbox/sad-cookies-8cl78m

zarin-maruf:
Closes #3177

## current behavior 

if you add the following context and variant to any component.
```

export const context = createStyledContext({
  color: '',
})


const StyledView = styled(View, {
  context,

  variants: {
       color: {
        '...color': () => {
          return {}
        },
      },
    }
})

```

it will give you this error 

TypeError: Cannot use 'in' operator to search for '' in undefined. 

This issue happens because config V3 and V4 tokens don't include color, and this snippet code tries to check the color

```
 const { tokensParsed } = conf
    for (const { name, spreadName } of tokenCats) {
      if (spreadName in variant && value in tokensParsed[name]) {
        return variant[spreadName]
      }
    }
```
so `tokensParsed[name]` is undefined  and  `value in tokensParsed[name]` crash the app.

## New behavior 

I removed the color from the tokenCats array, which will no longer check for the color, and it works nicely.

@natew Let me know if there's anything you'd like me to adjust!

### DIFF
--- a/code/core/web/src/helpers/propMapper.ts
+++ b/code/core/web/src/helpers/propMapper.ts
@@ -330,7 +330,7 @@ function getVariantDefinition(variant: any, value: any, conf: TamaguiInternalCon
   if (value != null) {
     const { tokensParsed } = conf
     for (const { name, spreadName } of tokenCats) {
-      if (spreadName in variant && value in tokensParsed[name]) {
+      if (spreadName in variant && name in tokensParsed && value in tokensParsed[name]) {
         return variant[spreadName]
       }
     }


natew:
We still need to check color as its a supported category, what we can do is add a check if it exists in the tokens though first.

zarin-maruf:
> We still need to check color as its a supported category, what we can do is add a check if it exists in the tokens though first.

Thanks, I will address it ASAP

zarin-maruf:
@natew  I’ve addressed your feedback. Let me know if you have any other suggestions. Thanks!

code/core/web/src/helpers/propMapper.ts
```diff
@@ -327,10 +327,14 @@ function getVariantDefinition(variant: any, value: any, conf: TamaguiInternalCon
...
     return exact
   }
-  if (value != null) {
+  if (value != null && (typeof value !== 'string' || value.trim() !== '')) {
     const { tokensParsed } = conf
     for (const { name, spreadName } of tokenCats) {
-      if (spreadName in variant && value in tokensParsed[name]) {
+      if (
+        spreadName in variant &&
+        typeof tokensParsed[name] === 'object' &&
```
natew, line 335: no you want `spreadName in variant && name in tokensParsed && ` i think

zarin-maruf:
You're right. I've fixed it now. Let me know if anything else needs adjustment. Thanks!

code/core/web/src/helpers/propMapper.ts
```diff
@@ -327,10 +327,10 @@ function getVariantDefinition(variant: any, value: any, conf: TamaguiInternalCon
   if (exact) {
     return exact
   }
-  if (value != null) {
+  if (value != null && (typeof value !== 'string' || value.trim() !== '')) {
```
natew, line 330: dont need this

zarin-maruf: Value could be an empty string. and that prevents the rest of the code from running. i removed it btw

PR #3320, opened 2025-03-14
Update react-native deep imports for 0.79 compatibility

huntie:
## Summary

In the incoming React Native 0.79 release, we've converted a number of internal modules to `export` syntax — affecting usages with `require()`. This PR attempts to update these imports ahead of time, to work with both versions.

💡 Secondary note: In a future release, we are intending to omit each of these deep-imported modules from the public API of React Native. There will be a consultation period on opening up new APIs via the root `'react-native'` exports. Some of the uses in this library may need `@ts-ignores` in future.

## Test plan

**Needs testing**. CI should pass for integration with current RN versions.

### DIFF
--- a/code/core/core/src/getBaseViews.native.ts
+++ b/code/core/core/src/getBaseViews.native.ts
@@ -6,7 +6,8 @@ export function getBaseViews() {
 
   if (process.env.NODE_ENV !== 'test') {
     View = require('react-native/Libraries/Components/View/ViewNativeComponent').default
-    TextAncestor = require('react-native/Libraries/Text/TextAncestor')
+    const TextAncestorModule = require('react-native/Libraries/Text/TextAncestor')
+    TextAncestor = TextAncestorModule.default ?? TextAncestor
   }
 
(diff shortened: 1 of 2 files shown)


code/core/core/src/getBaseViews.native.ts
```diff
@@ -6,7 +6,8 @@ export function getBaseViews() {
 
   if (process.env.NODE_ENV !== 'test') {
     View = require('react-native/Libraries/Components/View/ViewNativeComponent').default
-    TextAncestor = require('react-native/Libraries/Text/TextAncestor')
+    const TextAncestorModule = require('react-native/Libraries/Text/TextAncestor')
+    TextAncestor = TextAncestorModule.default ?? TextAncestor
```
huntie, line 10: This can either be merged as so (cross compatiblity), or simply updated to `require(...).default` (**IF** versioned for >=0.79 projects only).

natew:
Thanks @huntie, looks pretty benign so I'll merge.

PR #3324, opened 2025-03-16
fix: resisted and scrollview issue when dragging the sheet

nnphong1904:


code/ui/sheet/src/SheetImplementationCustom.tsx
```diff
@@ -227,6 +229,7 @@ export const SheetImplementationCustom = React.forwardRef<View, SheetProps>(
     const disableDrag = props.disableDrag ?? controller?.disableDrag
     const themeName = useThemeName()
     const [isDragging, setIsDragging] = React.useState(false)
+    const [scrollEnabled, setScrollEnabled] = React.useState(true)
```
natew, line 232: can this be a ref since its only read during callback? can avoid a setstate re-render?

code/ui/sheet/src/SheetImplementationCustom.tsx
```diff
@@ -308,18 +316,19 @@ export const SheetImplementationCustom = React.forwardRef<View, SheetProps>(
         }
         // prevent drag once at top and pulling up
         if (isNearTop) {
-          if (!isScrolled && isDraggingUp) {
+          if (scrollEnabled && isDraggingUp) {
             // TODO: pulling past the limit breaks scroll on native, need to better make ScrollView
-            if (!isWeb) {
-              return false
-            }
+            // if (!isWeb) {
```
natew, line 321: did you test this on native? can you show that working with scrollview inside? and if it works we can remove the commented out parts here and TODO comment

code/ui/sheet/src/SheetScrollView.tsx
```diff
@@ -50,6 +56,10 @@ export const SheetScrollView = React.forwardRef<
       isDragging: false,
     })
 
+    useEffect(() => {
+      setHasScrollView(true)
```
natew, line 60: needs to set back to falls un unmount

natew:
Overall looks good just left some comments

code/ui/sheet/src/SheetImplementationCustom.tsx
nnphong1904: https://github.com/user-attachments/assets/6c9ed8ad-433e-4b52-8a40-262a998a7840

@natew you can check my screen recording to see how it works in native. Everything is working normally like it was

natew: the sheet should drag down once its at the top even if you drag from the content of the sheet, it works like that today i believe (if not thats a bug). it's definitely one of the hard parts of this work, getting it to hand off from scrolling to dragging, but it should be possible (gorhom bottom sheet does this), and i think we support it today its just not as smooth as possible.

nnphong1904: To confirm, I tested with the current version before my fix, which has the same issue. So, when users start scrolling, there will be some bounce effect when they drag the sheet down. 

I will deep dive on this more to improve it.

https://github.com/user-attachments/assets/aa61c669-c5be-4c72-b39c-5043515c74e4

natew: i agree that is another thing we want to fix, but on native - today you can drag from anywhere in the sheet to pull it down once its at the top, it looks like in your video you are dragging only from the handle at the top. can you still drag from anywhere in the sheet to pull down?

natew:
i pushed some fixes here from type issues and missing props now on native, https://github.com/tamagui/tamagui/pull/3334

i'm going to merge that work, but we should follow on and make sure that logic is right on native before i can do a release.

PR #3330, opened 2025-03-18
Update Adapt Component

kendalleasterly:
- Using when="sm" no longer works. I just used when={0} and that works for mobile.


natew:
Hm when="sm" should work if its not that's bad, where does it break?

PR #3356, opened 2025-03-28
chore(Slider): use new styledContext and make it easier to customize

byteab:
### Patch notes
- Use new createStyledContext instead of old createScopeContext
- Make the slider easier to customize, especially the slider Thumb sizes


natew:
lgtm pending checking it works well, we should make it chore() as its not user facing feature.

byteab:
Thank you @natew 
fixed that commit message

Issue #3183, opened 2025-01-30
ToggleGroup bugs on demo website


ofersadgat:
### Current Behavior

1. "multiple" doesnt allow you to select multiple elements
2. clicking off of an element make it look like the option has been deselected

You can see the bugs on the documentation website: https://tamagui.dev/ui/toggle-group

### Expected Behavior

1. "multiple" allows you to select multiple options which all appear darkened
2. clicking off of the element doesnt change the visual state of the element

### Tamagui Version

```markdown
Latest
```

### Platform (Web, iOS, Android)

```markdown
Web
```

### Reproduction

```markdown
https://tamagui.dev/ui/toggle-group
```

### System Info

```markdown

```

ofersadgat:
@zetavg @natew its been two months, any comment about this?

PR #3362, opened 2025-03-31
🐛 fix: Enhance Toggle component styles and theme handling.

baronha:
Will close https://github.com/tamagui/tamagui/issues/3183
It will look like this:

Screenshot 2025-03-31 at 13 47 41


natew:
I don't think we want to inverse it by default actually tbh, that's confusing for users. We need an easy and optimizable way for the users to set active styles themselves, which kind of goes into deep Tamagui theory.

But for now we could just have the examples use state, or something like that. And not even do anything internally for active styles.

ofersadgat:
> I don't think we want to inverse it by default actually tbh, that's confusing for users. We need an easy and optimizable way for the users to set active styles themselves, which kind of goes into deep Tamagui theory.
> 
> But for now we could just have the examples use state, or something like that. And not even do anything internally for active styles.

Sorry, I dont quite understand. Why is it more confusing for users for it to work out of the box with no configuration than to require configuration to work in an expected way?

natew:
A few reasons: we'd need to document the inverse, but we don't inverse anywhere else by default and it's easy to miss that if you are just using it in practice. If they use custom themes that don't have an inverse, that would break / be confusing.

What we could do is use inverse tokens by default, but changing themes is definitely more confusing as you would basically need to read code / inspect things to understand that from the outside

natew:
Also inversing is subjective, arguably we should add a "backgroundActive" generic.

ofersadgat:
> A few reasons: we'd need to document the inverse, but we don't inverse anywhere else by default and it's easy to miss that if you are just using it in practice. If they use custom themes that don't have an inverse, that would break / be confusing.
> 
> What we could do is use inverse tokens by default, but changing themes is definitely more confusing as you would basically need to read code / inspect things to understand that from the outside

Specifics aside, I think that if you use the default config, you should have something shippable. Maybe not the most beautiful thing, but it should be shippable. A toggle which doesnt show that its depressed isnt shippable. I think that this is one of the strengths of UI toolkits: the speed with which you can get to something shippable. 

After you get to the point of it being shippable, it makes sense to make it pretty. That being said, there arent a lot of good examples for how to do this (or what the tamagui recommendations are). From chatting with other people, they've had similar experiences and their main recommendation was to look at the takeout/bento code for examples, but IMO this should really be part of documentation. To provide a concrete example:

```
function BreadCrumbUI({parts}: {parts:string[],onClick:(part:string,index:number)=>void}){
    return (
         <XStack gap={4} backgroundColor="grey" borderColor="lightGrey">
               {parts.map((part, i) => {
                    return (<>
                            <Button onPress={() => onClick(part, i)} key={i*2}>
                                {part}
                            </Button>
                            <ChevronRight key={i*2+1} />
                    </>);
               }}
         </XStack>
    )
}
```

This probably looks terrible to you, but its how I would normally use a toolkit like this...but now that I've written it...it looks terrible and idk how to fix it. I've learned that hard coding colors isnt what I should do, but then how do I choose what "backgroundColor" should be? What if I want "backgroundColor" to be different in different components? I also dont know if XStack is the correct choice vs ListItem. Also, these "color + number" combinations appear opaque to me...as well as sizing. What is "4" space? should I be using 4 for gutter spaces? space between items? padding inside a button? These questions are probably much more beginner oriented, but I think that would really help pick this up better. Thanks for all your hard work!

ofersadgat:
@natew @baronha why close this? the issue isnt addressed

natew:
I think we can have the docs at least have good default styles for the copy paste styles. And we can use our generic tokens from themes to have nice defaults as well. We have backgroundFocus already which we can use for this.

timheilman:
I respect that the proper solution to #3183 is perhaps more involved than what this PR provides.  All the same, [the demo webpage for this component](https://tamagui.dev/ui/toggle-group) claims that `type='multiple'` allows multiple items to be selected, which it doesn't.

Would it be better not to document or demo `type='multiple'` at all, until it does what it claims it does? I feel I would be better off if I went looking for a multiple-toggle-group and was unable to find it than the current situation: I can find it, and it doesn't work.

natew:
Multiple was working, I'm checking now for a regression and will remove docs for now. Sorry for that and thanks for the heads up. Will see about a fix as well. We have the improvements in v2 already for this as well.

PR #3392, opened 2025-04-18
Add AlertDialog Destructive button type

Pietrohl:
Adding support for native ios alert destructive button type


natew:
Thanks for this, I think we can roll this into our v2 PR once we get back to it, we're taking a couple weeks to finish cleaning up v1 bugs and then will be resuming that.

natew:
Merged to v2 in commit 1b4a52064c. Had to apply manually due to v2 refactors (createStyledContext pattern). Thanks @Pietrohl!

PR #3407, opened 2025-04-26
fix(dialog): dialog not working without adapt

nnphong1904:


code/ui/dialog/src/Dialog.tsx
```diff
@@ -307,6 +304,10 @@ const DialogOverlay = DialogOverlayFrame.extractable(
     return (
       <DialogOverlayFrame
         data-state={getState(context.open)}
+        onPress={() => {
+          // if the overlay is pressed, close the dialog
```
natew, line 308: i don't think we want to change functionality here if this wasn't the case before. i do agree it's a good change but we likely want to do this in v2

code/ui/dialog/src/Dialog.tsx
```diff
@@ -861,6 +862,8 @@ const Dialog = withStaticProperties(
           forwardProps: props,
         }}
       >
+        <PortalHost name={adaptName} />
```
natew, line 865: adding the host here is strange just on first read, don't we want the host generally to be up near the root of the app so it portals up to the root?

yarn.lock
natew, line 1: lets revert these changes no need to change deps here

code/ui/portal/src/GorhomPortalItem.tsx
```diff
@@ -19,8 +19,11 @@ export const GorhomPortalItem = (props: PortalItemProps) => {
     if (!props.hostName) return
     if (node) return
 
-    const listener = (node) => {
-      setNode(node)
+    const listener = (newNode: HTMLElement) => {
+      if (!node || newNode.isEqualNode(node)) {
```
natew, line 23: if its === then it shouldn't re-render? calling react setState with something === will skip re-rendering, i've actually never seen isEqualNode though is it diff from ===?

natew: also couldn't this sometimes be called on native? not sure haven't read this closely

nnphong1904: > also couldn't this sometimes be called on native? not sure haven't read this closely

no, this is only for the web, because we have a separate file `GorhomPortalItem.native.tsx` for the native side

code/ui/dialog/src/Dialog.tsx
nnphong1904: I think I should add a comment to explain the purpose here. However, this will mainly be used in the case where `modal=false`, which will render an inline dialog.

code/ui/portal/src/GorhomPortalItem.tsx
nnphong1904: > if its === then it shouldn't re-render? calling react setState with something === will skip re-rendering, i've actually never seen isEqualNode though is it diff from ===?

the `isEqualNode` is different from `===` is will make a deep compare for the attributes like id, children,... `===` just compare if 2 node are the same instance or not.

nnphong1904: you can check this doc: https://developer.mozilla.org/en-US/docs/Web/API/Node/isEqualNode

nnphong1904: And yeah, it shouldn't re-render when 2 node are identical.

nnphong1904: > also couldn't this sometimes be called on native? not sure haven't read this closely

no, this just for web only because we have a separate `.native` file

nnphong1904: > if its === then it shouldn't re-render? calling react setState with something === will skip re-rendering, i've actually never seen isEqualNode though is it diff from ===?

You're right. The `isEqualNode` condition was unnecessary. But you can refer the docs about `isEqualNode` here: https://developer.mozilla.org/en-US/docs/Web/API/Node/isEqualNode

PR #3424, opened 2025-05-03
starters: upgrade expo-53 sdk

baronha:


natew:
@baronha just note a couple conflicts in this PR

PR #3429, opened 2025-05-06
fix: mising exports field in package.json

nnphong1904:


code/core/web/package.json
```diff
@@ -55,7 +55,7 @@
     "./package.json": "./package.json",
     "./reset.css": "./reset.css",
     ".": {
-      "react-native": {
+      "react-native": {
```
natew, line 58: remove whitespace here

PR #3440, opened 2025-05-08
fix(lucide-icons): fix invalid exports

nnphong1904:


natew:
landed a more real fix see my commit changing TAMAGUI_BUILD_CLEANUP_JS_FILES on main

PR #3443, opened 2025-05-09
Fix expo-router starter

vovsemenv:


natew:
ah missed this but we merged sometihng similar, thanks

PR #3450, opened 2025-05-13
fix(animations): allow enterStyle animations to work on disabled component

anhquan291:
- Remove early return on disabled state to allow animations to work
- Allow state updates for animations when component is disabled
- Skip context group subscription for disabled components to prevent interactive states
- Maintain existing behavior for non-disabled components

https://github.com/user-attachments/assets/54d243ce-9468-4ce7-bd53-5023af464b55

### DIFF
--- a/code/core/web/src/createComponent.tsx
+++ b/code/core/web/src/createComponent.tsx
@@ -719,10 +719,8 @@ export function createComponent<
     }
 
     useIsomorphicLayoutEffect(() => {
-      if (disabled) {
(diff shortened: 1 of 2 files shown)


anhquan291:
iOS

https://github.com/user-attachments/assets/8827474f-3946-493d-8e48-0ebd2818d998

Android



https://github.com/user-attachments/assets/a459e19c-31a7-4e98-a23b-667deb325f3a

PR #3470, opened 2025-05-19
refactor: subscription

anhquan291:


code/tamagui.dev/app/api/discord/channel+api.ts
```diff
@@ -42,13 +42,26 @@ export default apiRoute(async (req) => {
...
+  // Get team subscription to get the total_seats for Tamagui Pro Team Seats plan
+  const teamSubscription = await supabaseAdmin
+    .from('team_subscriptions')
+    .select('*')
+    .eq('owner_id', user.id)
+    .single()
+
+  if (teamSubscription.error) {
+    throw teamSubscription.error
+  }
```
zetavg, line 63: Hmm, I don't think this will work as expected, since:

* Not all Tamagui Pro users will have a team subscription. For those who didn't purchase additional seats, there will be no `team_subscriptions`, so this will be broken for them.

Additionally, this implementation only lets team owners manage Discord seats for the whole team. Team members will not be able to add themselves since this query cannot find the subscription where the user is a team member.

I tried to make some modifications that addressed the above two points, you can take it if you wish: https://github.com/tamagui/tamagui/pull/3471/commits/a54510d3fab1179ba823ef217f24e475726413d9

zetavg (changes requested):

code/tamagui.dev/app/api/discord/channel+api.ts
zetavg: I think that API is also used by Pro users (without additional team seats purchased), who should also be able to access the general channel. I can repro the issue with a newly created Pro subscription:



Sending details via pm.

PR #3500, opened 2025-06-02
fix(popover): cannot override popover index

nnphong1904:


code/ui/z-index-stack/src/useStackedZIndex.tsx
```diff
@@ -17,7 +17,8 @@ export const useStackedZIndex = (props: {
     const id = useId()
 
     const zIndex = useMemo(() => {
-      if (stackZIndex && stackZIndex !== 'global') {
+      // should disable stackZIndex when zIndex prop is set
+      if (stackZIndex && stackZIndex !== 'global' && zIndexProp !== undefined) {
```
natew, line 21: maybe its just early but isnt this backwards - we should only use stackZIndex if zIndexProp is undefined?

PR #3505, opened 2025-06-06
Add missing 'r' in "Compiler"

esthor:
Super tiny typo. (Mainly just wanted an excuse to go spelunking through the codebase 🤓)

This is buried in the hamburger menu.

I considered pulling this out into a variable or switching on the types, but felt that was potentially overkill and a wider blast radius for something so small.
image


natew:
thanks!

PR #3548, opened 2025-07-13
fix: render last adapt context provider instead of the object itself

alexanderjhughes:
This is just a quick fix to be able to render the sheets. This context may not even be needed


natew:
yep, tested it and it fixes. bad type there allowing it :/

PR #3560, opened 2025-07-28
<Select /> types generic value

greg-schrammel:
I mean you can lie to these types because it's not fully integrated with `<Select.Item />` but Idk if we can do that without some sort of `createTypedSelect<'a' | 'b'>()`
more of I helpful thing, I understand why you'd want to keep it out the lib, also easy to do on user land
happy to do it if you have an idea to make it fully type safe

### DIFF
--- a/code/ui/select/src/Select.tsx
+++ b/code/ui/select/src/Select.tsx
@@ -336,7 +336,9 @@ const SelectSheetImpl = (props: SelectImplProps) => {
  * -----------------------------------------------------------------------------------------------*/
 
 export const Select = withStaticProperties(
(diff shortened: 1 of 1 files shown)


natew:
this is probably a fine half measure, i agree that `createSelect` should be something we provide during v2. Does that Select generic not need a `= string` though so users don't have to pass a type or is JSX typing different somehow?

greg-schrammel:
ts didn't scream at me so I'll say it's fine, but added it anyway

natew:
Merged via squash commit with feat(select) prefix. Thanks @greg-schrammel!

PR #3567, opened 2025-07-31
chore: remove typescript/native-preview

christophehurpeau:
`@typescript/native-preview` added in this commit https://github.com/tamagui/tamagui/commit/b88c3932a46fba1de5d8225f4539fb56735676d8#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519 should probably not be in dependencies. I'm not even sure it is used ?


natew:
good point, we can remove it entirely

christophehurpeau:
> good point, we can remove it entirely

PR updated !

Issue #3268, opened 2025-02-21
Button.Icon does not apply variant styles from ButtonContext


sabaturgay:
### Current Behavior

The `Button.Icon` component does not correctly apply the variant styles from `ButtonContext`. However, `Button.Text` applies variants properly.

### Expected Behavior

`Button.Icon` should apply the variant styles like `Button.Text`.
When using `<Button variant="secondary">`, the `Icon` should inherit the correct styles.

### Tamagui Version

```markdown
1.125.2
```

### Platform (Web, iOS, Android)

```markdown
iOS, Android
```

### Reproduction

```markdown
export const ButtonContext = createStyledContext<{
  variant: string
}>({
  variant: 'primary',
})

const ButtonText = styled(Button_.Text, {
  name: 'Button.Text',
  context: ButtonContext,
  fos: 18,
  fow: '600',
  variants: {
    variant: {
      secondary: {
        color: '$secondary',
      },
      outlined: {
        color: '$color',
      },
    },
  },
})

const ButtonIcon = styled(
  Button_.Icon,
  {
    name: 'Button.Icon',
    context: ButtonContext,
    scaleIcon: 1.7, 
    variants: {
      variant: {
        secondary: {
          color: '$secondary',
        },
      },
    },
  },
)

export const Button = withStaticProperties(
  styled(ButtonStyleable, {
    name: 'Button',
    context: ButtonContext,
    h: '$4.5',
    br: '$6',
    py: '$2',
    animation: '100ms',
    elevation: '$size.15',
    pressStyle: {
      elevation: '$size.7',
      scale: 0.95,
    },
    variants: {
      variant: {
        primary: {
          backgroundColor: '$primary',
          color: '$color',
        },
        secondary: {
          bg: '$colorTransparent',
          color: '$secondary',
          bw: '$0.5',
          bc: '$secondary',
        },
        outlined: {
          bg: '$colorTransparent',
          bw: '$0.5',
          bc: '$color',
        },
      },
    },
    defaultVariants: {
      variant: 'primary',
    },
  } as const),
  {
    Text: ButtonText,
    Icon: ButtonIcon,
  },
)
```

### System Info

```markdown

```

natew:
Hey, we've had no other reports of this and are doing some fall cleaning. If it's still an issue let me know\!

sabaturgay:
Hey @natew , I believe this issue is still present. I dug into the source and found the root cause.

[...]

Issue #2935, opened 2024-09-21
`createCheckbox` throws `React does not recognize the scaleIcon prop on a DOM element`


guillempuche:
### Current Behavior

The horizontal checkboxes in https://tamagui.dev/bento/forms/checkboxes works, but it produces a React error. It's because `createCheckboxe` from `@tamagui/checkbox`.

The error is produced in these lines https://github.com/guillempuche/test-tamagui-checkboxes/blob/cc1af9cae6a428e06a5bc5374239385a9e0564ed/packages/app/features/home/horizontal_checkboxes.tsx#L94-L100

### Expected Behavior

Checkboxes is clean of warnings

### Tamagui Version

```markdown
1.111.11
```

### Platform (Web, iOS, Android)

```markdown
Web
```

### Reproduction

```markdown
1. Clone this repo https://github.com/guillempuche/test-tamagui-checkboxes
2. `yarn install`
3. `yarn web`
4. Open http://localhost:3000/
5. Open browser's console and get the error
```

### System Info

```markdown
System:
    OS: macOS 14.7
    CPU: (8) arm64 Apple M1
    Memory: 76.52 MB / 16.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 20.14.0 - ~/.nvm/versions/node/v20.14.0/bin/node
    Yarn: 4.5.0 - ~/.nvm/versions/node/v20.14.0/bin/yarn
    npm: 10.7.0 - ~/.nvm/versions/node/v20.14.0/bin/npm
    bun: 1.1.17 - ~/.nvm/versions/node/v20.14.0/bin/bun
    Watchman: 2024.08.12.00 - /opt/homebrew/bin/watchman
  Browsers:
    Brave Browser: 128.1.69.162
    Chrome: 128.0.6613.139
    Safari: 17.6
  npmPackages:
    @babel/runtime: ^7.24.6 => 7.24.6 
    @manypkg/cli: ^0.21.4 => 0.21.4 
    check-dependency-version-consistency: ^4.1.0 => 4.1.0 
    eslint: ^9.3.0 => 9.4.0 
    node-gyp: ^10.1.0 => 10.1.0 
    prettier: ^3.3.3 => 3.3.3 
    turbo: ^1.13.4 => 1.13.4 
    typescript: ^5.6.2 => 5.6.2 
    vitest: ^2.1.1 => 2.1.1
```

sensor:
[...]

PR #3570, opened 2025-08-04
fix(helpers): resolve scaleIcon compatibility with external icon libraries

### LINKED ISSUE #2961: React does not recognize the `scaleIcon` prop on DOM element.
### Current Behavior

After upgrading from tamagui v1.104.2 to the latest - v1.112.14 using `yarn upgrade:tamagui`, a lot of my components, mainly bento components send this warning in the console.

This happens when using icons on input, or using the bento table example with cellWidth, etc.

```
Warning: React does not recognize the `scaleIcon` prop on a DOM element. If you intentionally want it to appear in the DOM as a custom attribute, spell it as lowercase `scaleicon` instead. If you accidentally passed it from a parent component, remove it from the DOM element.
```

### Expected Behavior

No warnings in the console

### Tamagui Version

```markdown
v1.112.14
```

```markdown
Web
```

### Reproduction

```markdown
Start a new tamagui project with npm create tamagui@latest

Add a bento input such as 


   <InputT size="$3">
        <InputT.Box h={40}>
          <InputT.Icon>
            <Search />
          </InputT.Icon>

          <InputT.Area
            paddingLeft={0}
            placeholder="Search..."
          />
        </InputT.Box>
      </InputT>
```

View your console
```


### System Info

```markdown
System:
    OS: macOS 14.6.1
    CPU: (8) arm64 Apple M1 Pro
    Memory: 124.36 MB / 16.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 20.16.0 - ~/.nvm/versions/node/v20.16.0/bin/node
    Yarn: 4.4.1 - /usr/local/bin/yarn
    npm: 10.8.1 - ~/.nvm/versions/node/v20.16.0/bin/npm
    bun: 1.1.8 - /opt/homebrew/bin/bun
    Watchman: 2024.05.06.00 - /opt/homebrew/bin/watchman
  Browsers:
    Chrome: 129.0.6668.71
    Safari: 17.6
```

lcf2212dev:
## What

Fixes `scaleIcon` property not working with external SVG icon libraries (like Iconoir) in Tamagui buttons.

## Problem

Currently, `scaleIcon` only works with Tamagui's native Lucide icons. External SVG icons are ignored because `useGetThemedIcon` doesn't detect their type correctly and applies the wrong props.

## Solution

Enhanced `useGetThemedIcon` to:
- Detect ForwardRefExoticComponent icons (Iconoir, etc.)
- Apply `width/height` props for external icons instead of `size`
- Differentiate Tamagui themed icons from external libraries
- Maintain full backward compatibility

## Testing

- ✅ Tamagui icons continue working exactly as before
- ✅ External icons now respect `scaleIcon` property  
- ✅ All icon formats supported: ForwardRef, Functions, React Elements

## Fixes

Closes #2961, #2935, #3268


natew:
this is good directionally but we should do it a bit differently, we should make themed attach 'acceptsSize' = true  to the component itself and then just use that as the heuristic, then otherwise just have a single fallback to width/height

lcf2212dev:
@natew Done exactly as you suggested.

Made themed attach acceptsSize = true to components, then switched the heuristic to just check el.acceptsSize === true.
Now icons get size prop, everything else gets width/height fallback.

(Same commit - did a squash and force push)

natew:
Thanks - feels messy though overall, a few thoughts:

- Can we use some React utility like isValidComponent?
- Passing color based on ref forwarded is not a good pattern
- I think what we could do is allow users to customize this externally and otherwise just pass width/height/color or size/color based on acceptsSize
- could do that with export `function configureIconProps(iconElement, propsIn) { return propsOut }` something like that and then re-export that from core and tamagui so people can do their own?
- I would refactor this entire thing to just have a single area at top that calculates the props, then a single piece of logic below that decides on clone vs create and passes the props

natew:
Another thought is we may not even need size, since i guess its always a number here, it may just work with width/height as well and no need for acceptsSize?

PR #3639, opened 2025-10-21
chore(package): move 'burnt' from devDependencies to dependency

anhquan291:


natew:
can't do this - last i checked metro will throw an error unless you install this, and then you get weird RN issues. we just recommend people to install if it they use optional native adapt stuff.

PR #3650, opened 2025-10-28
enhance(theme): add parents theme class for component themes

anhquan291:
## Problem
`<Button theme="green" bg="$color8">` renders grey instead of green, while `<Button theme="green" bg="$green8">` works correctly.

## Root Causes
When Button has `theme="green"`, Tamagui creates a combined theme `green_Button` and renders:
```html
<span class="t_green_Button">
  <button>
```

## Solution
Detects component themes (themes ending with uppercase component names like Button, Checkbox)
Adds both the parent theme class AND the component theme class.
```
Example: green_Button → classes become t_green t_green_Button
```

### DIFF
--- a/code/core/web/src/hooks/useThemeState.ts
+++ b/code/core/web/src/hooks/useThemeState.ts
@@ -379,21 +379,77 @@ function getNewThemeName(
 
   let found: string | null = null
 
-  const max = parentParts.length
-
-  for (let i = 0; i <= max; i++) {
-    const base = (i === 0 ? parentParts : parentParts.slice(0, -i)).join('_')
-
+  // If name is provided, try it as a standalone theme first (both with and without scheme)
+  // This allows explicit theme overrides like:
+  // - <Theme name="blue"><Button theme="dark_green"> → finds "dark_green_Button"
+  // - <Theme name="blue"><Button theme="green"> → finds "light_green_Button"
+  // - <Theme name="blue"><Button theme="green_active"> → finds "light_green_active_Button"
+  if (name) {
+    // First try the exact name as-is (for themes with explicit scheme like "dark_green")
     for (const subName of subNames) {
-      const potential = base ? `${base}_${subName}` : subName
-
-      if (potential in themes) {
-        found = potential
+      if (subName in themes) {
+        found = subName
         break
       }
     }
 
-    if (found) break
+    // If not found and name doesn't have a scheme, try adding parent's scheme
+    if (!found && !getScheme(name)) {
+      const parentScheme = getScheme(parentName)
+
+      if (parentScheme) {
+        // Get the parent theme name without component part (already done in parentParts)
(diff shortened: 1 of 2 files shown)


krakowski:
Hi @anhquan291, thanks for looking into this! :slightly_smiling_face:  I had a similar workaround until now.

```tsx
const componentThemePart = '_Button';

const StyleableButtonFrame = ButtonFrame.styleable<StyleableButtonExtraProps>(
	({ children, loading, ...props }, ref) => {
		const themeName = useThemeName();

		return (
			<Theme name={themeName.replace(componentThemePart, '') as ThemeName}>
				<ButtonFrame ref={ref} {...props}>
					{children}
				</ButtonFrame>
			</Theme>
		);
	},
);
```

Just to make sure, will this also work with more than one level of nesting?

For example, the [docs](https://tamagui.dev/docs/intro/themes#sub-themes) show how to define a `light_green_subtle` theme, where `subtle` is a sub-theme of `green`.

Reading the changed code this will result in `t_green_subtle` and `t_green_subtle_Button`, right? Will `t_green_subtle`, in this case, contain **all** CSS variables from its parent `green` (e.g. `--color4`, ...) or only those which are overriden by the `subtle` sub-theme (e.g. `--background`). If the latter is true, the CSS classes need to be built using all levels in a loop.
 * Example: `theme="light_green_subtle_Button"` would result in `class="t_green t_green_subtle t_green_subtle_Button"`

anhquan291:
Hi @krakowski , nested theme works fine for me — for example with Square, but Button still doesn’t. Looking into it more. Will get back to you soon.

krakowski:
Hey @anhquan291, I think `Square` works because [`defaultComponentThemes.ts`](https://github.com/tamagui/tamagui/blob/55cef8d6a0cf4cdf3d977e157fd7d3059f07b573/code/core/theme-builder/src/defaultComponentThemes.ts) doesn't contain a component theme for `Square`, which would override the CSS `class` property.

This bug only occurs (on Web) when a component theme is defined for the respective component.

natew:
Yea the issue maybe deeper, we do the single classname to override and generate CSS like green_Button so we avoid those specificity wars, it may break in the nested case, like having green + Sheet, and then blue + Button inside we probably want to test this and make sure its working.

the fix likely is just that we need to ensure the generated green_Button has the full css for green inside it

krakowski:
Hi @natew , adding the parent's stylesheet to sub-theme stylesheets should work from my point of view, although this also needs to be done at multiple levels, if I'm not mistaken. For example a `green_subtle_Button` theme would need both parents (`green` and `green_subtle`), right? Or would you let **all** sub-themes inherit their parents (i.e. `green_subtle` would also inherit `green`)?

anhquan291:
image

@krakowski Got it working, but it’s more complicated than I expected. I’ll need Nate to review it first.

Issue #3601, opened 2025-08-27
Theme colors with alpha are not rendered with alpha


Anthony-Gaudino:
### Current Behavior

If I define a color in my theme, for example: `"hsla(0, 0%, 100%, 0.1)"` and use this color as a background color, the alpha value is replaced by `1`: so I have always `hsla(0, 0%, 100%, 1)`.

Reproduction: [minimal_proj.zip](https://github.com/user-attachments/files/22008706/minimal_proj.zip)

### Expected Behavior

I expect Tamagui to maintain the alpha value in my theme colors.

### Tamagui Version

```markdown
1.132.20
```

### Platform (Web, iOS, Android)

```markdown
Google Chrome Version 139.0.7258.139 (Official Build) (arm64)
```

### Reproduction

```markdown
Please see attached zip file.

When running the application, a YStack square will be shown on index.
```

### System Info

```markdown

```

PR #3667, opened 2025-11-16
fix: resolve issue where hsla alpha value is ignored

oktaysenkan:
# Overview

This PR resolves an issue where alpha values defined in HSLA colors within theme tokens were incorrectly replaced with 1.

When defining a theme color such as:

```ts
hsla(0, 0%, 100%, 0.1)
```

and using it in styles, the resulting output becomes:

```ts
hsla(0, 0%, 100%, 1)
```

The alpha channel was being ignored and overwritten, causing transparency values to be lost.

**Cause**

During theme processing, HSLA values were normalized without preserving the original alpha component, defaulting to `1`.

**Solution**

Added proper parsing and serialization of HSLA values to ensure the alpha channel is correctly retained throughout the token transformation pipeline.

**Related Issues**

#3601

**Screenshots**

| Light Mode | Dark Mode |
|------------|-----------|
| Light Mode | Dark Mode |


natew:
fix looks good but i don't think we need a whole app to test it, could just have some one-off test somewhere either in the theme builder package or elswhere. if you don't have time for it i think @anhquan291 maybe able to just remove the playground and add more minimal testing for it elsewhere. since it is a pretty safe fix for an untested area we could also just leave out testing for now.

oktaysenkan:
I removed the playground app.

natew:
Thank you for this fix\! I've rebased, cleaned up the PR (removed unrelated yarn.lock and playground changes), and merged the core fix into #3682 which credits your work. Closing this in favor of the cleaned-up version.

PR #3683, opened 2025-11-25
toggle-group: fix active

anhquan291:


natew:
only thing is react is recommending moving away from cloning:

https://react.dev/reference/react/cloneElement

we may want to just rip this bandaid in v2 but it's not so simple, basically makes composable components kind of hard ultimately i think unless i'm missing something.

PR #3687, opened 2025-11-26
theme: add component-specific themes

anhquan291:


natew:
cleaning up lmk if we still need this

PR #3689, opened 2025-11-26
fix: add gridTemplateRows and gridArea CSS Grid properties

### LINKED ISSUE #3686: Missing support for grid-template-rows css property
Currently, we're able to use `gridTemplateColumns` and `gridTemplateAreas` properties to assign the web-only* `grid-template-columns` and `grid-template-areas` css properties - but not `gridTemplateRows` to assign `grid-template-rows`.

It looks like the types for it exist in some places but not others, and it isn't in the `stylePropsUnitless` object in https://github.com/tamagui/tamagui/blob/f1f1752e54d67ccfeafea85d379cb7373b3848fc/code/core/helpers/src/validStyleProps.ts#L92

*And looks like we [aren't far](https://github.com/facebook/yoga/pull/1865) from them existing in React Native too!

natew:
## Summary
- Adds `gridTemplateRows` and `gridArea` to `validStyleProps.ts`

Fixes #3686

---

cc @anhquan291


natew:
would need to be in types as well

PR #3691, opened 2025-11-26
fix: add role=dialog and aria-modal to DialogContent for accessibility

### LINKED ISSUE #3314: Dialog component has accessibility violation
Using VoiceOver on a Mac in Chrome:

When I activate the "Show Dialog" button in this page: https://tamagui.dev/ui/dialog, the VoiceOver cursor is not moved inside the dialog, and is instead left behind on the page, unable to get into the dialog.

Ideally, the rest of the page should be [inert](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/inert), and the VoiceOver cursor should be moved the heading of the dialog with the index set to -1 so the heading can't be tabbed to again.

natew:
## Summary
- Adds `role="dialog"` and `aria-modal` to DialogContentFrame for proper ARIA compliance

Fixes #3314

---

cc @anhquan291

### DIFF
--- a/code/ui/dialog/src/Dialog.tsx
+++ b/code/ui/dialog/src/Dialog.tsx
@@ -593,6 +593,8 @@ const DialogContentImpl = React.forwardRef<TamaguiElement, DialogContentImplProp
       <DialogContentFrame
         ref={composedRefs}
         id={context.contentId}
+        role="dialog"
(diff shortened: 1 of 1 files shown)


natew:
Merged into v2 branch

PR #3694, opened 2025-11-26
fix: preserve function props when using styled() on components with matching prop names

### LINKED ISSUE #2784: Using styled on a component with a matching prop name returning typescript error
### Current Behavior

I'm using `styled` to let FontAwesomeIcon use Tamagui styles.  In doing so, I had to overwrite some existing FontAwesomeIcon props with accept.  It works fantastic, and I'm able to use the size prop with Tamagui's size tokens.  However, typescript is throwing me this message of:

`
Type '"$6"' is not assignable to type '"unset" | Omit<never, "unset"> | undefined'.ts(2322)
index.tsx(12, 5): The expected type comes from property 'size' which is declared here on type 'Partial<{ theme?: ThemeName | null | undefined; icon: IconProp; secondaryOpacity?: number | undefined; maskId?: string | undefined; testID?: string | undefined; target?: string | undefined; ... 255 more ...; key?: Key | ... 1 more ... | undefined; }> & { ...; }'
`

bd4c762c-74d4-436c-81d1-bca57f132a8c
32e938b3-9e8a-46f6-9b5e-16ed250d315d

Here are the props being used in FontAwesomeIcon
image

I am unsure which definition it is using because '"unset" | Omit<never, "unset"> | undefined' is neither FontAwesome's, nor is it the standard `size?: SizeTokens` I've seen before.

### Expected Behavior

I expect the `size` prop on my new component to have a definition of `size?: SizeTokens`

### Tamagui Version

```markdown
1.102.1
```

```markdown
native
```

### Reproduction

```markdown
https://github.com/ryanespin/starter-free/blob/main/packages/ui/src/FontAwesomeIcon.tsx
```

### System Info

_No response_

natew:
## Summary
- Adds `PreserveFunctionProps` type utility to `GetFinalProps`
- When a component has a prop with a name matching a style prop (e.g., `color`) but with a function type, the original type is now preserved
- Previously, function-typed props would be overwritten by style prop types

Fixes #2784

---

cc @anhquan291

### DIFF
--- a/code/core/web/src/types.tsx
+++ b/code/core/web/src/types.tsx
@@ -2349,13 +2349,24 @@ export type Styleable<
   ParentStaticProperties
 >
 
+// Preserves function-typed props from the original component that would otherwise be
+// overwritten by style props with the same name (e.g., a `color` callback prop)
+type PreserveFunctionProps<NonStyleProps, StylePropsBase> = {
+  [K in keyof NonStyleProps as K extends keyof StylePropsBase
+    ? NonStyleProps[K] extends (...args: any[]) => any
+      ? K
+      : never
+    : never]: NonStyleProps[K]
+}
+
 export type GetFinalProps<NonStyleProps, StylePropsBase, Variants> = Omit<
   NonStyleProps,
   keyof StylePropsBase | keyof Variants
 > &
   (StylePropsBase extends Object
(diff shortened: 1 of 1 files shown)


natew:
risky - type perf looks bad in a sensitive area

PR #3698, opened 2025-11-27
detox testing

natew:


natew:
@anhquan291 we merged this right? i just did it from cli but just to double check i didnt miss anything. btw detox android still has been flaky so i made it not fail ci unfortunately.

PR #3702, opened 2025-11-29
refactor(input): align Input component with web standards

anhquan291:
Screenshot 2025-11-28 at 23 11 50


natew:
a few things i think would be nice -

- can we do TextArea as well?
- can we generate some basic tests for web and native? making sure a few things map over etc
- can we add a /v1 export like the others for the old version

anhquan291:
Sounds good. Will do

PR #3712, opened 2025-12-14
V2 render prop

natew:


natew:
Superseded by #3796 (feat/render-prop)

PR #3715, opened 2025-12-15
feat: fix SelectDemo component

DaveyEke:
This PR removes the unnecessary second arrow that was present in  `SelectDemo.tsx` for v2 that wasn't even present in v1 as seen on `tamagui.dev`


anhquan291:
Hi @DaveyEke have you checked kitchen sink for this demo? The icon should appear on native, right?

anhquan291:
Simulator Screenshot - iPhone 17 Pro - 2025-12-15 at 09 55 13
image
It should be like this @DaveyEke

PR #3720, opened 2025-12-16
fix: fixed slider-vertical thumb not moving in tandem with cursor on iOS

DaveyEke:
This PR fixes the issue of the Vertical Slider not moving in tandem with the cursor 

Here's a demo of what it looked like before:

https://github.com/user-attachments/assets/4cfe7e88-c006-408f-a457-6a79e1421655

And now after the fix:

https://github.com/user-attachments/assets/3baf5f2f-7958-4453-a7c3-81afdea253e1

I've tested on iOS, Android, and Web and they all work!

### DIFF
--- a/code/tamagui.dev/data/docs/components/slider/1.141.0.mdx
+++ b/code/tamagui.dev/data/docs/components/slider/1.141.0.mdx
@@ -1,0 +1,241 @@
+---
+title: Slider
+description: Drag to set values, vertically or horizontally.
+name: slider
+component: Slider
+package: slider
+demoName: Slider
+---
+
+<HeroContainer>
+  <SliderDemo />
+</HeroContainer>
+
+```tsx hero template=Slider
+
+```
+
+<Highlights
+  features={[
+    `Sizable, themed, works controlled or uncontrolled.`,
+    `Multiple thumbs support.`,
+    `Control steps and control with your keyboard.`,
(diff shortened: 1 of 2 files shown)


natew:
@DaveyEke wondering should we apply this to v1 too?

code/ui/slider/src/Slider.tsx
```diff
@@ -226,6 +227,7 @@ const SliderVertical = React.forwardRef<View, SliderVerticalProps>(
     const [state, setState_] = React.useState(() => ({ size: 0, offset: 0 }))
     const setState = useCreateShallowSetState(setState_)
     const sliderRef = React.useRef<View>(null)
+    const insets = isIos ? useSafeAreaInsets() : { top: 0, bottom: 0 }
```
natew: we can't just add a new dependency actually, does RN not have a way to get this? if not we'd need some way to either let users pass this in or something, in general we don't rely on third party libraries with native dependencies

DaveyEke:
> @DaveyEke wondering should we apply this to v1 too?

yeah makes sense

DaveyEke:
> @DaveyEke wondering should we apply this to v1 too?

done see [link](https://github.com/tamagui/tamagui/pull/3723)

natew:
Superseded by direct commit to v2 branch (8ca9571870) - changes ported from #3723 with cleanup.

PR #3727, opened 2025-12-18
fix: Android Input placeholder clipping at small sizes

DaveyEke:
This PR fixes the issue of the placeholder/text clipping in the input component when set to `size="$2"` on Android.

Before(Left: Android, Right: iOS)

Screenshot 2025-12-18 at 9 15 58 AM

After(Left: Android, Right: iOS)

Screenshot 2025-12-18 at 10 39 54 AM

### DIFF
--- a/code/ui/input/src/shared.tsx
+++ b/code/ui/input/src/shared.tsx
@@ -63,6 +63,13 @@ export const inputSizeVariant: SizeVariantSpreadFunction<any> = (
     ...fontStyle,
     ...buttonStyles,
     paddingHorizontal,
+    // Android fixes: reset padding and center text vertically
+    ...(!isWeb && {
+      textAlignVertical: 'center',
+      paddingVertical: 0,
+      paddingTop: 0,
+      paddingBottom: 0,
+    }),
   }
 }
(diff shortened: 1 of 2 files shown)


natew:
Nice fix, how come iOS looks too far down there on "Enter your details..."?

Any negative effects here? I wonder about TextArea which has multiple rows?

natew:
We can fix this in v1 probably as i dont see any reason a user would have relied on that messed up padding right?

DaveyEke:
> Nice fix, how come iOS looks too far down there on "Enter your details..."?

Just noticed it now, currently working on a fix

DaveyEke:
> Any negative effects here? I wonder about TextArea which has multiple rows?

yeah you have a point, I'm noticing some top padding while typing in the text area

DaveyEke:
> We can fix this in v1 probably as i dont see any reason a user would have relied on that messed up padding right?

yep

natew:
Superseded by consolidated fix that removes code duplication - inputHelpers now live only in @tamagui/input and tamagui re-exports from there.

PR #3728, opened 2025-12-18
Fix input placeholder android main

DaveyEke:
This PR reproduces the changes of this [other PR](https://github.com/tamagui/tamagui/pull/3727) for the main branch

### DIFF
--- a/code/ui/tamagui/src/helpers/inputHelpers.tsx
+++ b/code/ui/tamagui/src/helpers/inputHelpers.tsx
@@ -27,7 +27,6 @@ export const inputSizeVariant: SizeVariantSpreadFunction<any> = (
     paddingHorizontal,
     // Android fixes: reset padding and center text vertically
     ...(!isWeb && {
-      textAlignVertical: 'center',
       paddingVertical: 0,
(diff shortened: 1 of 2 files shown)


code/ui/tamagui/src/helpers/inputHelpers.tsx
```diff
@@ -38,6 +44,12 @@ export const textAreaSizeVariant: SizeVariantSpreadFunction<any> = (
   const lines = props.rows ?? props.numberOfLines
   const height =
     typeof lines === 'number' ? lines * getVariableValue(fontStyle.lineHeight) : 'auto'
+
+  // lineHeight messes up input on native
+  if (!isWeb && fontStyle) {
+    delete fontStyle['lineHeight']
+  }
```
natew: i dont' see any github issues for this online, where lineHeight messes up input. want to be sure we have the right fix

DaveyEke: This has been in the codebase for a while now and adding this to the textAreaSizeVariant once fixed an iOS text input bug.

PR #3729, opened 2025-12-19
fix: ensure theme key tracking for dynamic theme updates on iOS

DaveyEke:
This PR intends to fix the issue of the `ThemeDemo` in the kitchen-sink app not working properly on iOS.

This was before the fix(Left: Android, Right:iOS):

https://github.com/user-attachments/assets/3e5b778c-7eb3-4fe2-a2c4-8a16a9a0aedd

This was after the fix(Left: Android, Right:iOS):

https://github.com/user-attachments/assets/bb3e20bf-cdb5-4da4-8e8c-62ea667e7111


code/core/web/src/helpers/propMapper.ts
```diff
@@ -392,6 +385,13 @@ export const getTokenForKey = (
 
   const themeValue = theme ? theme[value] || theme[value.slice(1)] : undefined
 
+  // Ensure theme key tracking happens early for dynamic theme updates (fixes iOS)
+  // Accessing .val triggers the proxy getter which calls track(key)
+  if (themeValue) {
+    // eslint-disable-next-line @typescript-eslint/no-unused-expressions
+    themeValue.val
```
natew: not a deep enough fix, we want to avoid re-renders in many cases by avoiding tracking here for DynamicColorIOS stuff, this would de-opt us entirely and destroy a big performance gain we fought for.

instead we basically have a system in place think for this setup where when we mutate a theme it force re-render, we need to fix that, ideally if we could add a detox test for this would be cool way to validate it.

DaveyEke: Mhm, alright then. will look into it deeper

PR #3790, opened 2026-01-11
fix(web): nested Text components now inherit styles from parent

### LINKED ISSUE #3789: Nested Text breaks inheritance of some styles/props
### Current Behavior

When nesting text in tamagui the `numberOfLines` prop does not work if there are nested `Text` components. This does not follow expected behavior of nested test from react native or web and makes having a single line of text with decorated text inside a pain to achieve.

The issue is due to the following:

https://github.com/tamagui/tamagui/blob/ed492d6b45a078f94953a0f22ad3aebe9fb16b3d/code/core/web/src/views/Text.tsx#L20-L26

### Expected Behavior

Nesting Text should work like in web and react native and children should inherit styles from the parent. 

### Tamagui Version

```markdown
1.137.3
```

```markdown
Web
```

### Reproduction

```markdown
The text "This screen uses…" on the following:
https://codesandbox.io/p/github/tamagui/starter-free/main

Working example in normal react native:
https://snack.expo.dev/@chrisdrackett/number-of-lines
```

### System Info

```markdown

```

natew:
## Summary
- Fixes nested Text components not inheriting styles from parent (fontFamily, whiteSpace, color, letterSpacing, textTransform)
- Enables `numberOfLines` to work correctly with nested Text
- Nested bold text now inherits parent color

Fixes #3789

## Test plan
- [x] Added integration tests in kitchen-sink (`TextNestedInheritance.test.tsx`)
- [x] All 6 new tests pass
- [x] Existing text-related tests still pass

🤖 Generated with [Claude Code](https://claude.ai/code)


natew:
## Native Testing Note

The fix is web-specific since the main issues were:
- `whiteSpace: 'pre-wrap'` overriding parent's `nowrap` (from `numberOfLines`)
- CSS `fontFamily: 'unset'` blocking inheritance

On React Native, nested Text components inherit styles natively, and `fontFamily: 'unset'` gets resolved via `config.unset` to an actual font token.

However, it would be good to verify on native that:
1. `numberOfLines` works with nested Text
2. Color inheritance works for nested bold text
3. Font family inheritance works

@natew - let me know if you'd like me to add native-specific tests or investigate further.

natew:
@chrisdrackett this should be available in canary 1.144.1-1768163771506 - could you test and let us know if it works for you?

chrisdrackett:
@natew yeah, hopefully I can get to it tomorrow, I have basically the same context in my app, so I'll just see if I can take it out and get the same output from my tests!

chrisdrackett:
@natew the one thing I personally find strange (but I can just override in my own code if needed) is the following:

```jsx
<Text color="$red">Red Text <Text>Default Text Color Text</Text></Text>
```

I feel like like the text color should also inherit like it does in react-native and web instead of using the default from tamagui.

christophehurpeau:
Hello, this PR seems to broke our Text component which is using `Text` from `@tamagui/core` and not Text from tamagui.
It looks like something like this:

```tsx
import { Text } from "@tamagui/core";

const sizeVariant: VariantSpreadFunction<
  GetProps<typeof InheritableText>,
  Exclude<FontSizeTokens, number>
> = (size) => ({ fontSize: size, lineHeight: size, letterSpacing: size });

const weightVariant: VariantSpreadFunction<
  GetProps<typeof InheritableText>,
  Exclude<FontWeightTokens, number>
> = (weight) => ({ fontWeight: weight });

export const TextBody = styled(Text, {
  name: "TextBody",
  variants: {
    inherit: {
      false: {
        fontFamily: "$body",
        weight: "$regular",
        size: "$md",
        color: "$textColor",
      },
    },

    size: sizeVariant,
    weight: weightVariant,
    italic: {
      true: { fontStyle: "italic" },
      false: { fontStyle: "normal" },
    },
  } as const,
  defaultVariants: {
    inherit: false,
  },
} as const);
```

Usage example:

```tsx
  <TextBody size="$xs" theme="secondary">
    Some text with a <TextBody weight="$bold">bold</TextBody> text and a{" "}
    <TextBody italic>italic</TextBody> text inside.{" "}
    <TextBody theme="error">This text is also in error theme.</TextBody>
  </TextBody>
```

Result with 1.144.1:

image


Result with 1.144.2:

image

**With 1.144.2 color wrongly is marked as "inherit" but it should not as `inherit` prop is false. The default color inherit seems to override the color passed via the variant ?**

with 1.144.1:
image

with 1.144.2:
image

PR #3796, opened 2026-01-12
feat: migrate from tag to render prop

natew:
## Summary

- Replaces `tag` prop with new `render` prop for v2
- `render` accepts three forms:
  - **String** (e.g., `render="button"`) - renders as semantic HTML element, compiler optimized
  - **JSX Element** (e.g., `render={<a href="/about" />}`) - merges props with element, causes deopt
  - **Function** (e.g., `render={(props, state) => <Custom {...props} />}`) - full control with component state, causes deopt

## Changes

- Updated `@tamagui/core` types and createComponent to support render prop
- Added `mergeRenderElementProps` helper for JSX element prop merging
- Updated compiler (vite/webpack/static plugins) to handle render prop optimization
- Updated all 70+ UI components to use render instead of tag
- Updated animation drivers (CSS, React Native, Moti) for render prop support

## Test plan

- [x] 29 Playwright web tests for all render prop variants
- [x] Tests cover: styled() render prop, semantic HTML elements, form elements, heading elements, list elements, runtime override, JSX element render, function render with state
- [ ] CI passes

### DIFF
--- a/code/core/web/src/createComponent.tsx
+++ b/code/core/web/src/createComponent.tsx
@@ -627,7 +627,7 @@ export function createComponent<
 
     if (process.env.NODE_ENV === 'development' && time) time`theme`
 
-    elementType = Component || elementType
+    elementType = element || elementType
     const isStringElement = typeof elementType === 'string'
 
     const mediaState = useMedia(componentContext, debugProp)


natew:
Superseded by #3821 (v2-render-prop)

PR #3876, opened 2026-02-05
Fix DynamicColorIOS objects breaking backgroundImage/boxShadow CSS strings on iOS

npupko:
## Problem

`backgroundImage` with `linear-gradient()` and embedded theme tokens (like `$color3`) renders nothing on iOS when `fastSchemeChange` is enabled (the default in `@tamagui/config/v5`).

On iOS with `fastSchemeChange`, theme variable `.get()` returns DynamicColorIOS objects:
```js
{ dynamic: { light: '#F6DDE6', dark: '#3D1F2B' } }
```

In `propMapper.ts`, the regex replacement for embedded `$tokens` in CSS strings calls `getTokenForKey` → `resolveVariableValue` → `.get(undefined)`, then wraps the result with `String(r)`. On a DynamicColorIOS object this produces `'[object Object]'`, so the final gradient becomes:
```
linear-gradient([object Object] 0%, [object Object] 100%)
```
RN's `processBackgroundImage` then fails silently.

Same issue affects `boxShadow`, `textShadow`, `filter`, and `border` with embedded theme tokens.

## Fix

For the CSS string token interpolation path on native, force `resolveValues: 'value'` so `resolveVariableValue` returns `.val` (plain hex string) instead of calling `.get()` (which returns DynamicColorIOS).

This follows the same principle as the existing `shadowColor` guard in `resolveVariableValue` — both bypass DynamicColorIOS for style properties that don't support it. Theme reactivity is preserved because the `.val` getter still triggers `track(key, false)`, ensuring re-renders on scheme changes.

On web, behavior is unchanged (the `TAMAGUI_TARGET` check is dead-code eliminated).

## Tests

Added two regression tests in `getSplitStyles.ios.test.tsx` that:
- Create a properly proxied theme with `getThemeProxied` (so `.get()` returns DynamicColorIOS)
- Mock `doesRootSchemeMatchSystem` → `true`, `isIos` → `true`, `fastSchemeChange` → `true`
- Use `resolveValues: 'auto'` (matching real rendering)
- Verify `backgroundImage` and `boxShadow` resolve to hex strings, not `[object Object]`

All existing tests (native, web, iOS) continue to pass.

### DIFF
--- a/code/core/core-test/getSplitStyles.ios.test.tsx
+++ b/code/core/core-test/getSplitStyles.ios.test.tsx
@@ -1,6 +1,7 @@
 import { View, createTamagui } from '@tamagui/core'
 import { beforeAll, describe, expect, test, vi } from 'vitest'
 import { isColorStyleKey } from '../web/src/helpers/getDynamicVal'
+import { getThemeProxied } from '../web/src/hooks/getThemeProxied'
 
 // Set TAMAGUI_TARGET before importing getSplitStyles
 process.env.TAMAGUI_TARGET = 'native'
@@ -20,6 +21,10 @@ vi.mock('@tamagui/constants', async () => {
     isAndroid: false,
   }
 })
+
+vi.mock('../web/src/hooks/doesRootSchemeMatchSystem', () => ({
+  doesRootSchemeMatchSystem: () => true,
+}))
 
 // Now import from mocked modules
 import * as configModule from '../web/src/config'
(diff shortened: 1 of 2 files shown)


natew:
hm good diagnosis but i think on native i guess we want to try and map back to object-style if possible so we can keep the faster dynamiccolor

natew:
fix coming in next rc

PR #3879, opened 2026-02-05
tamagui: fix require native deps

anhquan291:


natew:
have a diff fix for this

PR #3890, opened 2026-02-07
package.json fixes

wootencl:
Was running into some build issues where the `create-menu` package was seeing the import from `focus-guard` as CJS. This seemed to resolve. Also checked to make sure all other package.json's are consistent and identified `context-menu` was missing `type`. Should resolve both issues.


natew:
thanks so much

PR #3901, opened 2026-02-10
Toast v2

anhquan291:


natew:
## Toast v2 Review

### Architecture: Good

The layering is solid — `ToastState` (global observer) → `ToastComposable` (React component tree) → `Toaster` (drop-in wrapper). Mirrors Sonner's proven architecture, properly adapted for cross-platform with Tamagui's animation system. The composable API (`Toast` + `Toast.Viewport` + `Toast.List` with `renderItem`) gives both simple and full-customization paths.

The v1 backward compat via `@tamagui/toast/v1` subpath is clean — no runtime cost for v2 users.

### Issues Worth Fixing

#### 1. `v2-toast` package is dead weight (high)
`code/ui/v2-toast/` still exists and is a dependency of `@tamagui/tamagui`. It has its own `ToastItem.tsx` (693 lines), `Toaster.tsx` (445 lines), and a near-duplicate `ToastState.ts`. Should be deleted entirely.

#### 2. Context value thrashing (medium)
`ToastComposable.tsx:394` — the context value is a plain object literal with ~20 fields, recreated every render. Any state change in `ToastRoot` triggers a new context → re-renders every `useToastContext()` consumer. With 4+ toasts, a height measurement change can cascade. The `setToastHeight` 2px threshold helps, but a `useMemo` on the context or splitting static/dynamic contexts would be proper.

#### 3. Drag math duplicated between web and native (medium)
`useAnimatedDragGesture.ts` (web) and `.native.ts` both have `resisted()`, `cappedExit()`, and offset calculation logic. The native file is better factored with `computeOffset()` / `computeExitDirection()` helpers. Extract shared math into `dragMath.ts` — ~40 lines of duplicated physics that would be a single-place change for threshold tuning.

#### 4. Timer reset behavior is contradictory (minor)
`resumeTimer` resets `remainingTimeRef.current = duration` then calls `startTimer()`. So after a hover, the toast gets a full fresh timer. But `pauseTimer` carefully tracks elapsed time. Either this is intentional (Sonner resets on hover-leave too) and should be documented, or it should use the remaining time.

#### 5. `Observer.toasts` array never shrinks (minor)
`addToast` pushes to `this.toasts`, but `dismiss()` only adds to `dismissedToasts` Set — never removes from the array. `getActiveToasts()` filters, but `getHistory()` exposes the growing array. Fine in practice but worth noting.

#### 6. Dead API surface: `style` / `className` on `ToastT` (minor)
Leftovers from Sonner's API — not wired to any rendering. Either remove or implement.

#### 7. `ReprogapateToastProvider` typo (minor)
In `ToastProvider.tsx:175` — exported in types.

### Animation Driver Abstraction: Clean

The `useAnimatedNumbersStyle` extension is well-designed — optional on the driver type, all four drivers implement it, Reanimated uses worklets for off-thread execution. The three-tier animation strategy (AnimatePresence for enter/exit, Tamagui `transition` for stacking, `useAnimatedNumber` for drag) is clearly documented. Web correctly uses direct DOM manipulation during gestures instead of going through the CSS driver.

Note: spring physics assume 60fps (`0.016` timestep) — on 120Hz the feel will vary slightly.

### Test Coverage Gaps

- **No animation driver variation** — all tests use CSS driver, but `useToastAnimations.ts` has `if (useDirectDom)` branching. Should have `.animated.test.tsx` tests for at least one non-CSS driver.
- **No native tests** — drag gestures (PanResponder/RNGH), native stacking, burnt/native dispatch untested.
- **`toast.promise()` API** not directly tested (only through UI buttons).
- **Reduced motion mode** not tested.
- **`toast.custom()` with JSX** not tested.

### Bottom Line

Core architecture is solid. The cleanup PR (#3979) already removed the biggest issues (duplication, reanimated deps, dead code). Remaining work: kill `v2-toast` package, extract shared drag math, memoize the context value, and add animation driver test variants. Everything else is minor polish.

PR #3902, opened 2026-02-10
feat: add compoundVariants support to styled()

DaveyEke:


code/compiler/static/src/extractor/createExtractor.ts
```diff
@@ -1983,6 +1983,64 @@ export function createExtractor(
             return acc
           }, [])
 
+          // expand compound variants at compile time
+          if (shouldFlatten && staticConfig.compoundVariants) {
+            for (const compound of staticConfig.compoundVariants) {
```
natew: this shouldnt be necessary at all - we call getSplitStyles which runs all the internal logic so we dont need to do this i think..

PR #3917, opened 2026-02-18
feat(toast): support native web notifications via browser Notification API

DaveyEke:
This PR adds the ability to use native browser toasts (through the Notifications API)


code/ui/toast/src/ToastComposable.tsx
```diff
@@ -299,6 +315,29 @@ const ToastRoot = React.forwardRef<TamaguiElement, ToastRootProps>(
           )
           return
         }
+
+        // Native dispatch: intercept NEW toasts before they enter React state
+        // so the in-app toast never renders. On failure, falls through to in-app.
+        const isNew = !toastsRef.current.some((t) => t.id === toast.id)
+        if (isNew && nativeRef.current) {
```
natew: this is off - can't be doing all this in the render loop this would definitely be an effect

PR #3934, opened 2026-02-27
fix(vite-plugin): fix CJS build crash from import.meta.resolve

lightstrike:
## Problem

`@tamagui/vite-plugin@2.0.0-rc.19` CJS build crashes on load. The CJS build at `dist/cjs/plugin.cjs` transpiles `import.meta.resolve()` into a broken shim:

```js
const import_meta = {};
const resolve = name => fileURLToPath(import_meta.resolve(name));
```

`import_meta` is an empty object, so `.resolve()` is `undefined`. This crashes immediately when Vite loads the plugin via the CJS entrypoint (the exports map routes `require` → `dist/cjs/index.cjs`).

**Repro:** Use `@tamagui/vite-plugin@2.0.0-rc.19` with Vite 6, run `vite build`:

```
TypeError: import_meta.resolve is not a function
  at resolve (node_modules/@tamagui/vite-plugin/dist/cjs/plugin.cjs:46:68)
  at config (node_modules/@tamagui/vite-plugin/dist/cjs/plugin.cjs:182:68)
```

## Fix

Replace `import.meta.resolve()` with a `createRequire`-based resolution that works in both ESM and CJS, following the existing pattern from `@tamagui/loader/src/requireResolve.ts`:

```ts
// before (broken in CJS)
const resolve = (name: string) => fileURLToPath(import.meta.resolve(name))

// after (works in both ESM and CJS)
const resolve = 'url' in import.meta
  ? createRequire(import.meta.url).resolve
  : require.resolve
```

In CJS the build tool shims `import.meta` to `{}`, so `'url' in import.meta` is `false` and it falls back to `require.resolve`. In ESM `import.meta.url` exists so it uses `createRequire`. `require.resolve` returns a file path directly, so `fileURLToPath` is no longer needed.

## Test plan

- [x] `bun run build` succeeds, CJS output contains no `import_meta.resolve` references
- [x] `node -e "require(...)"` smoke test passes for CJS entry
- [x] ESM import smoke test passes with `tamaguiAliases({ svg: true })` resolving correctly
- [x] `bun run test` — all 10 new unit tests pass (exports, aliases, plugin shape)

### DIFF
--- a/code/compiler/vite-plugin/tests/plugin.test.ts
+++ b/code/compiler/vite-plugin/tests/plugin.test.ts
@@ -1,5 +1,13 @@
-import { describe, expect, test } from 'vitest'
-import { tamaguiPlugin } from '../src/plugin'
+import { describe, expect, test, vi } from 'vitest'
+
+vi.mock('../src/loadTamagui', () => ({
+  loadTamaguiBuildConfig: vi.fn(),
+  getLoadPromise: vi.fn(() => null),
+  getTamaguiOptions: vi.fn(() => null),
+  ensureFullConfigLoaded: vi.fn(),
+}))
+
+const { tamaguiPlugin } = await import('../src/plugin')
 
 const expectedPluginNames = ['tamagui', 'tamagui-rnw-lite', 'tamagui-extract']
 


natew:
Closing as wontfix — the Vite plugin is ESM-only by design (`"type": "module"`). Adding documentation to clarify this requirement. Thanks for the report!

PR #3935, opened 2026-02-27
ci: skip integration tests for unrelated package changes

lightstrike:
## Summary

- Add a `changes` detection job using `dorny/paths-filter` to skip the ~15min kitchen-sink Playwright integration tests when a PR only touches unrelated packages (e.g., `vite-plugin`, `next-plugin`, `.github/`)
- Uses a positive allowlist of paths kitchen-sink actually depends on, so new unrelated packages are excluded by default
- `checks` and `unit-tests` jobs are unchanged and still run unconditionally

## Detail

Kitchen-sink uses webpack with `tamagui-loader` and `@tamagui/babel-plugin` — it has no dependency on the vite or next plugins. The allowlist includes:

`code/core/`, `code/ui/`, `code/packages/`, `code/kitchen-sink/`, `code/compiler/babel-plugin/`, `code/compiler/loader/`, `code/compiler/metro-plugin/`, `code/demos/`, `package.json`, `bun.lock`

## Test plan

- [ ] PR touching only `code/compiler/vite-plugin/` — `integration-tests` should skip
- [ ] PR touching `code/core/web/` — `integration-tests` should run
- [ ] `checks` and `unit-tests` still run unconditionally on both

### DIFF
--- a/.github/workflows/checks.yaml
+++ b/.github/workflows/checks.yaml
@@ -52,7 +52,31 @@ jobs:
       - name: Run unit tests
         run: bun turbo run test:web --filter='!@tamagui/kitchen-sink' --concurrency=1
 
+  changes:
+    runs-on: ubuntu-latest
+    outputs:
+      integration-relevant: ${{ steps.filter.outputs.integration-relevant }}
+    steps:
+      - uses: actions/checkout@v4
+      - uses: dorny/paths-filter@v3
+        id: filter
+        with:
+          filters: |
+            integration-relevant:
+              - 'code/core/**'
+              - 'code/ui/**'
+              - 'code/packages/**'
+              - 'code/kitchen-sink/**'
+              - 'code/compiler/babel-plugin/**'
+              - 'code/compiler/loader/**'
+              - 'code/compiler/metro-plugin/**'
+              - 'code/demos/**'
+              - 'package.json'
+              - 'bun.lock'
+
   integration-tests:
+    needs: changes
+    if: needs.changes.outputs.integration-relevant == 'true'
     runs-on: ubuntu-latest
     env:
       NODE_OPTIONS: '--max-old-space-size=6144'


natew (commented):
The approach is sound and the optimization is worth doing, but the allowlist is missing three compiler packages that kitchen-sink's integration tests actually depend on. Skipping tests when these change would give false confidence.

**Missing from the allowlist:**

The kitchen-sink webpack build uses `tamagui-loader` (from `code/compiler/loader/`), which is included — but `tamagui-loader` has direct workspace dependencies on packages that are *not* in the allowlist:

- `code/compiler/static/**` (`@tamagui/static`) — the core compiler logic; `tamagui-loader` depends on it
- `code/compiler/static-worker/**` (`@tamagui/static-worker`) — the worker used by `tamagui-loader`
- `code/compiler/static-sync/**` (`@tamagui/static-sync`) — used by `@tamagui/babel-plugin`, which is in the allowlist

You can verify via `code/compiler/loader/package.json` (depends on `@tamagui/static` and `@tamagui/static-worker`) and `code/compiler/babel-plugin/package.json` (depends on `@tamagui/static-sync`).

A change to `@tamagui/static`'s core evaluation logic, for example, would skip the integration tests entirely under this PR as written.

**Suggested fix** — add these three lines to the `integration-relevant` filter:

```yaml
- 'code/compiler/static/**'
- 'code/compiler/static-worker/**'
- 'code/compiler/static-sync/**'
```

**Minor note (not a safety issue):** `code/compiler/metro-plugin/**` is in the allowlist, but the kitchen-sink web tests use webpack, not Metro. Metro is only used for native tests. This means changes to `metro-plugin` will unnecessarily trigger the 15-minute Playwright suite. Not harmful, just slightly over-inclusive.

Once those three missing paths are added, this is a solid optimization.

natew:
on second thought - we do need the entire dependency chain to be analyzed... checking

PR #3937, opened 2026-02-28
fix(web): add lineHeight to unitless CSS properties

lightstrike:
## Summary

- Numeric `lineHeight` values (e.g., `1.15`) were incorrectly getting `px` appended during CSS extraction, producing `line-height:1.15px` instead of `line-height:1.15`
- Per CSS spec, unitless `line-height` is a multiplier of font-size — `1.15px` is an absolute ~1px height that causes text to visually collapse
- Added `lineHeight` to `stylePropsUnitless` (helpers) and `unitlessNumbers` (react-native-web-internals) so both runtime and compiler paths treat it as unitless
- Native/RN is unaffected — `normalizeValueWithProperty` bails on `!isWeb` before consulting the unitless list

## Trade-offs

- In React Native, `lineHeight: 62` means 62 absolute pixels. After this fix, on web it becomes `line-height: 62` (a 62x multiplier). However:
  - Font tokens resolve to CSS variables (strings), not raw numbers — unaffected
  - Users who need absolute px on web can write `lineHeight: '62px'`
  - This matches CSS spec and how other CSS-in-JS libraries handle `lineHeight`

## Test plan

- [x] Added 5 new tests to `getStylesAtomic.web.test.tsx`: numeric lineHeight, integer lineHeight, string percentage passthrough, opacity unitless guard, fontSize px guard
- [x] All 161 existing `core-test` web tests pass with zero regressions

### DIFF
--- a/code/core/core-test/getStylesAtomic.web.test.tsx
+++ b/code/core/core-test/getStylesAtomic.web.test.tsx
@@ -78,6 +78,49 @@ test(`outline longhands get doubled selector`, () => {
   expect(rule).toMatch(/\._[^\s]+\._[^\s]+\{/)
 })
 
+test(`numeric lineHeight should not get px appended`, () => {
+  const out = getCSSStylesAtomic({
+    lineHeight: 1.15,
+  })
+  const rule = out[0][StyleObjectRules][0]
+  expect(rule).toContain('line-height:1.15')
+  expect(rule).not.toContain('line-height:1.15px')
+})
+
+test(`integer lineHeight should not get px appended`, () => {
+  const out = getCSSStylesAtomic({
+    lineHeight: 2,
+  })
+  const rule = out[0][StyleObjectRules][0]
+  expect(rule).toContain('line-height:2')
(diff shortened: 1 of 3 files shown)


natew:
Tied into the flex one, this is one where if we're aligning to web then we'd want to do that, but it's a breaking change and would have to be in v3.

natew (approved):
LGTM. The fix is correct — CSS spec defines unitless `line-height` as a font-size multiplier, so appending `px` to numeric values was always wrong (e.g. `lineHeight: 1.15` producing `line-height: 1.15px` which collapses text to ~1px). Both `stylePropsUnitless` and `unitlessNumbers` need to be in sync for the runtime and compiler paths respectively, and the PR correctly updates both. The `\!isWeb` guard in `normalizeValueWithProperty` ensures native is unaffected. Font token values go through `getVarRef` and resolve to CSS variables (strings), so they're unaffected. The breaking change for users passing raw integer lineHeight props directly (e.g. `lineHeight={24}`) is warranted — the previous behavior was the bug. Tests are thorough.

PR #3938, opened 2026-02-28
chore: remove .claude/settings.local.json from version control

lightstrike:
## Summary

- Removes `.claude/settings.local.json` from git tracking — this file contains user-local, machine-specific Claude Code settings (personal permission overrides, disabled MCP servers) and should never be committed
- Updates `.gitignore` to specifically ignore `settings.local.json` while still allowing shared project config (`.claude/settings.json`, `.claude/skills/`) to be tracked

Per the [Claude Code settings docs](https://docs.anthropic.com/en/docs/claude-code/settings):

> **Local scope** (`.claude/settings.local.json`): Applies to **you, in this repository only**. Intended for personal overrides, testing configurations before sharing, and **machine-specific settings that won't work for others**.


natew:
did this on accident the other day, thanks for heads up

PR #3943, opened 2026-03-02
ci: add fetch-depth to checkout steps for faster CI clones

lightstrike:
## Summary
- Adds fetch-depth: 1 to 9 of 10 checkout steps across all CI workflows
- Keeps fetch-depth: 0 for release.yaml (needs full history for git tag/push)

## Context
The .git directory is 888 MB. Each full checkout downloads the entire history. With 3 parallel jobs in checks.yaml alone, that's ~2.5 GB per PR.

## Impact
- Clone per checkout: ~888 MB → ~50-80 MB (~90% reduction)
- Checkout time: ~30-60s → ~3-8s
- No functional impact: none of these workflows need git history

### DIFF
--- a/.github/workflows/build-ios-kitchensink-app.yml
+++ b/.github/workflows/build-ios-kitchensink-app.yml
@@ -51,6 +51,8 @@ jobs:
     steps:
       - name: Checkout
         uses: actions/checkout@v4
+        with:
+          fetch-depth: 1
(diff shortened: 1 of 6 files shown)


natew:
Thanks for this! The PR needs a rebase onto main to fix CI. Could you update the branch?

PR #3944, opened 2026-03-02
chore: add script to prune old pre-release tags

lightstrike:
## Summary
Adds `scripts/prune-old-tags.sh` to clean up 306 old pre-release tags from 2021-2023.

| Pattern | Count | Date Range |
|---------|-------|------------|
| v1.0.0-alpha.* | 69 | 2021-2022 |
| v1.0.0-beta.* | 17 | 2022 |
| v1.0.1-beta.* | 207 | 2022 |
| v0.2.0-alpha.* | 3 | 2021 |
| canary* | 10 | 2023 |

## NOT touched
- All stable release tags (v1.x.x, v0.x.x)
- v2.0.0-rc.* (active release candidates)

## Usage
```bash
./scripts/prune-old-tags.sh --dry-run  # preview
./scripts/prune-old-tags.sh            # execute
```

### DIFF
--- a/scripts/prune-old-tags.sh
+++ b/scripts/prune-old-tags.sh
@@ -1,0 +1,43 @@
+#!/bin/bash
+set -euo pipefail
+
+# Prune old pre-release tags superseded by stable releases.
+# Run with --dry-run first to preview.
+
+DRY_RUN=false
+[[ "${1:-}" == "--dry-run" ]] && DRY_RUN=true
+
+PATTERNS=(
+  'v1.0.0-alpha.*'
+  'v1.0.0-beta.*'
+  'v1.0.1-beta.*'
+  'v0.2.0-alpha.*'
+  'canary*'
+)
+
+TOTAL=0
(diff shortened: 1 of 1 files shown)


natew (approved):
Script looks safe and correct: `set -euo pipefail`, dry-run mode works, patterns are specific to the 296 old v1.0.x-alpha/beta and 10 old canary1.x tags, all stable tags and v2.0.0-rc.* are untouched. One minor note: the `canary*` glob would also match any future `canary2.*` git tags if the v2 release tooling ever uses that naming — but current v2 pre-releases use `v2.0.0-rc.*` tag format so this is low risk for a one-off cleanup script. Worth noting the PR bundles several unrelated changes (CSS animation reset fix, new v5-media breakpoints, popover hoverable test cases, useFloatingContext grace-period logic) alongside the script — the title/description only calls out the script.

natew:
Thanks! The PR needs a rebase onto main to fix CI. Could you update the branch? Also, the review noted some unrelated changes bundled in - might want to clean those up too.

PR #3946, opened 2026-03-02
chore: add Git LFS tracking rules for binary assets

lightstrike:
## Summary
- Adds .gitattributes rules to track binary file types via Git LFS going forward
- Covers: sketch, psd, otf, ttf, jpg, jpeg, png, mp4, gltf
- Forward-only — existing files in history are not affected
- New/modified binary files will automatically use LFS after this merges

## Context
The .git directory is 888 MB. Binary assets (sketch, fonts, images) total ~147 MB in HEAD and ~375 MB across history. This prevents future bloat.

## After merging
Contributors need `git lfs install` (one-time). GitHub Actions runners have git-lfs pre-installed.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,1 +1,12 @@
 package.json merge=packagejson
+
+# Git LFS tracking for binary assets
+*.sketch filter=lfs diff=lfs merge=lfs -text
+*.psd filter=lfs diff=lfs merge=lfs -text
+*.otf filter=lfs diff=lfs merge=lfs -text
+*.ttf filter=lfs diff=lfs merge=lfs -text
+*.jpg filter=lfs diff=lfs merge=lfs -text
+*.jpeg filter=lfs diff=lfs merge=lfs -text
+*.png filter=lfs diff=lfs merge=lfs -text
+*.mp4 filter=lfs diff=lfs merge=lfs -text
+*.gltf filter=lfs diff=lfs merge=lfs -text


natew (commented):
The `.gitattributes` change itself is fine in principle — the file types chosen (sketch, psd, otf, ttf, jpg, jpeg, png, mp4, gltf) are appropriate binary assets to track via LFS, and the forward-only approach is a reasonable starting point.

However, there are a few concerns before this can be merged:

**1. PR contains unrelated commits**

The PR includes 5 commits that have nothing to do with LFS: a popover hoverable fix, two version bump commits, a container query feature (`max-height-100` etc.), and a canary release bump. The diff consequently pulls in changes to `bun.lock`, ~170 `package.json` files, source files (`createAnimations.tsx`, `v5-media.ts`, `useFloatingContext.tsx`), and kitchen-sink test files. Please rebase onto current `main` so the PR only contains the single LFS commit.

**2. Missing font formats already in the repo**

There are ~145 `.woff2` files and other font formats committed to the repo (`code/core/font-inter/full/`, etc.) that are not covered by the proposed LFS rules. If the goal is to reduce binary bloat, `*.woff`, `*.woff2`, and `*.eot` should be included as well.

**3. No LFS infrastructure confirmation**

The PR description mentions GitHub Actions runners have git-lfs pre-installed, but there is no confirmation that LFS storage is enabled on the GitHub repo itself, or that the CI workflows have been audited to ensure `git lfs pull` is called where needed (e.g. font files used during builds). Worth verifying before merge.

**4. Existing binary files are not migrated**

The description acknowledges this is "forward-only," which is fine for now, but worth noting explicitly in follow-up: the existing 3,500+ binary files in the repo history won't benefit until a `git lfs migrate import` is run. That's a separate (larger) effort.

Please clean up the branch to contain only the `.gitattributes` commit and address the missing font format patterns, then this can be approved.

PR #3954, opened 2026-03-11
fix(detox): make iOS simulator device configurable via DETOX_DEVICE

lightstrike:
## Summary
- iOS simulator device was hardcoded to `iPhone 15` in `.detoxrc.js`
- Now configurable via `DETOX_DEVICE` env var, matching the existing `DETOX_AVD_NAME` pattern for Android
- Default remains `iPhone 15` for parity with current behavior

### Usage
```bash
# Default (iPhone 15)
bun run detox:run:ios

# Override device
DETOX_DEVICE="iPhone 16 Pro" bun run detox:run:ios
```

## Test plan
- [ ] `bun run detox:run:ios` works with default device
- [ ] `DETOX_DEVICE="iPhone 16" bun run detox:run:ios` picks up the override

### DIFF
--- a/code/kitchen-sink/.detoxrc.js
+++ b/code/kitchen-sink/.detoxrc.js
@@ -68,7 +68,7 @@ module.exports = {
     simulator: {
       type: 'ios.simulator',
       device: {
-        type: 'iPhone 15',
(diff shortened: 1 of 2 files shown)


natew:
Already included in #3956 which was merged. Closing as duplicate.

PR #3958, opened 2026-03-14
fix(core): fix focusWithinStyle

wwdrew:


natew:
Merged into main with refined tests covering CSS, styled(), animated (motion/css), blur, and re-render validation. Thanks @wwdrew for the fix!

PR #3960, opened 2026-03-17
fix(image): token-styles

anhquan291:
The v2 createImage.tsx uses StyledImage.styleable(), which means the styled system does correctly process props like `borderRadius="$4"` — resolving the token and injecting it into viewProps.style.

The issue occurs at line 178 in the original implementation:

style: {
  width: resolvedWidth,
  height: resolvedHeight,
}

This unintentionally overwrites rest.style, which already contains the resolved styles from the styled system (including borderRadius).

In this case, only width and height needed to be set inline (since setupReactNative.ts marks them as inlineProps, bypassing the styled system), but the current implementation ends up clobbering the entire style object.

Additional issue

There is also a second bug in the use of getTokenValue without specifying the 'size' group. Without the group hint, the lookup searches across all token categories, which can lead to incorrect matches — for example, $6 resolving to a radius token instead of a size token.

After applying the fix
simulator_screenshot_1ECCBEF3-92FC-4673-9C2F-6029E08B9BCF


natew:
Thanks for the detailed investigation and fix! We actually just merged a fix for the style clobbering issue on main today (`26594c384c`), so the core bug is addressed.

A couple notes on the approach here:

- **`usePropsAndStyle` is very slow** — we intentionally avoid it in the v2 Image. It adds a full runtime style pass which defeats the purpose of the styled system optimizations.

- **`source` is deprecated in v2** — we're moving to just `src` for the public API, so we don't want to add more `source` handling.

What we may want to do instead is just document that `width`/`height` props on Image are not responsive by default — if you need responsive sizing, use `useMedia()` and pass the values in explicitly. That way it's clear to the user that they're opting into slower runtime-based styling intentionally.

The `'size'` group hint on `getTokenValue` is a good catch though — we'll get that in separately.

PR #3995, opened 2026-04-23
fix(vite-plugin): import.meta.resolve crash in CJS

henryhobhouse:
## Problem

Version: v2 - problem persisting in latest (2.0.0-rc.41)

When `@tamagui/vite-plugin` is loaded in a CJS context the plugin crashes on startup with:

```
import_meta.resolve is not a function
```

**Root cause:** `plugin.ts` used `import.meta.resolve(name)` to resolve dependency paths relative to the plugin's own install location. When esbuild compiles the ESM source to `dist/cjs/plugin.cjs`, it shims `import.meta` to `const import_meta = {}` (an empty object). At runtime `import_meta.resolve` is `undefined`, so the first call to `resolve()` throws.

## Fix

Replace `import.meta.resolve` with [`createRequire`](https://nodejs.org/api/module.html#modulecreaterequirefilename) (already imported in the file, used elsewhere for project-root resolution). Use a runtime guard to pick the right anchor path for each context:

```ts
const _pluginRequire = createRequire(
  typeof __filename === 'string' ? __filename : fileURLToPath(import.meta.url)
)
const resolve = (name: string) => _pluginRequire.resolve(name)
```

- **CJS context** (`dist/cjs/plugin.cjs`): `__filename` is the CJS global (string) → `createRequire(__filename)` anchors resolution to the plugin file's own directory.
- **ESM context** (`dist/esm/plugin.mjs`): `__filename` is not defined → falls through to `fileURLToPath(import.meta.url)`, identical to the original behaviour.

No build-tooling changes needed — verified by inspecting the built CJS and ESM outputs.

### Out of scope

I ran and fixed formating issues in the impacted module as per the contributing guidelines. Not related but 🤷 .

### DIFF
--- a/code/compiler/vite-plugin/src/plugin.ts
+++ b/code/compiler/vite-plugin/src/plugin.ts
@@ -2,7 +2,6 @@ import type { TamaguiOptions, ExtractedResponse } from '@tamagui/static-worker'
 import * as Static from '@tamagui/static-worker'
 import { getPragmaOptions } from '@tamagui/static-worker'
 import { createHash } from 'node:crypto'
-import { existsSync } from 'node:fs'
 import { createRequire } from 'node:module'
 import path from 'node:path'
 import { fileURLToPath } from 'node:url'
@@ -15,7 +14,11 @@ import {
   ensureFullConfigLoaded,
 } from './loadTamagui'
 
-const resolve = (name: string) => fileURLToPath(import.meta.resolve(name))
+// handle ESM/CJS duality for plugin dependencies - resolve from plugin's location, not user's project
+const _pluginRequire = createRequire(
+  typeof __filename === 'string' ? __filename : fileURLToPath(import.meta.url)
+)
+const resolve = (name: string) => _pluginRequire.resolve(name)
 
 // shared cache across all plugin instances/environments via globalThis
 type CacheEntry = {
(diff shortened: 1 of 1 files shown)


code/compiler/vite-plugin/src/plugin.ts
```diff
@@ -2,7 +2,6 @@ import type { TamaguiOptions, ExtractedResponse } from '@tamagui/static-worker'
 import * as Static from '@tamagui/static-worker'
 import { getPragmaOptions } from '@tamagui/static-worker'
 import { createHash } from 'node:crypto'
-import { existsSync } from 'node:fs'
```
henryhobhouse, line 5: Unused.

code/compiler/vite-plugin/src/plugin.ts
```diff
@@ -355,23 +358,18 @@ export function tamaguiPlugin({
...
-        '@tamagui/core',
-        '@tamagui/web',
-        '@tamagui/toast',
-      ]) {
-        if (!userConf.resolve.dedupe.includes(id) && isInstalled(userConf.root || process.cwd(), id)) {
+      for (const id of ['tamagui', '@tamagui/core', '@tamagui/web', '@tamagui/toast']) {
+        if (
+          !userConf.resolve.dedupe.includes(id) &&
+          isInstalled(userConf.root || process.cwd(), id)
+        ) {
```
henryhobhouse, line 372: Unrelated formating fixes by oxlint.

PR #4012, opened 2026-05-26
fix(image): bridge accessibilityLabel <-> alt cross-platform

natew:
## Summary

- v2 Image dropped the bidirectional `accessibilityLabel` ⇄ `alt` mapping.
- On web, `` only honors `alt` — passing `accessibilityLabel` silently dropped the label and React warned about an unknown prop on ``.
- On native, RN's `Image` only reads `accessibilityLabel` — passing `alt` (HTML-style cross-platform code) silently dropped the label for screen readers.

Now web maps `alt ?? accessibilityLabel` → `alt`, and native maps `accessibilityLabel ?? alt` → `accessibilityLabel`. Either prop works on either platform.

## Test plan

- [ ] On web, `<Image src="..." accessibilityLabel="cat" />` renders `cat` with no React warning.
- [ ] On web, `<Image src="..." alt="cat" />` continues to render `cat`.
- [ ] On native, `<Image src="..." alt="cat" />` is announced as "cat" by VoiceOver/TalkBack.
- [ ] On native, `<Image src="..." accessibilityLabel="cat" />` continues to be announced as "cat".

### DIFF
--- a/code/ui/image/src/Image.tsx
+++ b/code/ui/image/src/Image.tsx
@@ -20,9 +20,16 @@ export const Image = StyledImage.styleable<ImageProps>(
       progressiveRenderingEnabled,
       resizeMethod,
       tintColor,
+      // bridge RN <-> web a11y: <img> only honors `alt`, but consumers writing
+      // cross-platform code reach for RN's `accessibilityLabel`. without this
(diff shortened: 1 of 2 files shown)


natew:
Closing — reverted on main.

PR #4013, opened 2026-05-26
docs(variants): note that variant names shadow HTML attributes

natew:
## Summary

I went looking for a code-level fix for an `unstyled` prop leaking to host elements on tooltip/popover paths, but the variant pipeline already handles this correctly — when a variant is declared anywhere in the inheritance chain, `isVariant` is true at line 468 and the prop is consumed before it can reach `viewProps`. My initial guard at the shortcut was unreachable dead code.

The real edge case worth documenting: **variant names own that prop**. If you pick a variant name that matches a real HTML attribute (`title`, `role`, `slot`, `form`, …) on a styled component that renders to a host element, the attribute is silently swallowed rather than forwarded.

## Changes

- **`code/tamagui.dev/data/docs/core/variants.mdx`** — new "Variant names and HTML attributes" section explaining the collision and how to avoid it.
- **`code/core/core-test/variantPropLeak.web.test.tsx`** — 5 regression tests asserting that variant props don't leak to host elements (direct, inherited via `styled` chain, variant value with no matching declaration, `.styleable` HOC chain mirroring the tooltip pattern, userland custom variant). Full web suite: 166 passed, no regressions.
- **`code/kitchen-sink/src/usecases/UnstyledLeakCase.tsx`** — manual reproducer for the soot tooltip pattern. Useful when chasing future reports of this class of issue.

## Test plan

- [x] `bun run test:web` in `core/core-test` — all 166 tests pass.
- [x] Manual kitchen-sink case loads and re-renders without producing the React-DOM "Received \`...\` for a non-boolean attribute \`unstyled\`" warning.

### DIFF
--- a/code/core/core-test/variantPropLeak.web.test.tsx
+++ b/code/core/core-test/variantPropLeak.web.test.tsx
@@ -1,0 +1,118 @@
+process.env.TAMAGUI_TARGET = 'web'
+
+import { getDefaultTamaguiConfig } from '@tamagui/config-default'
+import { render } from '@testing-library/react'
+import { beforeAll, describe, expect, test, vi } from 'vitest'
+import * as React from 'react'
+
+import { View, TamaguiProvider, createTamagui, styled } from '../web/src'
+
+const conf = createTamagui(getDefaultTamaguiConfig())
+
+const W = ({ children }: { children: React.ReactNode }) => (
+  <TamaguiProvider config={conf} defaultTheme="light">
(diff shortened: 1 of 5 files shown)


natew:
Closing — the bug was already fixed in v2.0.0 (5fe753e2c4, f7a1e8feec). soot's stale comment misled the investigation. The source-code guard was unreachable dead code. Docs change + regression tests not worth landing on their own.

PR #4021, opened 2026-05-30
ci(detox): launch-once directUseCase/remount conversion + CompilerExtraction-isolated shard matrix

natew:
Detox iOS speed + reliability work (split out from the combined branch; sheet-library fixes are in #4020).

## What
Every kitchen-sink e2e test now launches **straight into its use case** (`directUseCase` / deep-link remount) instead of `safeReloadApp()` + `navigateToTestCase()` (relaunch → home → quick-nav tap, ~60-90s/test). Plus:
- Lazy-load usecases registry so each launch evals only the navigated case (not all ~190).
- `CompilerExtraction`'s in-test `npx tamagui build` (2-27min, variable) isolated on its own shard; remaining files bin-packed by measured time, with the heavy anchors (`noRngh`, `SheetScrollableDrag`, `PressStyleNative`) on separate shards.
- launch/connect flake recovery + circuit-breaker; `App.native.tsx` directUseCase host.

## Measured (per-file, vs old)
MediaQueryGtMd 215→54s · TabsOnInteraction 277→60s · SelectAndroidOnPress 977→256s · SheetScrollableDrag 1463→732s · CompilerExtraction's build dominates its shard.

On a **healthy runner draw** the 4 shards land ~13-28min (all ≤30); shard 2 hit 22min even on a bad draw.

## Honest caveat
Per-shard wall time still has heavy **iOS Detox infra variance** — launch/connect hangs (~28min) and cold-Metro-bundle (2.6-22min) randomly inflate a shard to 40min+ (same matrix measured 13/28.7/26.4/33.4 one run, 42/22/40/41 the next). Guaranteeing ≤30 every run (and a 3-shard collapse) needs further infra work: pre-warm/cache the Metro bundle, cache the `@tamagui` build output, harden the launch/connect retry. This PR is the conversion + matrix foundation that those build on.

### DIFF
--- a/.github/scripts/check-ios-detox-shards.mjs
+++ b/.github/scripts/check-ios-detox-shards.mjs
@@ -5,7 +5,15 @@ const repoRoot = process.cwd()
 const workflowPath = join(repoRoot, '.github/workflows/test-native.yml')
 const e2eDir = join(repoRoot, 'code/kitchen-sink/e2e')
 
-const explicitlyExcludedTests = new Map()
+const explicitlyExcludedTests = new Map([
+  // android-only: every test skips on iOS (device.getPlatform() === 'ios'), so running
+  // it on the iOS sim produced zero coverage while burning ~6.6min. it still runs in the
+  // android Detox job, which executes all e2e files unsharded.
+  [
+    'SelectAndroidOnPress.test.ts',
+    'android-only (skips on iOS); runs in the android job',
+  ],
+])
 
 const workflow = readFileSync(workflowPath, 'utf8')
 const shardMatches = [...workflow.matchAll(/test_files:\s*'([^']*)'/g)]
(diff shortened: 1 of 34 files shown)


natew:
**Measured: all 4 shards ≤30 min ✅** (run 26692647501 on the matrix in this PR, healthy runner draw):

| Shard | Wall |
|---|---|
| 1/4 (CompilerExtraction isolated) | 19m31s |
| 2/4 | 27m11s |
| 3/4 | 25m21s |
| 4/4 | 19m46s |

Max 27m11s, all green. Note: per-shard wall still carries iOS-runner variance (the same matrix hit 42/22/40/41 on a degraded-runner draw), so ≤30 holds on a healthy draw but isn't yet *guaranteed* every run — the Metro-bundle/launch-connect infra hardening (described above) is the follow-up to make it robust. 3-shard collapse looks viable (~27min/shard projected) but is gated on CompilerExtraction's variable build (2-27min).

PR #4030, opened 2026-06-11
test: wait for reanimated hover spring to settle

natew:
## Summary
- make the reanimated stuck-hover spring regression wait for the actual base color instead of assuming the spring reaches exact RGB red within 450ms
- preserves the latch regression check while avoiding timing sensitivity on the current reanimated/web stack

## Local validation
- bun run test:web:seq tests/ReanimatedStuckHover.test.tsx -g "spring hover-out returns to base"
- bun run test:web:seq tests/ReanimatedStuckHover.test.tsx

Full suite is left to GitHub Actions per request.


natew:
Closing this since the useful reanimated-hover source/test hunks are already on main, and the branch is now stale/unmergeable with broad release-version churn. If the separate CI workflow changes are still wanted, they should come back as a focused PR.

PR #4038, opened 2026-06-15
fix: pin lodash to 4.17.23 (4.18.0 breaks kitchen-sink webpack build)

natew:
Hotfix for the regression from #4037. The lodash override bumped 4.17.23 -> 4.18.0, but 4.18.0 is a broken release: `_.template` throws `ReferenceError: assignWith is not defined`, which html-webpack-plugin calls to build kitchen-sink's HTML. Result: webpack compiled with errors -> blank page -> all integration tests timed out (50s) on main.

4.18.0 is the only non-vulnerable lodash, so there's no safe fix - pinning back to the original working 4.17.23 (lodash stays a documented high, gate is critical-level).

Verified locally: webpack compiles clean (0 errors), and ThemeReset + Toast integration suites pass (42/42).


natew:
Replaced by a clean hotfix branched off main (the original branch diverged post-squash).

PR #4042, opened 2026-06-20
fix(select): resolve Select.Viewport elevate variant so it stops leaking to the DOM

### LINKED ISSUE #3952: [V2] Select.Viewport leaks `elevate="true"` to DOM — variant cascade resolution skips ancestor variants in styled() output
## Bug

When using `<Select.Viewport>` (the default, styled variant), React logs a warning on every open in web/Next.js environments:

> Received `true` for a non-boolean attribute `elevate`.
> If you want to write it to the DOM, pass a string instead: elevate="true"

The `elevate="true"` attribute is written to the rendered `<div>`.

## Root cause

`SelectViewportFrame` outputs `elevate: true` from its `unstyled: false` variant:

```ts
// packages/select/src/SelectViewport.tsx
export const SelectViewportFrame = styled(YStack, {
  variants: {
    unstyled: {
      false: {
        elevate: true, // ← leaks to DOM
        bordered: true,
        // ...
      },
    },
  },
});
```

`elevate` is a named variant on `YStack` (via `@tamagui/stacks`) that should resolve to shadow styles via `getElevation`. However, when `elevate: true` appears as a _value inside another variant's output_ (rather than being set directly on the component), it is not re-processed through the variant system — it leaks as a raw prop to the DOM `<div>`.

The same class of fix was applied in rc.20 (`animatedBy`, `unstyled`, `marginLeft`) and rc.21 (`marginLeft`, `unstyled`), but `elevate` was not included.

## Consequences

1. **Shadow/elevation never applied** — `getElevation` is never called, so the dropdown viewport has no box-shadow on web, relying on z-index luck for stacking above other elements.
2. **DOM pollution** — `elevate="true"` is a non-standard HTML attribute on the dropdown `<div>`.
3. **Hydration risk** — If SSR and client paths handle `elevate` differently, React discards the server-rendered subtree, degrading LCP.

## Reproduction

```tsx
import { Select } from "tamagui";

// Open any Select — the warning fires on every open in the browser console.
<Select.Content>
  <Select.Viewport>...</Select.Viewport>
</Select.Content>;
```

## Workaround

Pass `unstyled` to `Select.Viewport` to bypass the broken default variant:

```tsx
<Select.Viewport unstyled minWidth={200} backgroundColor="$background" ...>
```

## Environment

- `@tamagui/select@2.0.0-rc.19` (unresolved through rc.22)
- Next.js 15, React 19, Turbopack

patrickwehbe:
Fixes #3952.

`SelectViewportFrame` extends `YStack`, which has no `elevate` variant. Its `unstyled: false` variant emits `elevate: true`, but because `YStack` cannot expand it, the value fell through to the DOM as `elevate="true"`. On web/Next.js that triggers React's "Received `true` for a non-boolean attribute `elevate`" warning on every open, and the elevation shadow was never applied (`getElevation` never ran).

This adds an `elevate` variant directly to `SelectViewportFrame` so the value resolves to shadow styles, the same way `ThemeableStack` and `SizableStack` wire it up. It mirrors the variant-cascade fixes from rc.20/rc.21 (`animatedBy`, `marginLeft`, `unstyled`).

After the change, `elevate` no longer reaches the DOM and the viewport gets its box-shadow.

I added a regression test in `code/core/core-test/selectViewportElevate.web.test.tsx` that runs the frame through `getSplitStyles` and checks:
- `elevate` is not present on the resulting view props (styled and unstyled)
- the resolved styles include the expected `boxShadow`

Tests pass, and `bun run turbo run build --filter=@tamagui/select` builds clean (types included). Lint and format pass on the changed files.

Note: while here I noticed `bordered: true` from the same `unstyled: false` block leaks to the DOM in the same way. I left that out to keep this PR focused on #3952, but happy to follow up on it.

### DIFF
--- a/code/core/core-test/selectViewportElevate.web.test.tsx
+++ b/code/core/core-test/selectViewportElevate.web.test.tsx
@@ -1,0 +1,70 @@
+process.env.TAMAGUI_TARGET = 'web'
+
+import { SelectViewportFrame } from '../../ui/select/src/SelectViewport'
+import { beforeAll, describe, expect, test } from 'vitest'
+
+import config from '../config-default'
+import {
+  StyleObjectProperty,
+  StyleObjectValue,
+  createTamagui,
+  getConfig,
+} from '../web/src'
+import { simplifiedGetSplitStyles } from './utils'
+
+let lightTheme: any
+
+beforeAll(() => {
+  // @ts-ignore
+  createTamagui(config.getDefaultTamaguiConfig())
+  lightTheme = getConfig().themes.light
+})
+
+function findRuleValue(rulesToInsert: Record<string, any>, property: string): any {
+  for (const rule of Object.values(rulesToInsert)) {
+    if (rule[StyleObjectProperty] === property) {
+      return rule[StyleObjectValue]
+    }
+  }
+  return undefined
+}
+
+// Regression for #3952: the styled() `unstyled: false` variant emits
+// `elevate: true`. Because SelectViewportFrame extends YStack (which has no
+// `elevate` variant), that value used to leak to the DOM as elevate="true"
+// (React "non-boolean attribute" warning) and the shadow was never applied.
+describe('Select.Viewport elevate resolution (#3952)', () => {
+  test('does not leak `elevate` to the DOM (styled default)', () => {
+    const { viewProps } = simplifiedGetSplitStyles(
+      SelectViewportFrame,
+      { unstyled: false },
+      { theme: lightTheme, themeName: 'light' }
+    )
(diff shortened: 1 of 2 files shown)


natew:
Thanks for this! The underlying bug was fixed on main via 5601bb17aa (SelectViewportFrame now extends ThemeableStack so elevate/bordered resolve through its variants), which made the variant shim here redundant — but your regression test was the missing piece, so it landed on main with attribution as 1072fe819b. Verified it passes against the ThemeableStack implementation. Closes out #3952.

PR #4047, opened 2026-06-23
perf(v2): optimize flat style and native render paths

natew:
Stacked on #4046 / v2-flat-styles.

This keeps the performance branch diff focused on the follow-up work after the flat style branch.


natew:
Superseded and split into two disjoint PRs:
- #4064 (v2-tailwind) — tailwind mode + third-style-mode removal + the full benchmark/comparison/conformance infra.
- #4065 (v2-perf2) — the pure-performance work redone cleanly off main: native deopt + partial-flatten + drop-dead-hover, native hook gating (data-disable-theme/media/events), use-element-layout refactor, and the gated $group/$theme->CSS web extraction. No tailwind/flat/bench code.

The runtime theme/media perf chain from this branch already landed in main. Closing in favor of #4064 + #4065.

PR #4064, opened 2026-07-01
Tailwind mode (v3): consolidate onto main, remove third style mode, wire converter tests

natew:
## Tailwind mode (v3) — single consolidated branch

Consolidates the tailwind-mode work (previously stranded on the local-only `flat-styles` branch) onto **current `main`** as one clean branch, removes the third/flat style mode per direction, and gets the converter tests running.

### What this ships
- **`styleMode` setting** on the Tamagui config: `'tamagui' | 'tailwind' | 'tamagui-and-tailwind'` (default `'tamagui'`). Additive and opt-in — existing configs are unaffected.
- **Runtime `className` → style conversion** (`code/core/web/src/helpers/getSplitStyles.tsx`): sizing/spacing (Tailwind `N × 0.25rem` scale), palette colors, named radii, flex/alignment, opacity, min/max prefixes, fractions/keywords. Unknown classes pass through.
- **`@tamagui/config/v6`** — Tailwind v4-compatible base (289-color palette generated oklch→sRGB, v4 radii, `w`/`h` shorthands).
- **`@tamagui/to-tailwind`** — Babel source codemod (Tamagui JSX props → tailwind `className`), for docs/bento codegen and downstream migration.
- **Docs + site**: `/docs/intro/tailwind-mode`, `docs/tailwind-mode.md`, `?syntax=tailwind` toggle, and the `/tailwind` promo page.
- **Conformance harness + benchmarks** (`code/comparisons/*`): pixel-parity vs real Tailwind v4 — **web 94% (115/122), native iOS 97% (116/120)**.

### Consolidation changes in this branch (the review-worthy delta on top of the feature)
- **Removed the third (flat) style mode.** The `'flat'`/`'tamagui-and-flat'` values were already gone from the union; this deletes the remaining dormant scaffolding (`IncludesFlatMode`, `WithFlat*` types and their `StackStyle`/`TextStyle`/`GetFinalProps` branches) and the duplicate runtime gate `isFlatModeEnabled` (it was byte-identical to `isTailwindModeEnabled`). The internal flat-`$`-prop machinery that tailwind conversion rides on is retained. Net: exactly one path — classic tamagui or tailwind, plus `tamagui-and-tailwind` interop.
- **Wired the converter tests into CI** — added a `test:web` script to `@tamagui/to-tailwind` (scoped to `src`) so the 39 transform tests run in the turbo `test:web` job; bumped its version `2.0.0-rc.31 → 2.4.0` to match the monorepo.
[...]

PR #4065, opened 2026-07-01
perf(v2): native deopt + partial-flatten, hook gating, gated $group/$theme CSS extraction

natew:
## What

The pure-performance subset of `v2-perf` (#4047), redone cleanly off current `main`. No tailwind/flat-mode feature code, no benchmark scaffolding — those stay on the tailwind PR (#4064). This is a focused native-render + compiler-deopt change plus one gated web extraction.

Splits out of #4047 per the review note there ("the rest of v2-perf is a large, separate compiler/perf effort; it should land as its own PR"). The runtime theme/media perf chain from v2-perf is already in `main`, so it is **not** re-included here.

## Commits

**Native compiler correctness + partial-flatten (the win)**
- `de-opt hoverStyle / $theme-* / $group-* on native` — correctness: these were flattening into the RN StyleSheet under literal keys RN treats as unknown props, so the styles silently never applied. Now preserved on the runtime path (mirrors pressStyle/focusStyle).
- `deopt native runtime-only props before split`, `drop native hover-only props` — native has no hover, so `hoverStyle` is dead → dropped.
- `partial-flatten static props on native deopt, drop dead hover` — when an element must deopt (pseudo/group/dynamic), pre-merge its pure-static props into one `style={}` so the runtime skips its per-prop loop. Theme tokens / dynamic props stay inline. Bench (×RN, runs=5): rich 6.95→3.64, heavy 4.08→2.32, group 2.74→2.14, animated 4.18→3.51.

**Native runtime hook gating (consumes compiler flags)**
- `skip theme+media hooks on compiler-proven theme/media-free deopt leaves` — `data-disable-theme` / `data-disable-media`; reuses `useThemeState`'s existing `disable` early-return.
- `skip useEvents on compiler-proven event-free deopt leaves (data-disable-events)` — skips RNGH gesture setup + refs + main-thread press hook on leaves with no event surface.
- `refactor(use-element-layout): extract emitLayoutIfChanged + rework registerLayoutNode`.

**Web build-time extraction + gate**
- `extract $group-* / $theme-* into CSS at build time` — replaces the old blanket `BailOptimizationError` with `@container` / theme-class extraction for flattened elements.
- `guard $group-/$theme- CSS extraction so animated elements stay on the runtime path` — the Q2 gate (see below).

## The $group/$theme animation gate (Q2)

The v2-perf commit removed the old guard that bailed all `$group-` extraction. That guard existed because a JS animation driver (reanimated, **motion**, moti) interpolates in JS and needs hard values — a static `@container` class can't drive it. This PR restores the invariant, but correctly:

- gated on **whether the element animates**, not on the driver — `motion` is `isReactNative: false` yet still needs JS, so a driver check would miss it;
- non-animated elements are safe to extract on every driver (incl. CSS), matching "if it doesn't animate, we can";
- in practice `createExtractor`'s `deoptProps` already de-opts every animated element upstream (`animation`/`animateOnly`/`animatePresence` always; enterStyle/exitStyle on RN drivers), so the guard is a local, future-proof invariant rather than the primary mechanism — verified: it never fires for the currently-reachable cases, and only ever bails to runtime (never widens what extracts).

## Validation

- `static-tests` native (deopt + partial-flatten + `data-disable-*` emission): **31 passed**
- `static-tests` web (CSS extraction + 2 new animation-gate tests): **34 passed**
- `core-test` native (runtime consumption incl. `eventsDisableGate.native`): **93 passed**
- typecheck (`@tamagui/web`, `@tamagui/static`, `@tamagui/use-element-layout` + graph): **clean**

## Related
- #4064 (`v2-tailwind`) — tailwind mode + benchmark/comparison/conformance infra.
- #4047 (`v2-perf`) — superseded by this + #4064.

### DIFF
--- a/code/compiler/static-tests/tests/__snapshots__/babel.native.test.tsx.snap
+++ b/code/compiler/static-tests/tests/__snapshots__/babel.native.test.tsx.snap
@@ -1,5 +1,18 @@
 // Vitest Snapshot v1, https://vitest.dev/guide/snapshot.html
 
+exports[`$group-* on native should de-opt (preserve as inline prop) 1`] = `
+"const __ReactNativeView = require('react-native').View;
+const __ReactNativeText = require('react-native').Text;
+import { YStack } from 'tamagui';
+export function Test() {
+  return <YStack group="row">
+          <YStack backgroundColor="red" $group-row-hover={{
+      backgroundColor: 'green'
+    }} />
+        </YStack>;
+}"
+`;
+
 exports[`$gtMd only should NOT apply on small screens 1`] = `
 "const __ReactNativeStyleSheet = require('react-native').StyleSheet;
 const _sheet = __ReactNativeStyleSheet.create({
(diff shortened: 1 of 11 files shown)


natew:
Merged to main directly (f1ca0d1e68..38b2c9b55c) after review: kept the compiler CSS extraction for $group-/$theme-, native deopt fixes, partial-flatten, and the use-element-layout refactor. Dropped the runtime hook-skip gates (data-disable-theme/media/events) — the theme gate broke native theme switching for styleable HOC leaves (regression test added), and the conditional-hook pattern can flip hook order on fiber reuse. Follow-up b3106c905b restored the spec sync initial onLayout measurement on mount/host-swap.

PR #4075, opened 2026-07-04
fix(radio-group): remove stray v1 pressTheme prop that leaks to the DOM on web

### LINKED ISSUE #4074: RadioGroup.Indicator leaks `pressTheme` to the DOM on web (v2) — React unknown-prop console error
### Current Behavior

On web, selecting a radio option logs a React console error the moment `RadioGroup.Indicator` mounts (it only renders once its item is checked):

```
React does not recognize the `pressTheme` prop on a DOM element. If you intentionally want it to appear in the DOM as a custom attribute, spell it as lowercase `presstheme` instead. If you accidentally passed it from a parent component, remove it from the DOM element.
```

The attribute also ends up on the rendered `div`.

### Expected Behavior

No unknown props reach the DOM; no console error when selecting a radio option.

### Cause

`pressTheme` was a v1 prop and is no longer handled by the v2 style system (grepping the published `@tamagui/core` / `@tamagui/web` v2 dist shows no handling for it), but `RadioGroupIndicatorFrame` still sets it inside its `unstyled: false` variant:

https://github.com/tamagui/tamagui/blob/bd9497fa9c9b9e76b0cc7db7fe93c1e4d086b453/code/ui/radio-group/src/RadioGroup.tsx#L98

Since the styled system no longer consumes it, it passes through to the underlying element. As far as I can tell this is the last remaining `pressTheme` in the v2 packages (the only other occurrences are in the legacy `@tamagui/button` `v1/` build), so deleting that one line fixes it. We've confirmed the fix via a pnpm patch removing the line.

### Tamagui Version

2.4.0 (still present in 2.4.1 and on `main` as of the permalink above)

Web

### Reproduction

Render any radio group with an indicator on web and select an option — e.g. in the starter-free repo:

```tsx
import { Label, RadioGroup, XStack } from 'tamagui'

export function Repro() {
  return (
    <RadioGroup defaultValue="a">
      <XStack alignItems="center" gap="$3">
        <RadioGroup.Item value="a" id="repro-a">
          <RadioGroup.Indicator />
        </RadioGroup.Item>
        <Label htmlFor="repro-a">Option A</Label>
      </XStack>
    </RadioGroup>
  )
}
```

Open the browser console: the warning fires as soon as an item is checked (with `defaultValue` set, immediately on mount; otherwise on first click).

### System Info

Reproduced in an Expo SDK 57 app (react-native-web) and traceable by inspection of the published dist files, so system specifics shouldn't matter.

sstur:
Fixes #4074

`pressTheme` was a v1 prop that the v2 style system no longer consumes, but `RadioGroupIndicatorFrame` still sets it in its `unstyled: false` variant. Since nothing extracts it anymore, it passes through to the underlying element on web, putting a `presstheme` attribute on the DOM and logging a React unknown-prop console error the first time a radio option is selected (the Indicator only mounts once its item is checked).

This removes the line — v2 already applies press feedback via the Item's `pressStyle`, so there's no visual change. Verified with `bun run build` in `code/ui/radio-group`, and confirmed in an Expo SDK 57 / react-native-web app (via a pnpm patch of the published package) that the console error is gone and radio behavior is unchanged.

### DIFF
--- a/code/ui/radio-group/src/RadioGroup.tsx
+++ b/code/ui/radio-group/src/RadioGroup.tsx
@@ -95,7 +95,6 @@ export const RadioGroupIndicatorFrame = styled(YStack, {
         height: '33%',
         borderRadius: 1000,
         backgroundColor: '$color',
-        pressTheme: true,
       },
     },
   } as const,


natew:
Ended up getting this before i saw yours.

PR #4089, opened 2026-07-09
feat: add Adapt live slot core

natew:
## Summary

- replaces Adapt's portal/teleport content transfer with a no-portal live-publish slot
- adds public Adapt target primitives: `useAdaptTarget`, typed state, handoff, presence latch, capabilities, and render-callback escape hatch state
- adds PR-A spike proof docs plus kitchen-sink web and Detox e2e coverage, including the no-portal inline `Sheet.Frame` touch-coordinate case for CI device shards

## Gate Evidence

- Fable cleared the PR-A spike gate for web/context/a11y/focus plus measured state-preservation.
- State preservation result: current v2 remounts/resets on inactive -> adapted; candidate does the same, so it is no worse than v2 and documented in `docs/spikes/adapt-live-slot-pr-a-spike.md`.
- iOS proof is `code/kitchen-sink/e2e/AdaptLiveSlotSpike.test.ts`; now that #4088 is merged, CI auto-shards it onto both devices at this PR head SHA.

## Validation

- `cd code/ui/adapt && bun run build`
- `npx tsc -p code/ui/adapt/tsconfig.json --noEmit --pretty false`
- `TAMAGUI_TARGET=native npx vitest --run --config ../../packages/vite-plugin-internal/src/vite.config.ts Adapt.native.test.tsx`
- `npx tsc -p code/kitchen-sink/e2e/tsconfig.json --noEmit --pretty false`
- `npx tsc -p code/kitchen-sink/tsconfig.json --noEmit --pretty false 2>&1 | rg "AdaptLiveSlotSpike|playwright\\.config" || true`
- `git diff --check`

## Notes

- Sheet seam is intentionally deferred to Adapt PR B.
- Local Detox was not run; iOS proof is the CI device run.

### DIFF
--- a/bun.lock
+++ b/bun.lock
@@ -1350,6 +1350,7 @@
         "@tamagui/constants": "workspace:*",
         "@tamagui/core": "workspace:*",
         "@tamagui/demos": "workspace:*",
+        "@tamagui/focus-scope": "workspace:*",
         "@tamagui/font-inter": "workspace:*",
         "@tamagui/get-token": "workspace:*",
         "@tamagui/input": "workspace:*",
--- a/code/kitchen-sink/e2e/AdaptLiveSlotSpike.test.ts
+++ b/code/kitchen-sink/e2e/AdaptLiveSlotSpike.test.ts
@@ -1,0 +1,114 @@
+import { by, device, element, expect as detoxExpect, waitFor } from 'detox'
+import { remountDirectUseCase } from './utils/navigation'
+import { safeLaunchApp, withSync } from './utils/detox'
+
+const testElement = (id: string) => element(by.id(id)).atIndex(0)
+
+async function getText(id: string) {
+  const attributes = (await testElement(id).getAttributes()) as any
+  return attributes.text as string
+}
+
+describe('AdaptLiveSlotSpike', () => {
+  beforeAll(async () => {
+    await safeLaunchApp({
+      newInstance: true,
+      launchArgs: { directUseCase: 'AdaptLiveSlotSpikeCase' },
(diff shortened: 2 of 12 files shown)


natew:
Review fixes pushed in `477d7521f1`.

Changes:
- Dropped `targetCount` / `contentsCount` / `renderCallbackCount` from reactive state and `AdaptContext`; registration now uses refs only, since the dev checks read refs and there are no reactive consumers.
- Moved `AdaptSlotPublisher` publish/clear into the layout effect before `notify()`, so the external slot store is only mutated after committed renders.
- Kept `AdaptParentProps.portal` as Dialog compile-compat and commented it as unused until old paths are removed in PRs B-E.

Confirming render-callback shape: yes, PR A intentionally lands the escape hatch as two args, `(contents, adaptState)`, to preserve v2 `children(contents)` callback compatibility while adding typed state as the second argument. PR E docs/design follow-up should state that shape explicitly.

Validation rerun:
- `cd code/ui/adapt && bun run build`
- `npx tsc -p code/ui/adapt/tsconfig.json --noEmit --pretty false`
- `TAMAGUI_TARGET=native npx vitest --run --config ../../packages/vite-plugin-internal/src/vite.config.ts Adapt.native.test.tsx`
- `npx tsc -p code/kitchen-sink/e2e/tsconfig.json --noEmit --pretty false`
- `npx tsc -p code/kitchen-sink/tsconfig.json --noEmit --pretty false 2>&1 | rg "AdaptLiveSlotSpike|playwright\\.config" || true`
- `git diff --check`

PR #4107, opened 2026-07-09
fix(slider): visible track and active fill after backgroundActive token removal

natew:
The `SliderTrack => Slider` rename (92d826a932) changed SliderActive's background from $backgroundActive to $background, and $backgroundActive no longer exists in themes. On any plain $background surface both the rail and the active fill blend into the page, so sliders render as floating thumbs with no track (web and native). Track now uses $backgroundPress and the active fill uses $color, with a kitchen-sink regression test asserting both are visible and distinct.

Verified visually on kitchen-sink web + iOS simulator; same fix landed on v3-beta as 3c3c3593ec.

### DIFF
--- a/code/kitchen-sink/tests/SliderStyled.test.tsx
+++ b/code/kitchen-sink/tests/SliderStyled.test.tsx
@@ -1,0 +1,34 @@
+import { expect, test } from '@playwright/test'
+
+import { setupPage } from './test-utils'
+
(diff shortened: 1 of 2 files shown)


natew:
The equivalent fix is already on v3-beta (slider track $backgroundPress + visible fill, noted in next.md). Consolidated there; closing the standalone v2 PR.

PR #4119, opened 2026-07-10
fix(core): v2 animated custom component + variant token crashes

natew:
## Summary
- keep custom animation-driver components when `render` is a string or an intrinsic React element, forwarding the intrinsic tag to drivers that accept it
- preserve styled-context token values when an explicitly present runtime prop is `undefined`
- add regressions for animated `styled(Button)` and styled-context token inheritance

## Root causes

For animated components, the string `render` path replaced the animation driver before this branch. The remaining `Button` failure used `render={<button />}`: `getCustomRender()` replaced the driver later in the render pipeline, sending an RN style array directly to the DOM button. Intrinsic render elements now retain the animated component; drivers with `acceptRenderProp` receive the tag through `viewProps.render`.

The variant token crash was separate from the v3 boolean-token shorthand fix. `SwitchCustomIcons` passed `size={undefined}`. `mergeComponentProps()` treated that as an override of the Switch styled-context `$true` size token, so the placement functional variant evaluated `tokens.space[undefined].val`. Undefined context props now inherit the context token and are not recorded as overrides.

## Validation
- reproduced Bento crashes on `ButtonPulse`, `IconTitleSwitch`, and `SwitchCustomIcons`
- patched Bento npm dist and confirmed all three pages render and interact with zero page errors via Playwright screenshots
- `@tamagui/web`: 61 tests passed, no type errors
- styled-context token regression: 4 tests passed
- animated styled Button regression: CSS, native, reanimated, and motion drivers passed
- existing styled Switch regression passed
- root `bun run build`: 162/162 package builds passed

### DIFF
--- a/code/core/core-test/styledContextTokens.web.test.tsx
+++ b/code/core/core-test/styledContextTokens.web.test.tsx
@@ -1,7 +1,14 @@
 import { beforeAll, describe, expect, test } from 'vitest'
 
 import config from '../config-default'
-import { View, Text, createStyledContext, createTamagui, styled } from '../core/src'
+import {
+  View,
+  Text,
+  createStyledContext,
+  createTamagui,
+  mergeComponentProps,
+  styled,
+} from '../core/src'
 // Stack was removed in v2 - use View instead (same props)
 import { simplifiedGetSplitStyles } from './utils'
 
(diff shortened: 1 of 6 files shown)


natew:
Consolidated into v3-beta per review preference for a single v3 surface (the equivalent fixes are on v3-beta natively; this branch stays available if a v2 patch release is ever needed).

PR #4120, opened 2026-07-10
fix(animations): avoid forwarding nativeID to DOM

natew:
## Root cause

The Moti and Reanimated web drivers transform Tamagui props and then spread the result directly onto a raw `div`/`span`. React Native `nativeID` survives that transform, so React warns and emits a lowercase `nativeid` DOM attribute.

This strips `nativeID` at the raw DOM boundary in both custom wrappers while preserving every other transformed prop.

## Regression coverage

- Adds an animated kitchen-sink case with an explicit `nativeID`.
- Verifies the Reanimated web wrapper emits neither the attribute nor the React console warning.

## Validation

- `PORT=9010 bun run test:web:seq tests/AnimatedDOMProps.animated.test.tsx` — 1 passed, 3 expected driver skips.
- Core web suite — 23 files, 185 tests passed.
- `bun run build` — 162 tasks successful.
- Bento Moti `ButtonPulse` repro — 0 `[nativeID]`/`[nativeid]` nodes and no nativeID console warnings.

### DIFF
--- a/code/core/animations-moti/src/createAnimations.tsx
+++ b/code/core/animations-moti/src/createAnimations.tsx
@@ -91,8 +91,9 @@ function createTamaguiAnimatedComponent(defaultTag = 'div') {
       const props = result?.viewProps || {}
       const Element = render
       const transformedProps = hooks.usePropsTransform?.(render, props, stateRef, false)
+      const { nativeID, ...webProps } = transformedProps ?? props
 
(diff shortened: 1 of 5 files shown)


natew:
Consolidated into v3-beta (cherry-picked as 0204c58bd8). Branch stays available if a v2 patch release is needed.

PR #4124, opened 2026-07-10
v3 beta

natew:
> ### Trying v3? Copy one of these into your agent.
>
> v3 is live on the npm `beta` dist-tag. A new beta cuts automatically on every green
> push to `v3-beta`, so the version moves — resolve it, don't hardcode it.
>
> **New project:**
>
> ```
> Add Tamagui v3 to this project. Run `npx tamagui@beta setup` and follow the brief it
> prints, adapting it to this project's bundler and framework. Do not skip the
> verification step at the end, and tell me anything you had to skip or guess at.
> ```
>
> **Upgrading from v1 or v2:**
>
> ```
> Upgrade this project to Tamagui v3. Run `npx tamagui@beta migrate --from v2` and
> follow the brief it prints. Run the flat-values codemod in dry-run first, show me the
> report, and wait for me to read it before you write anything. Tell me every site the
> codemod refused and why.
> ```
>
> Both briefs are printed by the CLI, so they describe the version actually installed.
> Full walkthrough: [Set up with an agent](https://github.com/tamagui/tamagui/blob/v3-beta/code/tamagui.dev/data/docs/intro/agents.mdx),
> or [Trying the beta](#trying-the-beta) below for the manual steps.

The v3 release branch. Everything tamagui-side lands here: 1,621 commits past `main`. The branch is green on Checks, Registry, Maestro iOS and Detox native, and every green push cuts a beta to npm automatically.

v3 asks for a real migration. The `$` sigil is gone, conditional style props like `hoverStyle` and `$sm` are gone, and there is no runtime compatibility shim. A codemod and an AI migration skill do most of the work.

### 1. Resolve the version

Every Tamagui package publishes together at one version, so whatever `tamagui@beta` resolves to is also the version of `@tamagui/core`, `@tamagui/config`, `@tamagui/vite-plugin` and the rest.

```bash
npm view tamagui dist-tags
npm view tamagui@beta version
```

Betas are cut automatically whenever Checks goes green on a `v3-beta` push, numbered `3.0.0-beta.<run>.<attempt>`, so the tag moves under you. Pin the exact version if you want a reproducible install.

### 2. Install, then check

Pin `tamagui` and every `@tamagui/*` package to that same version. Mixed versions are the most common way to end up with two copies of the runtime and styles that silently do nothing.

```bash
V=$(npm view tamagui@beta version)
npm i tamagui@$V @tamagui/core@$V @tamagui/config@$V
npx tamagui@$V check
```

`tamagui check` reports version mismatches, duplicate installs, lockfile problems, a missing config, and any v2 style syntax still in your source. Run it again after each step below; it is the fastest signal that a migration step actually landed.

### 3. Install the skills

```bash
npx skills add https://github.com/tamagui/tamagui/tree/v3-beta/skills
```

Two skills, both readable at [`skills/`](https://github.com/tamagui/tamagui/tree/v3-beta/skills):

- **`tamagui`** — authoring v3: `styled()`, variants, flat values, tokens and themes, components, animations, what the compiler does.
- **`tamagui-upgrade-v3`** — the migration: how to sequence it, the report flag playbook, the hard cases the codemod deliberately refuses, and how to prove the result.

For your app's own tokens and components, `npx tamagui@$V generate-prompt` writes a `tamagui-prompt.md` the skills read.

### 4. Run the codemod

Run it from your project root, which is where its paths, its report and the `tsconfig.json` it reads type information from all resolve. Dry run first, read the report, then write.

```bash
npx @tamagui/codemod-flat-values@$V --report flat-values-report.md ./src
npx @tamagui/codemod-flat-values@$V --write --report flat-values-report.md ./src
```

It converts only what it can prove: Tamagui provenance traced through imports, ordering preserved, every emitted program re-parsed with the real grammar. The write is transactional, so a parse error anywhere means nothing is written, and a path matching no file exits 2 rather than reporting a clean corpus. Everything it will not convert stays authored and lands in the report with a labeled reason.

Expect roughly a third of sites to need judgement. Measured on two real apps: 80% converted cleanly on the Bento corpus (1,681 of 2,113) and 67% on a mid-size control-room app (1,007 of 1,504).

### 5. Everything the codemod cannot do

```bash
npx tamagui@$V migrate --from v2   # or --from v1
```

That prints the checklist of record: dependency updates, the config and theme moves, the Sheet codemod, and every deprecated API with its before and after. The prose version is the [upgrade guide](https://github.com/tamagui/tamagui/blob/v3-beta/code/tamagui.dev/data/docs/guides/how-to-upgrade.mdx), 18 numbered steps plus a verification checklist.

### What to expect while it is a beta

- **Restart Metro after changing Tamagui config.** Metro's transform cache key does not include the lowering-plan generation, so a warm session can reuse stale compiled output. `expo start -c` after a config change; this is documented and deferred, not a mystery.
- **Editor support is not installable yet.** `@tamagui/lsp` is not on npm at all, and the VS Code extension (`code/core/language-service-vscode`) is still `private: true` at `0.0.1`, so it has never reached the Marketplace. `lsp-build.yml` cross-compiles and handshake-tests all eight platform binaries on every push, but publishing is a manual dispatch. Until that runs, editor support works from a source build via `TAMAGUI_LSP_BINARY`, and Neovim is the only editor with committed config (`code/lsp/editors/neovim`).
- **The native fast path is opt-in and experimental.** It needs RN 0.82+, `react-native-nitro-modules`, a native rebuild, and `experimental.nativeFastPath` in `tamagui.build.ts`.
- **Native compiled benchmarks are invalid pending re-measurement**, and native runtime v3 is currently slower than v2 on most scenarios. See Known gaps.

## The styling change

In v2, a conditional style was a separate prop holding an object. In v3, the prop names the style family and the value carries its own conditions, spelled the way Tailwind spells them:

```tsx
// v2
<View bg="$background" hoverStyle={{ bg: '$backgroundHover' }} $sm={{ p: '$6' }} p="$4" />

// v3
<View bg="background hover:background-hover" p="4 sm:6" />
```

Two things fall out of that:

- **The `$` sigil is gone.** A bare name resolves against your config first. If it is a token or variable you configured, it wins. Otherwise it passes through as literal CSS.
- **No runtime compat path.** `hoverStyle`, `$theme-dark`, media and group objects, and sigil values are deleted from the engine.

### Overriding

A value like `bg="background hover:background-hover"` is split by CSS property, then by the set of conditions in front of each piece. Each piece is a clause. When something later sets the same prop, it replaces only the clauses whose conditions it repeats. So `bg="red"` at a call site replaces a styled component's unconditional background and leaves its hover background alone. In v2 the override wiped the whole prop.

Default component skins are authored in this syntax from one file, so `tamagui`, `tamagui/unstyled` and the shadcn registry item all come from the same source:

```tsx
export const ButtonFrame = styled(ButtonBehaviorFrame, {
  backgroundColor: 'background hover:background-hover press:background-press',
  borderColor: 'border-color hover:border-color-hover',
  cursor: 'web:pointer',
  opacity: 'press:0.7',
})
```

Design and decisions: [dom-tailwind-flat-values.md](https://github.com/tamagui/tamagui/blob/v3-beta/plans/dom-tailwind-flat-values.md) (25 numbered decisions), [v3-clause-precedence-spec.md](https://github.com/tamagui/tamagui/blob/v3-beta/plans/v3-clause-precedence-spec.md), [v3-rip-token-sigil.md](https://github.com/tamagui/tamagui/blob/v3-beta/plans/v3-rip-token-sigil.md).

## What you get

**Native theme changes with zero re-renders.** The new `@tamagui/native-registry` links eligible Fabric views to a small C++ registry and applies a theme change in a single ShadowTree commit, with no React render at all. It also fixes v2's worst compiler outcome, where one themed value kept both a hook and an HOC. On the fast path that component becomes plain data.

Opt-in and experimental. Needs React Native 0.82+, `react-native-nitro-modules`, a native rebuild, and `experimental.nativeFastPath` in `tamagui.build.ts`. Verified on iOS and Android simulators. The registry is dependency-generic, so media, press, group press and container queries can move onto it later.

**Faster compiled web.** Rerender beats v2 across the board (rich 0.64x, group 0.76x, heavy 0.80x) and heavy mount is 0.79x. Full numbers and method below.

**Smaller JS.** A compiled app is smaller than the v2 equivalent: -9,381 gzip whole-app. An app that keeps the full runtime pays +4,964 gzip, and the optional features that make up most of the rest are gated off by default in the bench profile. Both numbers are in the numbers section. A zero-runtime mode that removes the runtime from the bundle entirely is being built on this branch now ([design](https://github.com/tamagui/tamagui/blob/v3-beta/plans/v3-zero-runtime-mode.md)).

**Tailwind as a frontend.** `@tamagui/tailwind` is a separate package built on the same grammar as everything else, selected by import. Components from `tamagui` keep the prop frontend, so both authoring styles work in one tree.

**Semantic html elements, on both platforms.** `html.div`, `html.a` and the rest are exported from `tamagui` and `@tamagui/core`. Each one is an ordinary Tamagui component that renders the element, takes the regular style props plus the strict DOM props for that element, and applies the browser's element defaults explicitly so a page renders the same with or without a css reset. They work at runtime on native too, rendering through the native primitives, with the non-style web prop mapping in a `.native` platform extension so it costs the web bundle nothing. The compiler is an optimizer here rather than a requirement, and a test renders the same tree compiled and uncompiled and compares the host trees. One set of tables holds the tag, attribute, event and native-backing definitions, read by the compiler, the runtime and the type generator, with a conformance test against a pinned snapshot of the standard prop surface.

**Config v6**, aligned to Tailwind v4. 289-color palette generated from oklch, matching spacing and radii, kebab-case multi-part names like `border-color`, recipe themes with a tree generator, and no implicit core default tokens. `@tamagui/config/v5` stays for static compatibility.

**Editor tooling that reads your real config**: completions, diagnostics, hover with per-theme token resolution, inline color swatches, an eslint/oxlint rule, and a project checker. Details below.

**Web and tooling installs no longer pull in React Native.** `@tamagui/native` and `@tamagui/proxy-worm` declared `react-native` as a required peer, which npm installs automatically, dragging in a 97 MiB subtree (Hermes, React Native, the RN debugger frontend). Both are optional now. Every `react-native` value import in core lives in a `.native.ts` file that web bundlers never load.

**One grammar behind all of it.** `@tamagui/style-grammar` is a dependency-free package that owns prop and shorthand metadata, token categories, modifier parsing, per-property expansion, and deterministic hashing. The runtime, the compiler, the Tailwind frontend and the editor tooling all read it instead of each keeping hand-maintained inverse maps.

**Compiler.** Babel passes are replaced by Rolldown workspace builds and native TypeScript declaration emit. `@tamagui/babel-plugin`, `@tamagui/static-sync` and `@tamagui/static-worker` are deleted. New: direct style emission, semantic DOM element lowering, diagnostics reported from successful lowers, and the native fast path.

**Migration.** `@tamagui/codemod-flat-values` publishes as an `npx`-runnable binary with a write mode, transactional rewrites, custom media discovery and pinned legacy fixtures, plus the `tamagui` and `tamagui-upgrade-v3` agent skills in [`skills/`](https://github.com/tamagui/tamagui/tree/v3-beta/skills). The upgrade guide is 18 numbered steps plus a verification checklist: [how-to-upgrade.mdx](https://github.com/tamagui/tamagui/blob/v3-beta/code/tamagui.dev/data/docs/guides/how-to-upgrade.mdx). Run against the v2 Bento corpus, the codemod found 2,113 sites, classified 1,681 clean, and left 432 (20.4%) for review by hand.

**Site.** Tailwind syntax toggle with a sticky cookie and `?syntax=tailwind`, docs version picker with archived v1 and v2 docs, the v3 blog post and upgrade guide, a full hydration pass, and v2 visual parity restored on v3.

## Everything else that comes with v3

**Inline theme values.** Changing two values for one part of the tree used to mean defining a whole new named theme. Now:

```tsx
<Theme background-hover="blue4 dark:blue2">
  <SettingsPanel />
</Theme>
```

Listed keys override, everything else inherits. Theme values are tracked per key, so only components that read a changed key do any work, and on web styled components skip even that since patches compile to CSS custom properties. You can also declare your own variables in config (`focusRingColor`, `disabledOpacity`, whatever you want), and they behave like theme keys everywhere. This replaces `defaultProps`, which is deprecated in v3.

**Field and Form do validation together.** A behavior-only `Field` primitive plus a `Form` that coordinates fields across web and native. Labels, descriptions, errors and controls associate automatically. Validation is a function or a Standard Schema from Zod 4, Valibot or ArkType. Validity starts at `null`, so untouched controls are not announced or styled as valid or invalid. Async checks never delay submit, stale results cannot overwrite newer ones, and the first invalid control gets focus on both platforms. No built-in aesthetic styles, so copied skins can respond to `invalid`, `dirty`, `filled` and `focused` themselves.

**A public animation API.** v3 stops shipping framework-owned fades, so the animation surface that used to live inside the drivers is now exported: `useAnimationDriver`, `useAnimatedNumber`, `useAnimatedNumberStyle`, `useAnimatedNumbersStyle`, `useAnimatedNumberReaction`. Every animated component gets a typed `onTransition` lifecycle with phase, cause and finished, replacing the untyped enter-only `onDidAnimate`. Sheet publishes its live drag position through the same hooks, so a drag-linked backdrop fade is ordinary user code. On the CSS driver, animated numbers now run on a rAF ticker with real completion callbacks instead of an estimated-duration timer.

**Adapt handoff is unified.** Dialog, Popover and Select share one model. Open state lives above the Adapt parent, adapted Sheet content stays mounted through the slide-out, and each part owns its own presence lifecycle. Media flips no longer tear content out mid-exit, and overlay pointer locks release correctly.

**Sheet anatomy is explicit.** `Sheet.Frame` splits into `Sheet.Container` (layout) and `Sheet.Background` (the visual surface), with a codemod for the mechanical part.

**Select does multiple selection** on every path, web and native, with ordered `string[]` values, repeated form entries on web, a real `<select multiple>` under `native="web"`, and correct behavior when the list is adapted into a Sheet. Also new: `Select.Separator`, `data-state` on more parts, `onEscapeKeyDown` and `onInteractOutside` on `Select.Content`, and `forceMount` on `Popover.Content`.

**`optimizeFor` makes a render tradeoff explicit.** Tamagui tracks exactly which theme and media values each component reads, so a theme change only re-renders components that used what changed. That bookkeeping costs a little on first render. Set `optimizeFor: 'first-render'` to skip it. Web defaults to `updates`, native defaults to `first-render`, since initial render dominates on native and full-tree re-renders are cheaper without a DOM.

**FocusScope** renders a `display: contents` wrapper, takes children directly instead of a render function, and gains a `noFocus` mode that makes the scope focus-inert during handoffs.

**ScrollView no longer routes through react-native-web.** `@tamagui/scroll-view` has a real web implementation covering `scrollTo`, `scrollToEnd`, `getScrollableNode`, RN-shaped `onScroll`, `contentContainerStyle`, horizontal scrolling and indicator props.

**Accessibility and form fixes.** Tabs defaults `activationMode` to `automatic`, matching the docs and ARIA guidance. Slider participates in forms via `name`, rendering one hidden input per thumb. Separator renders `role="separator"` with the right orientation.

**Icons size like text.** `<Icon size="4" />` resolves through the font size scale, so it lines up with `4` text instead of following the space scale.

**Deprecated APIs are gone**: `focusable` (use `tabIndex`), `fullscreen` (use `position` and `inset`), `themeInverse` (use `theme="inverse"`, which is SSR-safe now), the `styleable` static (use `createStyledHOC`), forwardRef component wrappers, `inlineWhenUnflattened`, `$true` tokens, and runtime token stepping (`stepTokenUpOrDown`, `getTokenRelative`).

**Component packages moved.** `@tamagui/ui` holds unstyled behavior, `tamagui` holds the default skins and now ships skinned controls directly. v1 Button, imperative Toast, `Toaster`, `ThemeableStack` and `SizableStack` are removed. Added: `Surface` plus facets. Tooling also consolidates: `@tamagui/cli/vite` and `@tamagui/cli/metro` give you what the standalone plugin packages do, so there is one version to keep in sync (the standalone packages are still published and remain the smaller install if you only use one bundler).

## Types

What ships ([`types.tsx`](https://github.com/tamagui/tamagui/blob/v3-beta/code/core/web/src/types.tsx)):

```ts
type FlatClausePrefix = `${ClauseModifierName}:`
export type FlatStyleValue<T> = T | FlatClausePrefix | (string & {})
```

A prop completes its config-derived tokens, plus every bare modifier prefix (`hover:`, `sm:`, `dark:`) as a completable entry, plus an escape hatch. It deliberately does not offer the token-times-modifier cross product: at real vocabulary size that union has 1.26 million members, TypeScript answers `TS2590` and tsserver returns zero completions, and the intermediate options fail too. Measurements: [v3-static-types-feasibility.md](https://github.com/tamagui/tamagui/blob/v3-beta/plans/v3-static-types-feasibility.md), [v3-type-performance.md](https://github.com/tamagui/tamagui/blob/v3-beta/plans/v3-type-performance.md).

Two type fixes worth knowing about: `bg` (now the `background` family) is theme-aware again, offering tokens instead of only CSS keywords. And the published web entries are guaranteed react-native-free in their type closure, checked by a [conformance test](https://github.com/tamagui/tamagui/blob/v3-beta/code/core/web/src/dom/styleTypes.test-d.ts) and a [CI check](https://github.com/tamagui/tamagui/blob/v3-beta/scripts/check-dom-types-standalone.ts) that walks the transitive `.d.ts` closure. Before this, a react-native type import reachable from a web entry silently degraded `StyleDefinition` to `any` under `skipLibCheck`, so that entry was typechecking nothing.

## Tooling

Since types stop at completion, validation lives in tooling. Four surfaces, all reading the one [`@tamagui/style-grammar`](https://github.com/tamagui/tamagui/tree/v3-beta/code/core/style-grammar) registry:

1. [`@tamagui/lsp`](https://github.com/tamagui/tamagui/tree/v3-beta/code/lsp), a standalone Rust binary giving completions, diagnostics, hover and color decorators to any editor over LSP.
2. The VS Code extension, now a wrapper that spawns that binary. The tsserver plugin is gone, so VS Code cannot drift from Neovim, Helix or Zed.
3. `@tamagui/eslint-plugin`, whose `valid-flat-values` rule also loads as an oxlint external JS plugin (oxlint 1.56+).
4. `tamagui check` ([`@tamagui/language-service`](https://github.com/tamagui/tamagui/tree/v3-beta/code/core/language-service)), a project-wide checker producing the same diagnostics as code frames, running in-process with no platform binary for CI and agent tooling.

Hover shows a color token's resolved value per theme plus the final rgba, and a media modifier's actual query JSON. The server loads this repo's 13.48 MB config artifact (1,152 themes, 236 keys each) in 45 ms, holds the theme data in about 1 MB, and answers a completion in about a microsecond. Node's `JSON.parse` alone on the same file takes 52.7 ms and retains 40 MB. It stays responsive while a file is mid-edit, verified by replaying a realistic edit keystroke by keystroke. Design and measurements: [plans/v3-lsp-rust.md](https://github.com/tamagui/tamagui/blob/v3-beta/plans/v3-lsp-rust.md).

Distribution follows the esbuild model: an umbrella package with eight per-platform `optionalDependencies` gated on `os`, `cpu` and `libc`, so npm installs exactly one binary with no postinstall and no download step. `TAMAGUI_LSP_BINARY` points it at your own build. Setup for VS Code, Neovim, Helix, Zed, Emacs, Sublime and JetBrains is copy-paste in [the README](https://github.com/tamagui/tamagui/blob/v3-beta/code/lsp/npm/tamagui-lsp/README.md).

### Compared to Tailwind CSS IntelliSense

| | Tailwind IntelliSense | Tamagui |
|---|---|---|
| Value / class autocomplete | yes | yes (tokens + modifier prefixes) |
| Inline color decorators | yes | yes, incl. theme values and `/50` opacity |
| Hover preview | compiled CSS | token value, per-theme resolution, media query |
| Linting / diagnostics | yes | eslint rule + oxlint jsPlugin + language server |
| Project / CI checker | no | `tamagui check` |
| `rem` pixel equivalents | yes | no |
| Works outside VS Code | no (VS Code only) | yes, any LSP editor |
| Published to Marketplace | yes | no, still private `0.0.1` |

### What is not finished in the tooling

- The VS Code extension is unpublished, and so are the eight platform binaries. Everything works from a local build via `TAMAGUI_LSP_BINARY`. Cross-compilation for all eight targets runs in CI on every `code/lsp/**` change, and publishing gets wired in once those are reliably green.
- The language server does not report a site whose string has no closing quote (the element collapses into one flat `ERROR` node). Editors that auto-close quotes produce `bg=""` instead, which works. Both behaviors are tested.
- `tamagui check` is wired into the Checks workflow but finds no config artifact at the repo root, so the flat-value half currently checks zero style files there. Pointed at `code/tamagui.dev` it checks 400 files clean.
- The oxlint jsPlugin path has tests but this repo's own `.oxlintrc.json` does not enable it yet.
- The eslint plugin has exactly one rule, and there is no settings surface or `rem` pixel-equivalent display.

## New packages

`@tamagui/style-grammar` · `@tamagui/lsp` (+ 8 platform binaries) · `@tamagui/tailwind` · `@tamagui/to-tailwind` · `@tamagui/native-registry` · `@tamagui/language-service` (+ VS Code extension) · `@tamagui/eslint-plugin` · `@tamagui/codemod-flat-values` · `@tamagui/ui` · `@tamagui/v3-canary`

Removed: `@tamagui/babel-plugin` · `@tamagui/static-sync` · `@tamagui/static-worker`

## Numbers

Web, 30 retained rounds after 3 warmups, shuffled per round, seed 73129, clean tree. Values are the V3/V2 mean ratio, so below 1.00 means v3 is faster. Receipt: [v3-v2-web-benchmarks.md](https://github.com/tamagui/tamagui/blob/v3-beta/code/comparisons/output/v3-v2-web-benchmarks.md).

| compiled | mount | rerender |
|---|---:|---:|
| simple | 0.93x | 0.86x |
| rich | 1.03x | 0.64x |
| group | 0.93x | 0.76x |
| heavy | 0.79x | 0.80x |
| animated | 1.42x | 1.09x |

Compiled v3 wins on rerender everywhere and on heavy mount. Animated mount is the one clear regression (0.597 ms vs 0.420 ms, CI `[0.098, 0.256]`), a real 0.18 ms rather than noise. Runtime (uncompiled) mode is 1.23-1.32x slower on mount across the board and faster on simple and rich rerender.

**Bundle: a compiled v3 app is 9,381 gzip SMALLER than v2; a runtime app costs +4,964.** Compiled whole-app gzip is 86,637 against v2's 96,018. The runtime arm is 99,821 against 94,857. Compiled CSS is 67 gzip larger (1,107 vs 1,040).

The compiled win comes from the compiler proving narrow runtime domains (numeric arithmetic, bounded primitive array lookups, finite string values that pass real atomic style resolution) and emitting them as host inline styles. Once every host in a tree flattens, the whole emitter stack tree-shakes out. It follows that the win scales with how much of an app flattens, and an app with unflattenable components lands between the two numbers above.

The runtime cost is feature weight, not duplication, and the optional parts are now declared rather than unconditional. Three `experimental.webRuntimeFeatures` fields (`inlineThemeValues`, `styleValueGrammar`, `safeArea`) compile to build-time literals in Vite, Next, Metro and the loader; omitted means enabled, so nothing changes for an app that says nothing. Turning them off drops inline theme and config variables (2,542 gzip, no v2 counterpart), the modifier grammar that orders chained conditions (2,372 across the parser, precedence, registry and state tables), and safe-area resolution (409). It also strands 1,719 gzip inside `directStyle` as unreachable, taking it from 5,367 to 3,648 with no change to emit semantics. `getSplitStyles` is now the component splitter only, and is smaller than v2's. v2 had none of the chained-modifier surface.

Caveat on every figure here: all four benchmark arms use the `animations-css` driver, so these describe a CSS-driver app. The v3-vs-v2 delta is fair because both sides use the same driver, but an app shipping `animations-motion` pays for that driver on top, uncounted.

## Known gaps

- **Native compiled benchmark cells are invalid** pending re-measurement. Metro workers supplied project-relative filenames while the plan cache was keyed by absolute realpaths, so the Release bundle missed every lowering plan and shipped the runtime path. The bug is fixed, with a loud `metro/plan-miss` diagnostic and a regression test, but the campaign has not been rerun. Native runtime v3 is currently slower than v2 on most scenarios.
- **Bundle budgets are specified but not enforced.** `compare-webpack-stats.mjs` warns on gzip growth and exits successfully.
- **Android device gate:** `Accordion` animation is a real Fabric layout-prop gap and `AdaptLiveSlotSpike` is flaky. Both are documented and neither is a regression from this branch.
- Bento's `v3` branch merges alongside the tamagui.dev cutover, as a pair.

## Where to read more

| | |
|---|---|
| Styling design + decisions | [plans/dom-tailwind-flat-values.md](https://github.com/tamagui/tamagui/blob/v3-beta/plans/dom-tailwind-flat-values.md) |
| Clause merge precedence | [plans/v3-clause-precedence-spec.md](https://github.com/tamagui/tamagui/blob/v3-beta/plans/v3-clause-precedence-spec.md) |
| Acceptance contract | [plans/v3-evolution.md](https://github.com/tamagui/tamagui/blob/v3-beta/plans/v3-evolution.md) |
| Zero-runtime design | [plans/v3-zero-runtime-mode.md](https://github.com/tamagui/tamagui/blob/v3-beta/plans/v3-zero-runtime-mode.md) |
| Static types feasibility | [plans/v3-static-types-feasibility.md](https://github.com/tamagui/tamagui/blob/v3-beta/plans/v3-static-types-feasibility.md) |
| Type performance | [plans/v3-type-performance.md](https://github.com/tamagui/tamagui/blob/v3-beta/plans/v3-type-performance.md) |
| Release blocker list | [plans/v3-beta3-release-readiness.md](https://github.com/tamagui/tamagui/blob/v3-beta/plans/v3-beta3-release-readiness.md) |
| State of the release (July snapshot) | [docs/v3-beta-state-of-the-release.md](https://github.com/tamagui/tamagui/blob/v3-beta/docs/v3-beta-state-of-the-release.md) |
| Native fast path | [plans/native-fast-path.md](https://github.com/tamagui/tamagui/blob/v3-beta/plans/native-fast-path.md) |
| Migration guide | [how-to-upgrade.mdx](https://github.com/tamagui/tamagui/blob/v3-beta/code/tamagui.dev/data/docs/guides/how-to-upgrade.mdx) |
| Benchmark receipts | [code/comparisons/output/](https://github.com/tamagui/tamagui/tree/v3-beta/code/comparisons/output) |
| Rust language server design | [plans/v3-lsp-rust.md](https://github.com/tamagui/tamagui/blob/v3-beta/plans/v3-lsp-rust.md) |
| Tooling packaging + react-native weight | [plans/v3-tooling-package-consolidation.md](https://github.com/tamagui/tamagui/blob/v3-beta/plans/v3-tooling-package-consolidation.md) |

A green Checks run on a `v3-beta` push publishes a `3.0.0-beta.<run>.<attempt>` to the npm `beta` dist-tag, and nothing else here publishes. The branch version is `2.7.7`; a stable release stays an explicit, gated owner dispatch.


code/tamagui.dev/components/bento-showcase/CodeWindow.tsx
```diff
@@ -103,44 +103,37 @@ export function CodeWindow({ code, isLoading }: Props) {
...
       >
         {tabs.map((tab, i) => {
           const isActive = String(i) === activeTabIndex
 
           return (
             <Tabs.Tab
-              bg={isActive ? '$color1' : '$color2'}
-              paddingVertical={'$4'}
-              key={tab.name}
+              bg={`${isActive ? 'color1' : 'color2'} hover:${isActive ? 'color2' : 'color3'} focus:${isActive ? 'color1' : 'color3'}`}
```
natew, line 123: would be way cleaner to just have an outer isActive ? `` : `` with the full styles separated

but either way this is a good test case for the compiler should be able to static optimize both of these types of conditionals actually.

PR #4125, opened 2026-07-11
fix(to-tailwind): borderWidth={1} emits bare `border`, not dangling `border-`

natew:
## what

The `@tamagui/to-tailwind` converter emitted a dangling `border-` class for
`borderWidth={1}` (Tailwind's `border` = 1px). Root cause: `transform.ts`
`propValueToClass` joins `${prefix}-${tailwindValue}`, and the 1px case returns
an empty `tailwindValue`, producing `border-` instead of the bare `border`.

## reproduction (before)

```
<View borderWidth={1} />                          => className="border-"
<View borderWidth={1} borderColor="$borderColor" /> => className="border- border-borderColor"
```

## after

```
<View borderWidth={1} />                          => className="border"
<View borderWidth={1} borderColor="$borderColor" /> => className="border border-borderColor"
<View borderWidth={2} />                          => className="border-2"   (unchanged)
```

## fix

When `tailwindValue` is empty, emit the bare `prefix` (the prefix itself is the
class) instead of `${prefix}-`.

## proof

- Added a regression test asserting `border border-borderColor` and no dangling
  `border-`.
- `@tamagui/to-tailwind` `test:web`: **40/40 pass**.

## context

Surfaced while converting Soot's `templates/app` to Tailwind against the v3
beta: across all 70 template `.tsx`, this was the sole source of dangling
classes; with the fix, the bulk codemod produces **zero** dangling classes.

### DIFF
--- a/code/core/to-tailwind/src/__tests__/transform.test.ts
+++ b/code/core/to-tailwind/src/__tests__/transform.test.ts
@@ -39,6 +39,13 @@ describe('tamaguiToTailwind', () => {
       expect(output).toContain('opacity-50')
     })
 
+    test('borderWidth 1 emits bare `border`, not `border-`', () => {
+      const output = tamaguiToTailwind(`<View borderWidth={1} borderColor="$borderColor" />`)
+      expect(output).toContain('border border-borderColor')
+      // regression: the 1px default must not leave a dangling `border-`
+      expect(output).not.toMatch(/border-(?=\s|")/)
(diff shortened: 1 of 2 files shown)


natew:
Closing as superseded: the borderWidth={1} converter fix is already in v3-beta. The remaining branch delta is directional regression-test coverage that we do not need to carry forward.

PR #4132, opened 2026-07-13
feat(core): v3 B1 styled() — compound variants, context props, class strings

natew:
## Summary

- add compound variant matching across direct, default, and styled-context props
- type styled-context keys as consumed component props
- add class-string styled() overloads and static class handling
- preserve call-site style and className precedence across web and native

## Validation

- bun install
- @tamagui/style-grammar build: passed
- @tamagui/web build: passed
- @tamagui/web type tests: 6 files, 83 tests, zero type errors
- focused web compound/variant suites: 4 files, 73 tests passed
- focused native compound/variant suites: 2 files, 9 tests passed
- @tamagui/adapt, button, dialog, popover, popper, sheet, slider, and toast builds: passed
- bun run lint: passed with zero errors (12 existing warnings)
- iOS core tests: 4 files, 26 tests passed
- diff sanity and git diff --check against v3-evolution-plan-baseline: passed (48 intended files only)

The focused web runtime gate used React deduplication in the local shared-dist worktree because untouched dist symlinks point at the original checkout. The repository config and CI checkout are unchanged.


natew:
Superseded by the consolidated v3 integration branch; remaining work will be completed and validated together.

PR #4151, opened 2026-07-28
fix: resolve reported v2 regressions

### LINKED ISSUE #4041: [v2] [TypeScript] Longhand props missing despite onlyShorthandStyleProps: false when extending defaultConfig
### Current Behavior

I am trying to enable longhand style props (like backgroundColor, padding) in my project by explicitly setting onlyShorthandStyleProps: false inside createTamagui.

However, when spreading defaultConfig from @tamagui/config, TypeScript still strips out the longhands at the type level, resulting in an Omit<StackStyleBase, Longhands> error on the components. It seems the type inference from the base config aggressively locks out longhands, even when overridden in the settings object.

### Expected Behavior

Setting onlyShorthandStyleProps: false should successfully override the default configuration's types, restoring the longhand style properties on Tamagui components like <YStack>, <XStack>, and <Text>.

### Tamagui Version

```markdown
2.3.0
```

```markdown
Android
```

### Reproduction

```markdown
import { defaultConfig } from '@tamagui/config/v5';
import { createTamagui } from 'tamagui';

export const config = createTamagui({
   ...defaultConfig,
   settings: {
      ...defaultConfig.settings,
      onlyShorthandStyleProps: false, 
   }
});

export type Conf = typeof config;

declare module 'tamagui' {
  interface TamaguiCustomConfig extends Conf {}
}
```

### System Info

```markdown
System:
    OS: Linux 6.8 Ubuntu 22.04.5 LTS 22.04.5 LTS (Jammy Jellyfish)
    CPU: (12) x64 AMD Ryzen 5 4500 6-Core Processor
    Memory: 7.71 GB / 15.51 GB
    Container: Yes
    Shell: 5.1.16 - /bin/bash
  Binaries:
    Node: 22.22.2 - /home/ybr/.nvm/versions/node/v22.22.2/bin/node
    npm: 11.12.1 - /home/ybr/.nvm/versions/node/v22.22.2/bin/npm
  Browsers:
    Chrome: 147.0.7727.55
  npmPackages:
    @babel/core: ^7.25.2 => 7.29.7 
    @babel/plugin-transform-class-static-block: ^7.28.6 => 7.29.7 
    @babel/plugin-transform-export-namespace-from: ^7.27.1 => 7.29.7 
    @babel/preset-env: ^7.25.3 => 7.29.7 
    @babel/preset-typescript: ^7.27.1 => 7.29.7 
    @babel/runtime: ^7.25.0 => 7.29.7 
    @eslint/js: ^9.14.0 => 9.39.4 
    @fastify/busboy: ^3.2.0 => 3.2.0 
    @gorhom/bottom-sheet: ^5.2.9 => 5.2.14 
    @hookform/resolvers: ^5.2.1 => 5.4.0 
    @mswjs/http-middleware: ^0.10.3 => 0.10.3 
    @notifee/react-native: ^7.8.2 => 7.9.0 
    @react-native-community/cli: 20.1.3 => 20.1.3 
    @react-native-community/cli-platform-android: 20.1.3 => 20.1.3 
    @react-native-community/cli-platform-ios: 20.1.3 => 20.1.3 
    @react-native-community/netinfo: ^12.0.1 => 12.0.1 
    @react-native-vector-icons/fontawesome6: ^13.0.0 => 13.1.2 
    @react-native/babel-preset: 0.84.1 => 0.84.1 
    @react-native/eslint-config: 0.84.1 => 0.84.1 
    @react-native/metro-config: 0.84.1 => 0.84.1 
    @react-native/typescript-config: 0.84.1 => 0.84.1 
    @react-navigation/native: ^7.2.2 => 7.3.3 
    @react-navigation/native-stack: ^7.14.10 => 7.17.5 
    @reduxjs/toolkit: ^2.11.2 => 2.12.0 
    @tamagui/babel-plugin: ^2.3.0 => 2.3.0 
    @tamagui/cli: ^2.3.0 => 2.3.0 
    @tamagui/config: ^2.3.0 => 2.3.0 
    @tamagui/toast: ^2.3.0 => 2.3.0 
    @testing-library/react-native: 13.2.2 => 13.2.2 
    @types/base-64: ^1.0.2 => 1.0.2 
    @types/express: ^5.0.6 => 5.0.6 
    @types/jest: ^30.0.0 => 30.0.0 
    @types/lodash.debounce: ^4.0.9 => 4.0.9 
    @types/node: ^24.10.0 => 24.13.2 
    @types/react: ~19.1.10 => 19.1.17 
    @types/react-native: ^0.73.0 => 0.73.0 
    @types/sanitize-html: ^2.16.0 => 2.16.1 
    async-mutex: ^0.5.0 => 0.5.0 
    babel-plugin-module-resolver: ^5.0.2 => 5.0.3 
    base-64: ^1.0.0 => 1.0.0 
    base64-js: ^1.5.1 => 1.5.1 
    baseline-browser-mapping: ^2.10.0 => 2.10.37 
    detox: ^20.46.0 => 20.51.3 
    eslint: ^9.14.0 => 9.39.4 
    eslint-plugin-react: ^7.37.2 => 7.37.5 
    express: ^5.2.1 => 5.2.1 
    globals: ^15.12.0 => 15.15.0 
    i18next: ^26.0.4 => 26.3.1 
    jest: ^29.7.0 => 29.7.0 
    jwt-decode: ^4.0.0 => 4.0.0 
    lodash.debounce: ^4.0.8 => 4.0.8 
    marked: ^15.0.7 => 15.0.12 
    prettier: 3.3.3 => 3.3.3 
    react: 19.2.3 => 19.2.3 
    react-error-boundary: ^6.1.1 => 6.1.2 
    react-hook-form: ^7.62.0 => 7.79.0 
    react-i18next: ^17.0.2 => 17.0.8 
    react-native: ^0.84.1 => 0.84.1 
    react-native-asset: ^2.2.15 => 2.3.0 
    react-native-bootsplash: ^5.3.0 => 5.5.3 
    react-native-config: ^1.6.1 => 1.6.1 
    react-native-date-picker: ^5.0.13 => 5.0.13 
    react-native-element-dropdown: ^2.12.4 => 2.12.4 
    react-native-fs: ^2.20.0 => 2.20.0 
    react-native-gesture-handler: ^2.31.0 => 2.32.0 
    react-native-get-location: ^6.0.1 => 6.0.1 
    react-native-get-random-values: ^2.0.0 => 2.0.0 
    react-native-image-picker: ^8.2.1 => 8.2.1 
    react-native-json-tree: ^1.5.0 => 1.5.0 
    react-native-keyboard-controller: ^1.21.5 => 1.21.11 
    react-native-keychain: ^10.0.0 => 10.0.0 
    react-native-launch-arguments: ^4.1.1 => 4.1.1 
    react-native-marked: ^8.0.1 => 8.1.0 
    react-native-mmkv: ^4.3.0 => 4.3.1 
    react-native-nitro-modules: ^0.35.3 => 0.35.9 
    react-native-reanimated: ^4.3.0 => 4.4.1 
    react-native-reanimated-carousel: ^4.0.3 => 4.0.3 
    react-native-safe-area-context: ^5.7.0 => 5.8.0 
    react-native-screens: 4.24.0 => 4.24.0 
    react-native-signature-canvas: ^5.0.2 => 5.0.2 
    react-native-sound: 0.13.0 => 0.13.0 
    react-native-toast-message: ^2.3.3 => 2.3.3 
    react-native-url-polyfill: ^3.0.0 => 3.0.0 
    react-native-vision-camera: 4.7.3 => 4.7.3 
    react-native-webview: ^13.16.1 => 13.16.1 
    react-native-worklets: ^0.9.2 => 0.9.2 
    react-redux: ^9.1.2 => 9.3.0 
    react-test-renderer: 19.2.3 => 19.2.3 
    redux-persist: ^6.0.0 => 6.0.0 
    sanitize-html: ^2.14.0 => 2.17.5 
    tamagui: ^2.3.0 => 2.3.0 
    ts-node: ^10.9.2 => 10.9.2 
    tsconfig-paths: ^4.2.0 => 4.2.0 
    typescript: ~5.9.2 => 5.9.3 
    typescript-eslint: ^8.46.2 => 8.61.1 
    uuid: ^11.1.0 => 11.1.1 
    wait-on: ^9.0.4 => 9.0.10 
    zod: ^4.0.17 => 4.4.3
```

### LINKED ISSUE #4074: RadioGroup.Indicator leaks `pressTheme` to the DOM on web (v2) — React unknown-prop console error

### LINKED ISSUE #4133: [Compiler/Native] Media query styles with dynamic values are applied unconditionally after extraction
## Environment

- `tamagui`: 2.3.3
- `@tamagui/static` / `@tamagui/babel-plugin`: 2.3.3
- `react-native`: 0.85.3
- `expo`: 56.0.12
- Platform: native (iOS/Android), `@tamagui/babel-plugin` with `disableExtraction: false`

## Summary

On the native compiler target, when a media-query style prop contains a **dynamic value** (e.g. a ternary that references a runtime variable), the extractor emits the media key as a bare string inside a logical expression. At runtime the media condition is never resolved against `useMedia()`, and since a non-empty string is truthy, the media styles are applied **unconditionally on every screen size**.

Media props with fully static values are unaffected. Web extraction is unaffected.

## Reproduction

```tsx
import { YStack } from 'tamagui'

export function Screen({ fillContainer = false }: { fillContainer?: boolean }) {
  return (
    <YStack
      width="100%"
      $gtLg={{
        width: fillContainer ? '100%' : '50%',
        pb: '$4',
      }}
    />
  )
}
```

Expected on a phone (< `gtLg` breakpoint): `width: 100%`, no `pb`.
Actual with extraction enabled: `width: 50%` and `pb` are applied. With `disableExtraction: true` the behavior is correct.

## Compiled output (native target)

```js
jsxs(_ReactNativeViewStyled0, {
  _expressions: [
    !!("gtLg" && fillContainer),   // ← media key is a bare truthy string
    !!("gtLg" && !fillContainer),  // ← always true when fillContainer is false
    "gtLg"                         // ← static media style: this one works
  ],
  ...
})

var _ReactNativeViewStyled0 = _withStableStyle(
  __ReactNativeView,
  (theme, _expressions) => [
    _sheet['1'],                            // base styles
    _expressions[0] ? _sheet['2'] : null,   // width '100%'
    _expressions[1] ? _sheet['3'] : null,   // width '50%'  ← applied on every screen
    _expressions[2] ? _sheet['4'] : null,   // pb           ← resolved correctly via media
  ],
  false,
  true
)
```

## Root cause

`_withStableStyle` (`@tamagui/web/src/_withStableStyle.native.tsx`) only resolves media keys when the whole expression entry is a **plain string**:

```js
var resolvedExpressions = media
  ? _expressions.map((expr) => (typeof expr === 'string' ? media[expr] : expr))
  : _expressions
```

That works for static media styles, whose test is emitted as a bare `"gtLg"` string literal.

But when the media object contains a dynamic value, `createExtractor` produces a ternary attr whose test is a logical expression embedding the media key as a string literal, e.g. `"gtLg" && fillContainer`. Then in `extractToNative.ts`, `safeExpressions` only preserves bare string literals and wraps everything else in `!!`:

```js
const safeExpressions = expressions.map((expr) =>
  t.isStringLiteral(expr) ? expr : t.unaryExpression('!', t.unaryExpression('!', expr))
)
```

So the emitted expression is `!!("gtLg" && fillContainer)`. `"gtLg"` is just a truthy string in plain JS, the `media[...]` lookup never happens, and the expression reduces to `!!fillContainer`. The media condition is silently dropped and the `$gtLg` styles leak to all screen sizes.

## Suggested fix directions

- In `extractToNative`, when a ternary test contains a media-key string literal, don't collapse it with `!!`. Either:
  - emit the test in a shape `_withStableStyle` can resolve (e.g. keep `["gtLg", <boolExpr>]` tuples and resolve `media[key] && expr` at runtime), or
  - bail out of extracting that media prop (leave it as a runtime prop / skip flattening for that element), matching the current behavior of static-value-only detection.
- Alternatively, teach `_withStableStyle` to resolve media keys nested inside expression entries, though that requires changing what the compiler emits anyway.

### LINKED ISSUE #4148: [Bug] Tamagui is mutating nested properties of a shared-value object inside a worklet.
### Current Behavior

Root Cause: `@tamagui/animations-reanimated` v2.5.1

The `emitted` key comes from **line 1056** of `node_modules/@tamagui/animations-reanimated/dist/cjs/createAnimations.native.js`:

```js
var mapperStateRef = useSharedValue({
  emitted: {}            // ← line 1056
});
```

This is inside Tamagui's `useAnimations` hook — the core animation engine used by every Tamagui styled component (`YStack`, `XStack`, `Button`, etc.). The worklet at line 1058 does:

```js
var mapperState = mapperStateRef.value;    // read
mapperState.emitted = {};                   // mutate (line 1064, when disabled/hydrating)
// ... animate keys ...
mapperState.emitted = emitted;              // mutate (line 1141, final state)
```

## Why It Started Recently

This is **not** a Tamagui version change (still `2.5.1`), but the interaction with the **React Navigation 8 + React 19 `<Activity>`** changes:

1. **RN8's default `inactiveBehavior: 'pause'`** wraps background screens in `<Activity mode="hidden">`
2. On re-activation, Tamagui's layout effects re-fire (`useIsomorphicLayoutEffect` at lines 923-952), setting `renderSnapshotRef.value` to a new object
3. This triggers the `useAnimatedStyle` worklet to re-evaluate
4. **Reanimated 4.x** (`~4.5.3`) now detects that `mapperState.emitted = emitted` (inside the worklet, line 1141) is modifying an object that was derived from `mapperStateRef.value` — the shared value object that was already serialized/frozen for the UI thread

The warning was always latent in Tamagui's code (mutating nested properties of a shared-value object inside a worklet is an anti-pattern). What changed is Reanimated 4.x's stricter enforcement of the "don't mutate worklet-bound objects" rule.

### Expected Behavior

You see no errors from Reanimated

### Tamagui Version

```markdown
v2.5.1
```

```markdown
iOS & Android
```

### Reproduction

```markdown
Using React Navigation v8
+ Latest Reanimated (that automatically has stricter checks)
+ Tamagui v2

You get

` WARN  [Worklets] Tried to modify key `emitted` of an object which has been already passed to a worklet. `


Stack Trace:


 WARN  [Worklets] Tried to modify key `emitted` of an object which has been already passed to a worklet. See
        https://docs.swmansion.com/react-native-reanimated/docs/guides/troubleshooting#tried-to-modify-key-of-an-object-which-has-been-converted-to-a-serializable
        for more details.

Call Stack
  warn (node_modules\react-native-worklets\src\debug\logger.ts)
  Object.defineProperty$argument_2.set (node_modules\react-native-worklets\src\memory\serializable.native.ts)
  useAnimatedStyle$argument_0 (node_modules\@tamagui\animations-reanimated\dist\esm\createAnimations.native.js)
  initialUpdaterRun (node_modules\react-native-reanimated\src\animation\util.ts)
  styles.map$argument_0 (node_modules\react-native-reanimated\src\createAnimatedComponent\PropsFilter.tsx)
  map (<native>)
  filterNonAnimatedProps (node_modules\react-native-reanimated\src\createAnimatedComponent\PropsFilter.tsx)
  filterNonAnimatedProps (node_modules\react-native-reanimated\src\createAnimatedComponent\PropsFilter.tsx)
  render (node_modules\react-native-reanimated\src\createAnimatedComponent\AnimatedComponent.tsx)
  callRender.react_stack_bottom_frame (node_modules\react-native\Libraries\Renderer\implementations\ReactFabric-dev.js)
  finishClassComponent (node_modules\react-native\Libraries\Renderer\implementations\ReactFabric-dev.js)
  updateClassComponent (node_modules\react-native\Libraries\Renderer\implementations\ReactFabric-dev.js)
  beginWork (node_modules\react-native\Libraries\Renderer\implementations\ReactFabric-dev.js)
  runWithFiberInDEV (node_modules\react-native\Libraries\Renderer\implementations\ReactFabric-dev.js)
  performUnitOfWork (node_modules\react-native\Libraries\Renderer\implementations\ReactFabric-dev.js)
  workLoopSync (node_modules\react-native\Libraries\Renderer\implementations\ReactFabric-dev.js)
  renderRootSync (node_modules\react-native\Libraries\Renderer\implementations\ReactFabric-dev.js)
  performWorkOnRoot (node_modules\react-native\Libraries\Renderer\implementations\ReactFabric-dev.js)
  performSyncWorkOnRoot (node_modules\react-native\Libraries\Renderer\implementations\ReactFabric-dev.js)
  flushSyncWorkAcrossRoots_impl (node_modules\react-native\Libraries\Renderer\implementations\ReactFabric-dev.js)
  processRootScheduleInMicrotask (node_modules\react-native\Libraries\Renderer\implementations\ReactFabric-dev.js)
  scheduleMicrotask$argument_0 (node_modules\react-native\Libraries\Renderer\implementations\ReactFabric-dev.js)



This is now crystal clear. The stack trace shows:


initialUpdaterRun (react-native-reanimated/src/animation/util.ts)
useAnimatedStyle$argument_0 (@tamagui/animations-reanimated/dist/esm/createAnimations.native.js)


## The Root Cause

Reanimated runs `initialUpdaterRun` on the **JS thread** during the first render. Tamagui's worklet at lines 1018-1102 does this:


var mapperState = mapperStateRef.value;   // line 1021 — gets the frozen object
// ...
mapperState.emitted = {};                  // line 1024 — ❌ mutates frozen nested property
// ...
mapperState.emitted = emitted;             // line 1101 — ❌ mutates frozen nested property


When `useSharedValue({ emitted: {} })` is created, Reanimated serializes/freezes the object. The `initialUpdaterRun` then runs the worklet body on the JS thread, where `mapperState.emitted = ...` tries to write to the frozen `.emitted` setter → warning.

After the initial run, the worklet runs on the UI thread where the object isn't frozen, so subsequent runs don't warn. That's why this is `WARN`-only and animations still work — it fires once per animated component mount.

## The Fix

Replace nested mutation (`mapperState.emitted = x`) with full-value replacement (`mapperStateRef.value = { emitted: x }`), which is Reanimated's correct API for updating shared values:

**In `dist/esm/createAnimations.native.js`:**

| Line | Before | After |
|---|---|---|
| 1024 | `mapperState.emitted = {};` | `mapperStateRef.value = { emitted: {} };` |
| 1101 | `mapperState.emitted = emitted;` | `mapperStateRef.value = { emitted: emitted };` |

**In `dist/cjs/createAnimations.native.js`:**

| Line | Before | After |
|---|---|---|
| 1064 | `mapperState.emitted = {};` | `mapperStateRef.value = { emitted: {} };` |
| 1141 | `mapperState.emitted = emitted;` | `mapperStateRef.value = { emitted: emitted };` |

This is behavior-preserving: `mapperState = mapperStateRef.value` still reads the current value for `previouslyEmitted` (line 1027), and the replacement at line 1101 ensures the next worklet invocation sees the updated `emitted` map.
```

### System Info

```markdown
Expo: SDK 55
React Native: 0.83.10
React: 19.2.8
Tamagui: Latest
Reanimated: Latest
```

### LINKED ISSUE #4149: Compiler drops an element's other static style props when any style prop is a variable (regression: 2.4.0 → 2.5.1)
### Current Behavior

When a Tamagui `View`/`Stack` has **any single style prop whose value is a variable reference** (e.g. `borderColor={SOME_CONST}`, `bg={SOME_CONST}`, `shadowColor={SOME_CONST}`), the compiler **drops all the element's OTHER static style props** during flattening/extraction. Only the variable prop is applied; `backgroundColor` (even a literal `bg="white"`), `borderRadius`, `margin`, `padding`, `alignItems`, `justifyContent`, etc. silently disappear.

### Expected Behavior

A variable-valued style prop should be applied without affecting the element's other static style props (all props render, exactly as on 2.4.0).

### Tamagui Version

**2.5.1** — broken. Last known-good: **2.4.0** (also broke on 2.4.1–2.4.5 in a real app). This is a regression between 2.4.0 and 2.5.1.

Reproduced on **Web (Metro)** and **iOS (native)** — identical broken output on both, consistent with the bug being in the compiler (build-time), not the renderer.

### Reproduction

Minimal repro repo (Expo 57 + Metro): **https://github.com/js2kpb/tamagui-bg-repro**

```
npm install
npx expo start --web   # or: npx expo run:ios
```

Each example is a **left = literal (correct)** vs **right = variable (broken)** pair.

**Minimal case** — the trigger prop (`borderColor`) is a variable; the *other* props (incl. a literal white `bg`) are dropped:

const VAR = "#4a90d9"; // module-level const

// A) all literal — everything applies
<View bg="white" borderWidth={4} borderColor="#4a90d9" rounded={20} m={16} p={16}><Text>box</Text></View>

// B) borderColor is a variable — BUG
<View bg="white" borderWidth={4} borderColor={VAR}      rounded={20} m={16} p={16}><Text>box</Text></View>

`getComputedStyle`:

| case | background | border-radius | margin | padding | border |
|------|------------|---------------|--------|---------|--------|
| A all literal | `white` ✅ | `20px` ✅ | `16px` ✅ | `16px` ✅ | blue 4px ✅ |
| B `borderColor={VAR}` | **`transparent`** ❌ | **`0px`** ❌ | **`0px`** ❌ | **`0px`** ❌ | blue 4px |

The trigger can be **any** style prop with a variable value (`bg`, `borderColor`, `shadowColor`, …); all the element's *other* static props are dropped.

**Real-world patterns** (the repo's `App.tsx` reproduces each; these are common shapes that all break):

1. Status chip — `<XStack bg={config.bgColor} px={8} py={4} rounded={6}>` → padding + rounding dropped (chip becomes a flush rectangle).
2. Icon box — `<View bg={bgColor} width={48} height={48} rounded={24} items="center" justify="center">` → square instead of circle, child pinned top-left.
3. Absolute badge — `<View position="absolute" t={8} l={8} px={10} py={4} rounded={100} bg={COLORS.badgeBg}>` → padding/rounding/inset dropped, badge sticks to the corner.
4. List row — `<XStack bg="white" px={20} py={16} rounded={12} borderWidth={1} borderColor={COLORS.borderColor} items="center">` → here `bg` is a **literal** `"white"` and only `borderColor` is a variable; the white background itself drops to **transparent**, plus rounding/padding are lost.

Because deriving a color from a constant / config object / status map is extremely common, this manifests app-wide as "rounded corners not applying, margins/padding gone, white backgrounds turning transparent, absolutely-positioned children mis-placed".

**Confirming it's a regression:** in the repo, set `tamagui`, `@tamagui/babel-plugin`, `@tamagui/config` to `2.4.0` in `package.json`, `npm install`, re-run → every example renders correctly. Switch back to `2.5.1` → broken.

### Screenshots (tamagui 2.5.1)

Left = literal bg (correct) · Right = variable bg (broken). The same code renders correctly on 2.4.0.

**Web (Metro):**

web

**iOS (native):**  *(the black shape at the top is the iOS Dynamic Island, not part of the repro)*

ios

### System Info

```markdown
System:
  OS: macOS 26.5.2
  CPU: (12) arm64 Apple M2 Pro
  Shell: 5.9 - /bin/zsh
Binaries:
  Node: 24.14.1
  npm: 11.18.0
  Watchman: 2026.06.01.00
Browsers:
  Chrome: 150.0.7871.129
  Safari: 26.5.2
npmPackages:
  @tamagui/babel-plugin: 2.5.1 => 2.5.1
  @tamagui/config: 2.5.1 => 2.5.1
  expo: ~57.0.8 => 57.0.8
  react: 19.2.3 => 19.2.3
  react-native: 0.86.0 => 0.86.0
  react-native-web: ^0.21.2 => 0.21.2
  tamagui: 2.5.1 => 2.5.1
```

natew:
## Summary

- avoid Reanimated shared-object writes during native JS-thread initial renders without notifying the mapper from its own UI-runtime state update
- preserve runtime conditions inside native media props and route even a single media-only `View` through the media-aware wrapper, preventing the media style from applying at every screen size
- remove unsafe native partial flattening that emitted shorthand keys inside a React Native style object and let the surviving static styles override dynamic or pseudo props
- remove the stale RadioGroup indicator `pressTheme` DOM prop and the unexported v1 Button source that contained the other stale copies
- document the correct v5 `onlyAllowShorthands` override for restoring longhand props
- run the compiler, core, and component native unit suites in CI, with native codegen components represented by host test components

## Validation

- `bun run build` (162 packages)
- `@tamagui/animations-reanimated` build
- `@tamagui/button` build
- native CI command: 3/3 packages passed
  - compiler native tests: 30 passed
  - core native tests: 95 passed, 7 expected failures, 11 skipped
  - component native tests: 23 passed
- core web tests: 185 passed, 1 skipped, 1 todo
- component web tests: 30 passed
- iOS Maestro Select and Toast flows
- iOS Detox: 4/4 shards
- root typecheck
- root lint: 0 errors, 1 pre-existing warning
- workflow validation with `actionlint`: only pre-existing shellcheck findings
- Prettier and `git diff --check`

Fixes #4148
Fixes #4149
Fixes #4133
Fixes #4074
Fixes #4041


OwenMelbz:
Do we know when this will be released on npm? Thanks 🙏

natew:
@OwenMelbz in a few minutes

PR #4157, opened 2026-08-06
fix(animations-reanimated): don't wrap onStart when the initial updater returns a plain value

wootencl:
### Bug

  During reanimated's initial updater run, `withTiming`/`withSpring` return the plain starting value instead of an animation descriptor
  ([`defineAnimation`](https://github.com/software-mansion/react-native-reanimated/blob/4.2.2/packages/react-native-reanimated/src/animation/util.ts): `IN_STYLE_UPDATER` returns `starting`). `applyAnimation` still assigns the `onStart`
  seed wrapper onto it, and color keys enter that branch unconditionally (`validateStartAsColor`). When the mapper re-initializes after a color key has already emitted, the assignment lands on a raw color string:

  ```
  TypeError: Cannot create property 'onStart' on string 'rgba(0, 0, 0, 0)'
  ```

  This throws during render, so React 19 unmounts the root (full app teardown). In expo-router apps the first visible error is often the downstream "Attempted to navigate before mounting the Root Layout" from the next navigation call.

  Hit on 2.7.2 (reanimated 4.5.1, RN 0.86.2 new arch, Expo SDK 57) when a mutation response re-rendered a screen using `transition` with color keys resting on `transparent`. Present since 2.5.0, 2.4.x unaffected.

  ### Fix

  Skip the `onStart` wrapper unless `animatedValue` is actually a descriptor. When the updater short-circuits nothing starts, so there is no seed to substitute.

  ### Test

  `ReanimatedInitialUpdater.native.test.tsx` (follows the `ReanimatedMapperState` pattern): mocks `withTiming`/`withSpring` to return the plain starting value, renders an animated color, re-renders with a new one, asserts no throw.
  Reproduces the exact TypeError before the fix, green after. All existing native and web component tests pass.

_Credit: Claude Code_


natew:
Thanks @wootencl — fixed on main in `25d97e1ab6` + `a756be9a5f`, and your version had the shape that survived.

#4196 proposed the same guard as a positive `typeof === 'object'` check, which is what landed first and broke all native animations: SWC's ES5 transform turns that into a `_type_of()` helper that isn't a worklet, so Reanimated 4's UI runtime throws `Tried to synchronously call a Remote Function`. Your comment called that hazard out ahead of time:

```ts
// checked as not-string/not-number: comparing typeof against 'object' compiles
// to a _type_of helper that is a remote function on the UI runtime
typeof animatedValue !== 'string' && typeof animatedValue !== 'number'
```

Main is that, widened to also cover `boolean`/`undefined`/`null`.

Your test file was the only regression coverage for this anywhere — the fix landed with none — so I've salvaged it in #4201 with you as co-author. I tightened two assertions: it now requires zero uncaught exceptions rather than the absence of one message, and pins that the tree actually rendered so a no-op render can't pass. Closing this in favor of that.

PR #4175, opened 2026-08-22
fix(toast): map accessibilityLabel to aria-label on web

### LINKED ISSUE #4160: [toast v2] Toaster leaks 'accessibilityLabel' onto a DOM element — React warning on every toast (web)
### Current Behavior

Triggering any toast via the v2 API (`toast('hi')` with `<Toaster />`) on web logs this React warning on every toast:

```
React does not recognize the `accessibilityLabel` prop on a DOM element. If you
intentionally want it to appear in the DOM as a custom attribute, spell it as
lowercase `accessibilitylabel` instead.
```

This happens even when the app never passes `accessibilityLabel` anywhere.

### Expected Behavior

No warning. On web the prop should either be dropped or mapped to `aria-label`.

### Root cause

In `@tamagui/toast/src/ToastComposable.tsx` (line ~1166), `ToastItem` unconditionally forwards the prop to `ToastPositionWrapper` (a styled `YStack`, i.e. a `div` on web):

```tsx
<ToastPositionWrapper
  ref={ref}
  testID={rest.testID}
  accessibilityLabel={rest.accessibilityLabel}   // <-- always present, usually undefined
```

Two things combine to surface it:

1. Tamagui v2's web pipeline no longer maps RN `accessibility*` props to ARIA — `getSplitStyles` only special-cases `disabled` / `testID` / `id`, and `createComponent` only reads `accessibilityLabel` for the press-debug name. So the key passes straight through to the DOM element.
2. React 19 warns based on the *presence* of the camelCased key, even when the value is `undefined` (verified on react-dom 19.2.8 — `renderToString(<div accessibilityLabel={undefined} />)` warns).

`testID` on the same line is fine because `getSplitStyles` maps it to `data-testid`.

### Suggested fix

Pluck `accessibilityLabel` out of the props destructure (it's also spread onto `ToastItemFrame` via `{...rest}`) and only emit it when defined — mapped to `aria-label` on web:

```tsx
const { toast, index, children, accessibilityLabel, ...rest } = props
...
<ToastPositionWrapper
  ref={ref}
  testID={rest.testID}
  {...(accessibilityLabel != null &&
    (isWeb ? { 'aria-label': accessibilityLabel } : { accessibilityLabel }))}
```

We're running this as a pnpm patch in production and it resolves the warning with no behavior change (native keeps the RN prop, where it's valid).

### Tamagui Version

2.7.4 (`@tamagui/toast` 2.7.4, latest at time of writing)

Web (Next.js 16.3.0, webpack, react / react-dom 19.2.8). Native is unaffected — `.native.js` builds keep the prop, which RN accepts.

### Reproduction

Any web app with the v2 Toaster — no custom props needed:

```tsx
import { Toaster, toast } from '@tamagui/toast/v2'

// in root providers:
<Toaster position="bottom-center" />

// anywhere:
toast('hello')  // warning appears in the console on every toast
```

Happy to push a commit to a starter-free fork if needed, but the trace above pinpoints the exact line.

### System Info

```
System:
  OS: macOS 26.5.2
  CPU: (12) arm64 Apple M4 Pro
Binaries:
  Node: 24.3.0
  pnpm: 11.20.0 (node-linker=hoisted)
npmPackages:
  tamagui: 2.7.4
  @tamagui/toast: 2.7.4
  react: 19.2.8
  next: 16.3.0
```

dennytosp:
Fixes #4160

### Problem

Every toast logs a React warning on web:

```
React does not recognize the `accessibilityLabel` prop on a DOM element.
```

Two things combine:

1. v2 dropped the RN `accessibility*` → ARIA mapping (see the note in `core/web/src/types.tsx`) — `getSplitStyles` only special-cases `disabled` / `testID` / `id`, and `createComponent` reads `accessibilityLabel` purely for the press-debug name. So the key passes straight to the DOM element.
2. React 19 warns on the **presence** of the camelCased key, even when the value is `undefined`.

`ToastItem` forwarded it unconditionally:

```tsx
const { toast, index, children, ...rest } = props
...
<ToastPositionWrapper testID={rest.testID} accessibilityLabel={rest.accessibilityLabel}
```

so the key was always there, and it also rode along in `{...rest}` onto `ToastItemFrame`. `testID` on the same line is fine because `getSplitStyles` maps it to `data-testid`.

### Fix

Pull `accessibilityLabel` out of the destructure and only emit it when it is actually defined — as `aria-label` on web, as the RN prop on native:

```tsx
const { toast, index, children, accessibilityLabel, ...rest } = props
...
{...(accessibilityLabel != null &&
  (isWeb ? { 'aria-label': accessibilityLabel } : { accessibilityLabel }))}
```

Pulling it out of `rest` also stops the second copy from landing on `ToastItemFrame`, so the label now sits once on the outer wrapper instead of on two nested elements.

### Test

`code/ui/components-test/Toast.web.test.tsx` renders the v2 `Toaster`, fires a toast, and asserts no `[accessibilitylabel]` attribute reaches the DOM and that React logged no `accessibilityLabel` warning.

Verified red before the change (`expected true to be false` on the warning assertion).

### Checks

- `bun run typecheck` ✅
- `bun run lint` ✅
- `@tamagui/components-test` web suite: 6 files / 31 tests ✅

### DIFF
--- a/code/ui/components-test/Toast.web.test.tsx
+++ b/code/ui/components-test/Toast.web.test.tsx
@@ -1,0 +1,42 @@
+import '@testing-library/jest-dom'
+import { getDefaultTamaguiConfig } from '@tamagui/config-default'
+import { TamaguiProvider, createTamagui } from '@tamagui/core'
+import { PortalProvider } from '@tamagui/portal'
+import { toast, Toaster } from '@tamagui/toast/v2'
+import { act, render } from '@testing-library/react'
+import { afterEach, describe, expect, test, vi } from 'vitest'
+
+const conf = createTamagui(getDefaultTamaguiConfig())
+
+afterEach(() => {
+  toast.dismiss()
+  vi.restoreAllMocks()
+})
+
+function renderToaster() {
+  return render(
+    <TamaguiProvider config={conf} defaultTheme="light">
+      <PortalProvider shouldAddRootHost>
+        <Toaster duration={Number.POSITIVE_INFINITY} />
+      </PortalProvider>
+    </TamaguiProvider>
+  )
+}
+
+describe('Toast v2 web accessibility props', () => {
+  test('does not forward accessibilityLabel to the DOM', async () => {
+    // react warns on the presence of the camelCased key, value or not
+    const errorSpy = vi.spyOn(console, 'error').mockImplementation(() => {})
+
+    const { container } = renderToaster()
+    await act(async () => {
+      toast('hello')
+    })
+
+    expect(container.querySelector('[accessibilitylabel]')).toBeNull()
+    const warned = errorSpy.mock.calls.some((args) =>
+      args.some((arg) => typeof arg === 'string' && arg.includes('accessibilityLabel'))
+    )
+    expect(warned).toBe(false)
+  })
+})
--- a/code/ui/toast/src/ToastComposable.tsx
+++ b/code/ui/toast/src/ToastComposable.tsx
@@ -901,7 +901,7 @@ export interface ToastItemProps extends GetProps<typeof ToastItemFrame> {
 
 const ToastItemInner = ToastItemFrame.styleable<ToastItemProps>(
   function ToastItem(props, ref) {
-    const { toast, index, children, ...rest } = props
+    const { toast, index, children, accessibilityLabel, ...rest } = props
     const ctx = useToastContext('Toast.Item')
 
     const [mounted, setMounted] = React.useState(false)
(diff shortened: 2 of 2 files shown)


dennytosp:
Closing in favour of #4169, which predates this by six days and is the better fix.

I missed the open PR before pushing — my apologies for the noise.

Beyond the duplication, @jhhayashi is right that mapping to `aria-label` on `ToastPositionWrapper` is the wrong call, and I verified the hazard against main: the Escape handler at `ToastComposable.tsx:1216` does

```tsx
const container = current.closest('[aria-label]')
```

walking up from the toast frame to find the **viewport**, which carries `aria-label={`${label} ${hotkeyLabel}`}` at line 596. `ToastPositionWrapper` sits between the two, so labelling it captures that lookup one level too low — `querySelectorAll('[tabindex="0"]')` then sees only the toast being dismissed and focus never moves to the next one. That only bites when a caller actually passes `accessibilityLabel`, but it is a real regression that #4169 avoids by simply dropping the line.

I did write an automated regression test for this while working on it, which #4169 doesn't have — it renders the v2 `Toaster`, fires a toast, and asserts both that no `[accessibilitylabel]` attribute reaches the DOM and that React logged no `accessibilityLabel` warning. It's red on main and green with either fix. Happy to hand it over if it's useful — just say the word and I'll open it against #4169 or as a follow-up.

PR #4178, opened 2026-08-22
fix(slider): use viewport-relative pointer coords on web

### LINKED ISSUE #4146: Vertical Slider not following cursor
### Current Behavior

Vertical slider's thumb is not following cursor when there's some scroll-top distance.

https://github.com/user-attachments/assets/58c290e8-be1f-4ad3-88c7-415398feb707

### Expected Behavior

the thumb should follow cursor, just as horizontal slider does.

### Tamagui Version

```markdown
latest as docs site
```

```markdown
Web (Firefox and Chrome)
```

### Reproduction

```markdown
go to https://tamagui.dev/ui/slider, drag and move the vertical slider; then scroll down some distance, do the same again.
```

### System Info

```markdown

```

dennytosp:
Fixes #4146

### Problem

Dragging a vertical slider stops tracking the cursor once the page is scrolled — reproducible on https://tamagui.dev/ui/slider by scrolling down before dragging.

The two sides of the pointer math use different origins:

- `Slider.tsx` measures the track through `measure()`, which on web resolves to `getBoundingClientRect()` — `use-element-layout` returns `pageY: a.top` (`getRelativeDimensions`, line ~353), i.e. **viewport-relative**.
- The responder event's coordinates come straight off the DOM event — `createResponderEvent.ts` line ~74, `const pageY = domEventChangedTouches?.[0].pageY || domEvent.pageY` — i.e. **document-relative**.

So `event.nativeEvent.pageY - state.offset` is off by exactly `window.scrollY`. With a tall page the extrapolated value saturates and the slider pins to its minimum.

Native is unaffected: there `measure` reports against the screen and so do the touch coordinates.

Two details that explain the reported shape of the bug:

- **The initial press is fine.** `onSlideStart` uses `locationY`, which `createResponderEvent` computes as `clientY - currentTarget.getBoundingClientRect().top` — both viewport-relative. Only `onSlideMove` / `onSlideEnd` mix origins.
- **Horizontal looks fine** only because `window.scrollX` is normally 0. It has the same latent defect.

### Fix

Put the pointer back on the same origin as the measured track, on web only:

```ts
const toTrackX = (pageX: number) => (isWeb ? pageX - window.scrollX : pageX)
const toTrackY = (pageY: number) => (isWeb ? pageY - window.scrollY : pageY)
```

applied to move/end in both orientations. `pageY - scrollY` is `clientY`, so it stays correct for a slider inside a scrollable container too, and it re-reads the scroll on every event rather than depending on a re-measure.

I did consider switching move/end to `locationY` like `onSlideStart` already does, which would delete the `state.offset` math entirely — but that changes the native drag path too, and this bug is web-only, so I kept the change scoped.

### Test

`code/kitchen-sink/tests/SliderScrollOffset.test.tsx` + `SliderScrollOffsetCase` usecase: a tall spacer forces a real scroll, then the test drags (not just clicks — a click passes even unfixed) and asserts the value matches the cursor position, for both orientations.

Verified red on main:

```
Expected: > 71
Received:   0
```

The horizontal case passes on main as well; it's there as a guard that `toTrackX` didn't change working behaviour.

### Checks

- `bun run typecheck` ✅
- `bun run lint` ✅
- kitchen-sink Playwright `default` project, the one this change gates in CI: **584 passed, 13 skipped, 0 failed**


natew:
Reviewed. The origin mismatch you found is real and correctly explained — `measure()` resolves to `getBoundingClientRect()` (viewport-relative) while `pageY` off the DOM event is document-relative. The catch that the initial press works because `onSlideStart` uses `locationY` is the key observation, and I think it points at a better fix than the one here.

### What this fix leaves open

`toTrackY` makes the pointer side fresh and viewport-relative, but `state.offset` on the other side of the subtraction is a **cached** viewport-relative value, and nothing re-measures it on scroll. The only triggers are:

```ts
useOnDebouncedWindowResize(measure)                                      // resize only
const io = new IntersectionObserver(..., { threshold: [0, 0.5, 1.0] })   // debounced 200ms
setInterval(() => activeSliderMeasureListeners.forEach(cb => cb()), 1000)
```

So scroll a **fully visible** slider by 200px — no threshold crossing, no resize — then drag within the next second, and the thumb drifts by that 200px again. Your test scrolls the slider *into* view, which crosses an IntersectionObserver threshold and forces a re-measure, so it doesn't exercise that window.

### The cause-level version

You considered `locationY` and rejected it for touching the native path. Scoping it to web gets the benefit without that risk:

```ts
onSlideMove={(event) => {
  const value = getValueFromPointer(
    process.env.TAMAGUI_TARGET === 'web'
      ? event.nativeEvent.locationY
      : event.nativeEvent.pageY - state.offset
  )
```

`locationY` is computed per event against the live rect:

```ts
function locationY(y) {
  rect = rect || getBoundingClientRect(responderEvent.currentTarget)
  return y - rect.top
}
```

which is `pageY - scrollY - trackTop` read fresh. That makes all three handlers consistent with `onSlideStart`, deletes the `- state.offset` math on web, needs no `window.scrollX` reference (so it also survives nested scroll containers and iframes), and closes the stale-offset window entirely.

Your test case and both-orientation coverage carry over unchanged — please keep them. If you add a second drag after a small scroll that doesn't cross an IntersectionObserver threshold, that new assertion should be red against the `toTrackY` version and green against this one.

natew:
Landed in ec4bbcedc2 with you as co-author, thanks @dennytosp. Same root cause you identified: `pageX/pageY` are document-relative while the track is measured with `getBoundingClientRect`. Web now uses `locationX/locationY`, which are computed against a live rect on every event, so there's no offset to subtract and no stale-offset window either (nothing re-measures on scroll). Native keeps the offset since it measures both against the screen. Three Playwright tests, including one for the stale-offset case specifically.

PR #4179, opened 2026-08-22
fix(core): deliver the host to a swapped forwarded ref

### LINKED ISSUE #4031: Cached composed ref: forwarded ref only invoked once, breaks react-hook-form focus after reset
### Current Behavior

The `Input` component doesn't provide its host input node to a forwarded ref after a `react-hook-form` reset re-render. This manifests with RHF no longer having the ref to the element, so focus APIs such as `setFocus()` and `shouldFocusError` are not able to focus on form inputs.

I hit this on:

- `tamagui@2.1.0`
- `react@19.2.3`
- `react-dom@19.2.3`
- `react-native-web@0.21.2`
- `react-hook-form@7.72.0`

The bug still reproduces on `tamagui@2.2.0`.

### Expected Behavior
Control case with a plain `<input>` works

Using the same RHF `Controller` setup with a native input:

```tsx
<input {...field} ref={field.ref} value={field.value ?? ''} />
```

RHF keeps a focusable ref after `reset()`

`Input` should re-deliver the host input node when the forwarded ref identity changes, or otherwise compose refs in a way that does not permanently close over the initial `forwardedRef`.

That would allow RHF to retain the focusable field ref after `reset()`, so the behavior of `setFocus()` and `shouldFocusError` continues to work.

### Tamagui Version
```
v2.2.0
```

```
Web, didn't check iOS or Android
```
### Reproduction

```
https://github.com/hackily/tamagui-starter-free/tree/repro/rhf-reset-ref-tamagui-input
```

### System Info
```
System:
    OS: macOS 15.7
    CPU: (12) arm64 Apple M2 Pro
    Memory: 136.13 MB / 32.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 24.11.0 - /Users/hackily/.nvm/versions/node/v24.11.0/bin/node
    Yarn: 4.5.0 - /opt/homebrew/bin/yarn
    npm: 11.6.1 - /Users/hackily/.nvm/versions/node/v24.11.0/bin/npm
    Watchman: 2025.03.10.00 - /opt/homebrew/bin/watchman
  Browsers:
    Chrome: 149.0.7827.103
    Firefox: 151.0.3
    Safari: 18.6
  npmPackages:
    @babel/runtime: ^7.24.6 => 7.26.9 
    @biomejs/biome: ^1.9.3 => 1.9.4 
    @tamagui/cli: ^2.2.0 => 2.2.0 
    @tamagui/react-native-svg: ^2.2.0 => 2.2.0 
    @tamagui/react-native-web-lite: ^2.2.0 => 2.2.0 
    check-dependency-version-consistency: ^4.1.0 => 4.1.1 
    expo: ^54 => 54.0.31 
    husky: ^9.1.6 => 9.1.7 
    turbo: ^1.13.4 => 1.13.4 
    typescript: ~5.8.3 => 5.8.3 
    ultra-runner: ^3.10.5 => 3.10.5 
    vitest: ^2.1.1 => 2.1.9 
```

dennytosp:
Fixes #4031

### Problem

`createComponent` caches the composed ref on `stateRef` and closes over `forwardedRef`:

```ts
if (!stateRef.current.composedRef) {
  stateRef.current.composedRef = composeRefs<TamaguiElement>(
    (x) => (stateRef.current.host = x as TamaguiElement),
    forwardedRef,          // captured on the first render, forever
    setElementProps,
    animatedRef
  )
}
viewProps.ref = stateRef.current.composedRef
```

The cache is deliberate — one stable ref identity means React never detaches and reattaches the host mid-life. But it also pins the *first* `forwardedRef`. A parent that swaps ref identity later never reaches the node again, and because `viewProps.ref` keeps the same identity React never re-runs it either.

That's #4031: react-hook-form re-registers a field after `reset()`, its `field.ref` is a new callback, and `setFocus()` / `shouldFocusError` silently stop working from then on.

### Fix

Keep exactly one composed-ref identity, but read the forwarded ref through `stateRef` instead of closing over it, and hand the host across when the identity changes:

```ts
const prevForwardedRef = stateRef.current.composedForwardedRef
stateRef.current.composedForwardedRef = forwardedRef

if (!stateRef.current.composedRef) {
  stateRef.current.composedRef = composeRefs<TamaguiElement>(
    (x) => (stateRef.current.host = x as TamaguiElement),
    (x) => setRef<TamaguiElement>(stateRef.current.composedForwardedRef, x),
    setElementProps,
    animatedRef
  )
}

useIsomorphicLayoutEffect(() => {
  if (!stateRef.current.composedRefAttached) {
    stateRef.current.composedRefAttached = true
    return                       // the composed ref already delivered on mount
  }
  if (prevForwardedRef === forwardedRef) return
  setRef<TamaguiElement | null>(prevForwardedRef, null)
  if (stateRef.current.host) {
    setRef<TamaguiElement>(forwardedRef, stateRef.current.host)
  }
}, [forwardedRef])
```

The mount guard matters: without it a callback ref is invoked twice for a single attach.

### A rejected first attempt, and why it's worth knowing

My first version simply keyed the cache on the forwarded ref and rebuilt the composed ref when it changed. It passes the same tests — but it made React detach and reattach the host on every render for any component receiving an inline callback ref, which is precisely what the cache exists to prevent.

That showed up as a measurable regression: `Accordion.test.tsx` (`reanimated: toggling and reversing samples numeric wrapper height`) failed **2/10** runs with it, against **0/12** on main. The version in this PR removes that — the host is never reattached.

### Behavioural change worth your call

An unstable ref (inline arrow, or anything whose identity changes per render) now gets invoked on each change instead of once, ever. That is what React itself does with a changing ref callback, so I believe it is the correct semantics — but it is a change, and there are components inside Tamagui that pass unstable refs. Instrumenting the accordion page showed 10 handovers, all `function -> function`, mostly from `CollapsibleTrigger`. Stabilising those looks worthwhile on its own, and would make this path free; I left it out of this PR.

If you would rather keep the old deviation and fix only object refs, say so and I will narrow it.

### Tests

`code/core/core-test/composedRef.web.test.tsx`:

- a swapped `RefObject` receives the node, and the old one is released — red on main (`expected null not to be null`)
- a swapped callback ref is invoked with the node, and the old one with `null` — red on main (`expected [['a', true]] to contain ['a', false]`)
- a **stable** ref is not detached on re-render — this one guards the cache and would have caught the rejected attempt above

### Checks

- `bun run typecheck` ✅
- `bun run lint` ✅
- `turbo test:native` (static-tests, core-test, components-test) 3/3 ✅
- `turbo test:web` 13/15 — the two failures are `test-next-turbopack` and `next15-plus-cli-optimize`, which fail identically on a clean `main` here
- kitchen-sink Playwright `default` project: **582 passed, 13 skipped, 0 failed, 0 flaky**
- 16 paired Accordion runs (rebuild main / rebuild this branch, alternating, same machine state): 0 failures on either side


natew:
Reviewed. The diagnosis is right and the write-up is good work — especially documenting the rejected first attempt and keeping the stable-ref regression test as a guard against it. Two things before this can land.

### The handover silently no-ops under StrictMode

The fix reads "previous" out of a ref it writes **during render**:

```ts
const prevForwardedRef = stateRef.current.composedForwardedRef
stateRef.current.composedForwardedRef = forwardedRef   // render-phase write
```

StrictMode double-invokes render. Pass 1 sets `composedForwardedRef = refB` and computes `prev = refA`. Pass 2 recomputes `prev` and reads back `refB` — the value pass 1 just wrote. The effect closes over pass 2's values, hits `if (prevForwardedRef === forwardedRef) return`, and bails. Same hazard for any render React discards and re-runs concurrently.

I isolated the mechanism in a plain-React jsdom probe (no Tamagui), swapping `refA` → `refB`:

```
NON-STRICT  calls [["a",true],["a",false],["b",true]]   handovers 1   pass
STRICT      calls [["a",true],["a",false],["a",true]]   handovers 0   fail
```

Under StrictMode `refB` never receives the node and `refA` is left holding it — the `a,false`/`a,true` pair is just StrictMode's own detach/reattach. The three tests here pass because testing-library doesn't wrap in StrictMode, so the fix is green while doing nothing for anyone in dev.

### The repair

Track what the ref **actually attached to**, from inside the callback rather than from render:

```ts
stateRef.current.composedRef = composeRefs<TamaguiElement>(
  (x) => {
    stateRef.current.host = x as TamaguiElement
    const cur = stateRef.current.composedForwardedRef
    stateRef.current.attachedForwardedRef = x ? cur : undefined
    setRef<TamaguiElement>(cur, x)
  },
  setElementProps,
  animatedRef
)

useIsomorphicLayoutEffect(() => {
  const attached = stateRef.current.attachedForwardedRef
  if (attached === forwardedRef) return   // covers mount, so no composedRefAttached flag
  if (attached) setRef(attached, null)
  stateRef.current.attachedForwardedRef = forwardedRef
  if (stateRef.current.host) setRef(forwardedRef, stateRef.current.host)
}, [forwardedRef])
```

Same probe, this mechanism: `[["a",true],["a",false],["b",true]]` in **both** modes. It's idempotent, so a repeated effect run can't double-fire, and it drops the `composedRefAttached` flag entirely.

Please add a StrictMode variant of all three existing tests — the current ones can't catch this class of bug.

### On the behavior change you flagged

Agreed it's the correct React semantics, and it's your call to have raised it. But 10 handovers on the accordion page, mostly from `CollapsibleTrigger`, means those components would start paying a detach/reattach round trip through user callbacks on every render. I'd rather stabilize those refs in the same change than land the churn first and clean up after. Happy to take that as a separate commit in this PR.

natew:
Landed in 26eb212569 with you as co-author, thanks @dennytosp - the diagnosis was right. I kept your fix's shape and changed two things: the current forwarded ref is read off `stateRef` at attach time instead of being closed over, and the handover comparison reads what the ref callback actually attached rather than a pointer written during render, so it survives StrictMode's second render pass. Tests cover a swapped object ref, a swapped callback ref, and the no-detach-on-rerender guarantee, each in plain React and StrictMode.

PR #4181, opened 2026-08-22
fix(core): keep config theme declaration order so the default theme isn't dark

### LINKED ISSUE #3764: Input component's default background color is fixed in dark
### Current Behavior

Image

Just 2 days ago, this issue is not raised in my case. 
But suddenly today, my input component's background color is fixed in dark.
I tried to change my test device's native theme, but it's not work.
When I tested about this issue with 'react-native' <TextInput/>, and its bg color is white.

this is my own component code
```tsx
import { ComponentProps, memo, useMemo } from 'react';
import { FlexAlignType, OpaqueColorValue } from 'react-native';
import {
  ColorTokens,
  FontSizeTokens,
  FontWeightTokens,
  Label,
  SizeTokens,
  Stack,
  Input as TamaguiInput,
} from 'tamagui';

type IInput = {
  w?: SizeTokens;
  labelContent?: string;
  labelFontSize?: FontSizeTokens;
  labelFontWeight?: FontWeightTokens;
  labelPosition?: 'flex-start' | 'center' | 'flex-end';
  labelFontColor?: ColorTokens | OpaqueColorValue;
} & ComponentProps<typeof TamaguiInput>;

export const Input = memo<IInput>(
  ({ w, labelContent, labelFontSize, labelFontWeight, labelPosition, labelFontColor, ...rest }) => {
    const width = useMemo<SizeTokens>(() => {
      if (!w) {
        return '$fluid';
      }
      return w;
    }, [w]);

    const labelAlignItems = useMemo<FlexAlignType>(() => {
      if (!labelPosition) {
        return 'flex-start';
      }
      return labelPosition;
    }, [labelPosition]);

    const labelWeight = useMemo<FontWeightTokens>(() => {
      if (!labelFontWeight) {
        return '500';
      }
      return labelFontWeight;
    }, [labelFontWeight]);

    const labelColor = useMemo<ColorTokens | OpaqueColorValue>(() => {
      if (!labelFontColor) {
        return '$colors.black';
      }
      return labelFontColor;
    }, [labelFontColor]);

    return (
      <Stack width={width} justify="center" items={labelAlignItems} gap="$size.x1">
        {labelContent && (
          <Label
            width="$fit"
            px="$size.x2"
            fontSize={labelFontSize}
            fontWeight={labelWeight}
            color={labelColor}
            lineHeight="$1"
            htmlFor={labelContent}>
            {labelContent}
          </Label>
        )}
        <TamaguiInput width={width} {...rest} />
      </Stack>
    );
  }
);
```

and this is my test Input code with default tamagui input, not my own component. `import {Input} from 'tamagui'`
```tsx
<Input width="$fluid" height="$size.x12" placeholder="검색" px="$size.x3" borderColor="$colors.lightGray" />
```

Both things are in same issue. 

### Expected Behavior

_No response_

### Tamagui Version

```markdown
1.143.0
```

```markdown
IOS
```

### Reproduction

```markdown
add <Input/> in ios native device
```

### System Info

```markdown
n/a
```

dennytosp:
Fixes #3764

### Problem

An app that renders `<TamaguiProvider config={config}>` without a `defaultTheme` gets the **dark** theme, whatever its config declares first. That is what #3764 reports as "Input's background color is fixed in dark" — the Input isn't hardcoding anything (`code/ui/input/src/shared.tsx` uses `$background`/`$borderColor`), the whole app is simply in the dark theme.

Two pieces combine:

- `TamaguiProvider` falls back to the first key of the runtime themes object — `code/core/web/src/views/TamaguiProvider.tsx:20,36`, `defaultThemeProp || firstThemeKey(config) || 'light'`.
- `getThemesDeduped` iterates `Object.keys(themes).sort()` (`code/core/web/src/createTamagui.ts:282`), and `proxyThemesToParents` rebuilds `config.themes` from that array — so the runtime object comes out alphabetical, and `dark` sorts before `light`.

Confirmed against built main:

```
RAW     first keys: ["light","dark","dark_blue","dark_Card",...]
RUNTIME first keys: ["dark","dark_Button","dark_Card","dark_blue",...]
useThemeName() with no defaultTheme -> "dark"
```

The sort itself is deliberate — its comment says "deterministic CSS output order" — so the bug is that one array is doing double duty: it drives both the CSS cascade and the runtime key order.

### Fix

Leave the sorted array alone (see below for why that matters) and hand `proxyThemesToParents` the config's declaration order, so only the runtime object changes:

```ts
const themes = proxyThemesToParents(dedupedThemes, Object.keys(themesIn))
```

`proxyThemesToParents` now inserts in declaration order first, then falls through to its existing loop for anything the order didn't cover (and behaves exactly as before when the argument is omitted). Parent resolution is untouched — it already runs as a separate first pass over `themesRaw`, so it never depended on ordering.

**The emitted CSS is byte-for-byte identical to main.** I verified that by dumping `config.getCSS()` on both and diffing — 8041 bytes, no difference.

### The version I threw away

My first attempt was the obvious one: drop the sort in `getThemesDeduped` so the runtime object inherits declaration order directly. It made the new tests pass — and broke `ThemeReset.test.tsx` ("Reset from dark with button shows different themes"), **4/4 runs failed against 3/3 passing on main**.

The reason is that theme rules land at equal specificity, so their order in the sheet decides the cascade. Alphabetical emission puts `dark` before `light`; declaration order flips that, and `<Theme reset>` resolves to the wrong one. So the sort is load-bearing for CSS and only incidentally wrong for the runtime object — which is why the fix above separates the two rather than removing it.

### Behaviour change

Apps that have been silently running dark since this regression will go back to their first declared theme (light, for `@tamagui/config` v4). That is the documented intent, and passing `defaultTheme` explicitly overrides it either way — there is a test for that.

### Tests

`code/core/core-test/defaultThemeOrder.web.test.tsx`:

- the runtime themes object keeps the config's first-declared theme first
- a provider with no `defaultTheme` resolves `useThemeName()` to `light`
- an explicit `defaultTheme="dark"` still wins (guards against over-correcting)

The first two are red on main (`expected 'dark' to be 'light'`), the third passes both ways.

### Checks

- `bun run typecheck` ✅
- `bun run lint` ✅
- `turbo test:native` (static-tests, core-test, components-test) 3/3 ✅ — no CSS snapshot churn
- `turbo test:web` 13/15 — the two failures are `test-next-turbopack` and `next15-plus-cli-optimize`, which fail identically on a clean `main` here
- kitchen-sink Playwright `default` project: **581 passed, 13 skipped, 0 failed**; `ThemeReset.test.tsx` verified 4/4 green


natew:
Thanks for this @dennytosp — the diagnosis was right and it's now fixed on main via #4200, which reimplements it with a smaller change and keeps you as co-author.

The difference is only in where the ordering happens. This PR threads a `declarationOrder` map through the two-pass insert in `proxyThemesToParents`, which restructures the existing loop. #4200 leaves that loop untouched and reorders the result object's keys once at the end:

```ts
// dedupedThemes is sorted for deterministic CSS cascade, but TamaguiProvider
// falls back to Object.keys(themes)[0] — reorder to match config (#3764)
if (declarationOrder) {
  const ordered: Record<string, ThemeParsed> = {}
  for (const k of declarationOrder) if (k in themes) ordered[k] = themes[k]
  for (const k in themes) if (!(k in ordered)) ordered[k] = themes[k]
  return ordered
}
```

Same behavior, one plain object instead of a Map plus a helper closure, and nothing added to the hot path. Your other five PRs (#4174 #4176 #4177 #4180 #4182) all merged as-is — thanks for the batch.

PR #4184, opened 2026-08-23
fix(sheet): track keyboard state without provider

natew:
## Root cause

`useKeyboardControllerSheet` selected `react-native-keyboard-controller` events whenever the module was configured. Those events only fire while `KeyboardProvider` is mounted. Detox deliberately launches without that provider, so the sheet stayed at its pre-keyboard position even though the keyboard was visible. The safe-area path remained live and reported the correct top inset.

## Fix

Use React Native keyboard notifications as the single source for sheet visibility and height. Keep keyboard-controller only for imperative dismissal when configured.

## Validation

- `bun run build` in `code/ui/sheet`
- `SheetKeyboardFitContent.test.ts`: 1 passed
- `SheetPressRegression.test.ts`: 4 passed, including the keyboard movement regression

No test assertions, thresholds, timeouts, retries, or swipe behavior changed.

### DIFF
--- a/code/ui/sheet/src/useKeyboardControllerSheet.native.ts
+++ b/code/ui/sheet/src/useKeyboardControllerSheet.native.ts
@@ -1,12 +1,12 @@
 /**
- * Native implementation of keyboard controller sheet hook.
+ * native implementation of keyboard controller sheet hook.
  *
- * Simplified to just track keyboard state (height, visibility).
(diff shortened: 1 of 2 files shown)


natew:
Replacing this with a PR from the v2-prefixed branch so the push runs Native Tests (Detox).

PR #4196, opened 2026-08-31
fix(animations-reanimated): guard onStart on non-animation values

### LINKED ISSUE #4193: reanimated driver crashes on remount: Cannot create property 'onStart' on number 'N' (numeric style prop + already-painted component)
### Current Behavior

Any styled component with `animation`/`transition` (reanimated driver) that has a **numeric** animated style prop (e.g. `width={54}`) crashes with:

```
Render Error: Cannot create property 'onStart' on number '54'
```

The crash does **not** happen on a component's first-ever mount in the app's lifetime. It happens when a component that has **already painted once** gets **remounted** (actually torn down and recreated — e.g. after an error-boundary reset, a full tree wipe, or a fast unmount→remount toggle).

Once it fires once, every subsequent `withSpring`/`withTiming` call across the app also returns a raw number instead of an animation object, because `IN_STYLE_UPDATER.current` in `react-native-reanimated` is never reset after a throw (see root cause below) — so every animated component in the app starts crashing the same way until a full JS reload.

### Expected Behavior

Remounting a component with `animation` should animate from its previous painted value (or skip the animation), without crashing and without poisoning the animation driver for the rest of the session.

### Tamagui Version

- `tamagui`: 2.7.7
- `@tamagui/animations-reanimated`: 2.7.7
- `@tamagui/config`: 2.7.7
- Also reproduces at `tamagui` 2.7.4 / `react-native-reanimated` 4.5.1 (pre-upgrade) — the two implicated code paths are byte-identical before and after the upgrade

- iOS Simulator (iPhone 17 Pro, iOS 26.5), React Native 0.86.2 (new arch), Expo SDK 57
- Physical device not tested, but the failing code path has no platform branch
- Web is not affected (the CSS driver takes a different code path)

### Reproduction

Single-file repro — works in any Expo SDK 57 app with `react-native-reanimated` 4.x + `@tamagui/animations-reanimated`:

```tsx
import { createAnimations } from '@tamagui/animations-reanimated'
import { createTamagui, TamaguiProvider, YStack } from 'tamagui'
import { useState } from 'react'
import { Button } from 'react-native'

const config = createTamagui({
  animations: createAnimations({
    quick: { type: 'spring', damping: 25, mass: 1, stiffness: 550 },
  }),
  // ...rest of a normal v5-style config (tokens/themes omitted for brevity)
})

function Fab() {
  return (
    <YStack
      width={54}
      height={54}
      animation="quick"
      pressStyle={{ y: 2 }}
      backgroundColor="red"
    />
  )
}

export default function App() {
  // Toggling `mounted` off and back on remounts <Fab/> — this is the trigger.
  const [mounted, setMounted] = useState(true)
  return (
    <TamaguiProvider config={config}>
      <Button title="remount" onPress={() => {
        setMounted(false)
        requestAnimationFrame(() => setMounted(true))
      }} />
      {mounted ? <Fab /> : null}
    </TamaguiProvider>
  )
}
```

Steps:

1. Mount the app, let `<Fab/>` paint once (first mount does **not** crash — see root cause).
2. Tap "remount" to unmount and immediately remount `<Fab/>`.
3. Crash on the remount: `Cannot create property 'onStart' on number '54'`.

(We originally hit this via a "wipe local cache" action that clears app state and forces a full tree remount. Navigating away from a screen containing such a component and back should also work, if the instance is fully unmounted rather than kept alive.)

### Root cause (traced through source; line numbers from the versions above)

1. `animation="quick"` (+ `pressStyle`) routes style resolution through the reanimated driver in `@tamagui/animations-reanimated`.

2. In `createAnimations.native.js`, `useAnimations`'s style updater calls `animateSnapshotValue` once per animated key (`createAnimations.native.js:1066`).

3. `animateSnapshotValue` has a guard (`createAnimations.native.js:385`):

   ```js
   if (!previouslyEmitted && seedValue === void 0 && !cycleGated) {
     return targetValue; // plain first mount takes this path — no crash
   }
   ```

   A component's very first mount has no `seedValue`, so it returns the raw target value and never reaches the animation call. This is why the bug does not reproduce on cold start.

4. `seedValue` comes from `lastPaintedRef` — values from a **previous paint** (`buildSnapshot`, `createAnimations.native.js:165-240`, populated at lines 897 and 978). On a genuine remount (new component instance) this ref is fresh — the crash additionally requires the style-emitter state surviving the fast unmount/remount cycle so that `seedValue` is defined on the new instance's first style computation. (We did not dig further into exactly which internal cache carries it, since the fix belongs in `applyAnimation`'s output validation regardless of how `seedValue` becomes defined.)

5. When `seedValue` is defined, `animateSnapshotValue` calls `applyAnimation` (`createAnimations.native.js:354`):

   ```js
   var applyAnimation = function (targetValue, config, callback, seedValue) {
     "worklet";
     var animatedValue;
     if (config.type === "timing") {
       animatedValue = withTiming(targetValue, reanimatedConfig, callback);
     } else {
       animatedValue = withSpring(targetValue, reanimatedConfig, callback);
     }
     if (seedValue !== void 0 || validateStartAsColor) {
       var innerOnStart = animatedValue.onStart;
       animatedValue.onStart = function (...) { ... };  // <-- line 368, throws
     }
     ...
   };
   ```

   There is **no check** that `animatedValue` is actually an animation object before writing `.onStart` on it.

6. `withTiming`/`withSpring` are built on `defineAnimation` (`react-native-reanimated`, `lib/module/animation/util.js:350`):

   ```js
   export function defineAnimation(starting, factory) {
     'worklet';
     if (globalThis.__RUNTIME_KIND === RuntimeKind.ReactNative && IN_STYLE_UPDATER.current) {
       return starting; // returns the raw target value, not an animation object
     }
     ...
   }
   ```

7. `IN_STYLE_UPDATER.current` is set by `initialUpdaterRun` (`util.js:66-70`), which `useAnimatedStyle` calls on a component's first style computation after mount (`hook/useAnimatedStyle.js:362`, guarded by `if (!animatedUpdaterData.current)` — a fresh `useRef`).

8. So on the specific remount where the style-updater's *first* run coincides with a defined `seedValue`, `withSpring` returns the raw number (here `54`, the `width`) instead of an animation object, and `@tamagui/animations-reanimated` immediately writes `.onStart` on it → `TypeError: Cannot create property 'onStart' on number '54'`.

### Secondary bug that turns one crash into a persistent one

`initialUpdaterRun` has no `try/finally`:

```js
export function initialUpdaterRun(updater) {
  IN_STYLE_UPDATER.current = true;
  const result = updater();   // if this throws, the flag below never resets
  IN_STYLE_UPDATER.current = false;
  return result;
}
```

Once `updater()` throws (as it does here, inside the `.onStart` assignment), `IN_STYLE_UPDATER.current` is left `true` on the JS runtime forever. Every subsequent `withSpring`/`withTiming` call anywhere in the app then also returns a raw primitive, so **every** animated component with a numeric style prop starts crashing the same way. The only recovery is a full JS reload.

This means the practical severity is much worse than "one component crashes once" — a single transient reanimated-side throw during a style update takes down animation for the rest of the session.

### Suggested fixes

Either of these independently prevents the crash (both would be ideal, since they're in different packages):

1. **`@tamagui/animations-reanimated`**, `applyAnimation` (`createAnimations.native.js:366-374`): validate that `animatedValue` is actually an object before assigning `.onStart`. Reanimated already marks valid animation definitions with `create.__isAnimationDefinition = true` (`util.js:366`) — checking for that flag would be more precise than a bare `typeof` check:

   ```js
   if ((seedValue !== void 0 || validateStartAsColor) &&
       animatedValue !== null && typeof animatedValue === "object") {
     var innerOnStart = animatedValue.onStart;
     animatedValue.onStart = function (...) { ... };
   }
   ```

2. **`react-native-reanimated`**, `initialUpdaterRun` (`animation/util.js:66-70`): wrap in `try/finally` so a throw during the initial updater run doesn't leave `IN_STYLE_UPDATER.current` stuck `true` (will be reported to software-mansion/react-native-reanimated separately):

   ```js
   export function initialUpdaterRun(updater) {
     IN_STYLE_UPDATER.current = true;
     try {
       return updater();
     } finally {
       IN_STYLE_UPDATER.current = false;
     }
   }
   ```

alexey-khatskelevich:
Fixes #4193 

## Summary

Prevents the `Cannot create property 'onStart' on number 'N'` crash in `@tamagui/animations-reanimated`.

`applyAnimation` attaches a seed-substituting `onStart` wrapper onto the value returned by `withSpring`/`withTiming` without checking that the call actually produced an animation definition. When the style updater runs inside reanimated's `initialUpdaterRun` — `useAnimatedStyle`'s first style computation, which happens on remount of an already-painted component — `defineAnimation` short-circuits and returns the raw starting value, a primitive:

- in react-native-reanimated `src/animation/util.ts`, `initialUpdaterRun` sets `IN_STYLE_UPDATER.current = true`, and `defineAnimation` returns `starting` when that flag is set on the JS runtime;
- tamagui then writes `.onStart` onto that primitive → `TypeError: Cannot create property 'onStart' on number '54'`;
- `initialUpdaterRun` has no try/finally, so the throw leaves `IN_STYLE_UPDATER.current` stuck at `true` — every subsequent `withSpring`/`withTiming` call also returns a raw value, and animation breaks app-wide until a full JS reload.

## The fix

A definition from `withSpring`/`withTiming` is either the decorated animation object (UI thread) or the `__isAnimationDefinition` factory function (JS thread) — both can carry the `onStart` property. Only the primitive short-circuit path is now skipped: the key just paints its target for that first computation instead of crashing, and seeds plus the color-validation wrapper are untouched on every normal path.

This is the tamagui-side half of the fix suggested in #4193; either half independently prevents the crash. The reanimated-side half (try/finally around `initialUpdaterRun`) is tracked in software-mansion/react-native-reanimated#10433.

## Testing

- Package build (esm/cjs/types), `oxfmt --check` and `oxlint` all pass.
- Walkthrough of the #4193 repro (remount of a painted `width={54}` + `animation="quick"` + `pressStyle` component): on that remount `withSpring` returns `54`, the assignment is now skipped, the render completes, and the next mapper runs (outside `initialUpdaterRun`) produce real animation definitions again — before the patch the same path threw the documented `TypeError`.

### DIFF
--- a/code/core/animations-reanimated/src/createAnimations.tsx
+++ b/code/core/animations-reanimated/src/createAnimations.tsx
@@ -426,7 +426,13 @@ const applyAnimation = (
   // wrapper substitutes the seed for whatever start value reanimated passes.
   // color history gets the same treatment only when reanimated's own parser
   // rejects it, preserving valid in-flight values during interruption.
-  if (seedValue !== undefined || validateStartAsColor) {
+  // inside initialUpdaterRun withSpring returns a raw value, not a definition —
+  // writing onStart onto it throws and poisons animation app-wide until reload.
+  if (
+    (seedValue !== undefined || validateStartAsColor) &&
+    animatedValue !== null &&
+    (typeof animatedValue === 'object' || typeof animatedValue === 'function')
+  ) {
     const innerOnStart = animatedValue.onStart
     animatedValue.onStart = (
       animation: unknown,


alexey-khatskelevich:
Fixes #4193

natew:
Thanks @alexey-khatskelevich — the diagnosis here is exactly right, and this is now fixed on main (`25d97e1ab6`, then `a756be9a5f`).

One thing worth flagging, because it cost us a follow-up commit. The positive object check in this PR:

```ts
animatedValue !== null &&
(typeof animatedValue === 'object' || typeof animatedValue === 'function')
```

is what shipped first, and it broke every native animation. Inside a `'worklet'`, SWC's ES5 transform rewrites `typeof x === 'object'` into a `_type_of()` helper call, and that helper isn't a worklet — calling it on Reanimated 4's UI runtime throws `Tried to synchronously call a Remote Function`.

Main now guards with negated primitive checks, which SWC leaves as plain `typeof`:

```ts
const isAnimationDescriptor =
  animatedValue !== null &&
  typeof animatedValue !== 'number' &&
  typeof animatedValue !== 'string' &&
  typeof animatedValue !== 'boolean' &&
  typeof animatedValue !== 'undefined'
```

Closing as fixed. The write-up on the `IN_STYLE_UPDATER` stuck-flag cascade was genuinely useful for understanding why the crash poisons animation app-wide, so thank you for that.

PR #4197, opened 2026-09-02
fix(portal): skip gorhom root host under teleport

natew:
## Summary

react-native-teleport keeps one native host per name and the last registered wins. Tamagui's PortalProvider was also rendering a styleless gorhom `<PortalHost name="root">` inside teleport's own PortalProvider, which already mounts an absolute-fill root host. The gorhom host replaced the real one, laid out at zero height, and every teleported sheet drew offscreen.

This branch skips the gorhom root host when `getPortal().state.type === 'teleport'` (`6831debc28`) and adds a native test that records teleport host names: under teleport, Tamagui's PortalProvider must leave exactly one `'root'` host (teleport's), and an explicit `<PortalHost name="x">` still renders.

The same test is on v3-beta (`2baa0f55c9`). A canary for this SHA is already in flight; this PR is the main-line landing.

## Test plan

- [x] `TAMAGUI_TARGET=native vitest` `PortalTeleportRootHost.native.test.tsx` in `code/ui/components-test` (2 passed)
- Temporarily dropping the `portalState.type !== 'teleport'` guard on v3-beta produced two `'root'` hosts and failed the assertion; restoring it passed

### DIFF
--- a/code/ui/components-test/PortalTeleportRootHost.native.test.tsx
+++ b/code/ui/components-test/PortalTeleportRootHost.native.test.tsx
@@ -1,0 +1,78 @@
+import { getDefaultTamaguiConfig } from '@tamagui/config-default'
+import { TamaguiProvider, createTamagui } from '@tamagui/core'
+import { getPortal } from '@tamagui/native'
+import { PortalHost, PortalProvider } from '@tamagui/portal'
+import React from 'react'
+import TestRenderer, { act } from 'react-test-renderer'
+import { afterEach, describe, expect, test } from 'vitest'
+
+const conf = createTamagui(getDefaultTamaguiConfig('native'))
+
+const recordedHostNames: string[] = []
+
+function Portal({ children }: { children?: React.ReactNode }) {
+  return <>{children}</>
+}
+
+function PortalHostFake({ name }: { name: string }) {
+  recordedHostNames.push(name)
+  return null
+}
+
+function PortalProviderFake({ children }: { children?: React.ReactNode }) {
+  return (
+    <>
+      {children}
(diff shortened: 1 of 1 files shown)


natew:
Not merging this one — it's a re-land of code that was already reverted.

The implementation on this branch is byte-identical to `6831debc28`, which `a756be9a5f` reverted because Android Detox never settled: the view hierarchy didn't reach a stable state once the gorhom root `PortalHost` was removed.

The new `PortalTeleportRootHost.native.test.tsx` can't catch that. It substitutes fakes for the whole teleport module:

```ts
function PortalHostFake({ name }: { name: string }) {
  recordedHostNames.push(name)
  return null   // renders nothing, so there is no view hierarchy to settle
}
...
expect(recordedHostNames).toEqual(['root'])
```

That asserts which name strings got rendered. The Android failure was about a real native view hierarchy. This test is green either way, so it isn't a check on the thing that broke.

The underlying diagnosis still reads as correct — two hosts registered under the same name, last one wins, the unstyled one lays out at zero height and draws teleported sheets offscreen. But something on Android depends on that gorhom root host existing, and re-landing without knowing what would just produce the same revert. Reopening this needs the Detox hang understood first, and a test that runs against the real hierarchy rather than fakes.

natew:
Closing this one. It re-lands the `shouldAddRootHost && portalState.type !== 'teleport'` skip that was already reverted in a756be9a5f for breaking Android Detox, and nobody diagnosed why it broke. Re-landing the same change without that diagnosis just buys the same red back.

Worth reopening once someone can say what the root host was actually doing for the Android native path, ideally with a Detox run that fails on the old code for a named reason.

Issue #2970, opened 2024-10-06
Switch component does not support RTL


Ammar1999y:
### Current Behavior

![CREATE GOOGLE MAP API](https://github.com/user-attachments/assets/f1f7ed3e-0347-4a2c-8977-044dcff55edf)

### Expected Behavior

I try to use react native animated driver in `tamagui.config.ts`, but still didn't fix it.
```
import { config } from "@tamagui/config/v3";
import { createTamagui } from "tamagui";
import { createAnimations } from "@tamagui/animations-react-native";

const appConfig = createTamagui({
  ...config,
  animations: createAnimations({
    bouncy: {
      type: "spring",
      damping: 10,
      mass: 0.9,
      stiffness: 100,
    },
    lazy: {
      type: "spring",
      damping: 20,
      stiffness: 60,
    },
    quick: {
      type: "spring",
      damping: 20,
      mass: 1.2,
      stiffness: 250,
    },
  }),
});
export type AppConfig = typeof appConfig;

declare module "tamagui" {
  interface TamaguiCustomConfig extends AppConfig {}
}

export default appConfig;
```

### Tamagui Version

```markdown
"tamagui": "^1.112.3"
"@tamagui/config": "^1.112.3"
```

### Platform (Web, iOS, Android)

```markdown
Android (The one I tried)
```

### Reproduction

```markdown
https://github.com/Ammar1999y/tamagui-RTL-error
```

### System Info

```markdown
System:
    OS: Windows 10 10.0.19045
    CPU: (2) x64 Intel(R) Core(TM)2 Duo CPU     E7500  @ 2.93GHz
    Memory: 1.44 GB / 7.75 GB
  Binaries:
    Node: 20.11.1 - C:\Program Files\nodejs\node.EXE
    Yarn: 1.22.21 - ~\AppData\Roaming\npm\yarn.CMD
    npm: 10.2.5 - C:\Program Files\nodejs\npm.CMD
  Browsers:
    Chrome: 129.0.6668.90
    Edge: Spartan (44.19041.3570.0), Chromium (125.0.2535.92)
    Internet Explorer: 11.0.19041.3570
  npmPackages:
    @babel/core: ^7.20.0 => 7.25.2
    @tamagui/config: ^1.112.3 => 1.112.3
    @types/react: ~18.2.45 => 18.2.79
    expo: ~51.0.28 => 51.0.34
    expo-router: ~3.5.23 => 3.5.23
[...]

PR #4208, opened 2026-09-08
fix(switch): mirror the thumb translate under rtl so it travels across the frame

giaBaoJS:
Refs #2970

> **Draft.** The approach below does not hold up. `useDirection()` is not connected to the
> signal that flips `alignSelf`, so the case reported in #2970 (Android, `I18nManager` RTL,
> no `DirectionProvider`) is unaffected, and an LTR document with a `DirectionProvider`
> would translate the thumb the wrong way. The `Fixes` link is removed so this cannot close
> the issue. See the comment below. Reworking it to derive one flag from the layout direction.

### What breaks

Under RTL the `Switch` thumb slides the wrong way. Instead of crossing the track it moves out of the frame, which is what the reproduction in the issue shows.

### Root cause

`code/ui/switch/src/createSwitch.tsx:75-86`. The thumb is positioned by two properties that disagree about direction:

```tsx
alignSelf={initialChecked ? 'flex-end' : 'flex-start'}
x={x}
```

`alignSelf` resolves against the writing direction, so the resting edge already flips when the direction is RTL. The `x` translate does not flip: `translateX` is a physical transform on both web and React Native. So the thumb is anchored on the right and then translated further right by `frameWidth - thumbWidth`, ending up outside the frame.

That matches the workaround a second reporter posted on the issue, which negates `translateX` by hand when RTL.

### The fix

Read the direction with `useDirection()` from `@tamagui/use-direction` (the same hook `Slider`, `Tabs`, `Accordion`, `ToggleGroup`, `RovingFocusGroup` and `createBaseMenu` already use) and negate the translate when it is `rtl`, so the translate agrees with the anchor:

```tsx
const direction = useDirection()
const thumbX = direction === 'rtl' ? -x : x
```

LTR is untouched: `useDirection()` returns `'ltr'` when there is no `DirectionProvider`, so `thumbX === x`.

### How I proved it

New test `code/ui/components-test/Switch.web.test.tsx`. It renders `Switch.Thumb` inside a `SwitchStyledContext.Provider` with a frame width (jsdom reports no layout, so the width is supplied directly), toggles it to checked, and reads the resulting `translateX`. One test guards the existing LTR travel, the other asserts RTL travel mirrors it.

Green with the fix:

```
 ✓ Switch.web.test.tsx > Switch.Thumb travel > should move the thumb towards the trailing edge in ltr
 ✓ Switch.web.test.tsx > Switch.Thumb travel > should mirror the thumb travel in rtl
 Test Files  1 passed (1)
      Tests  2 passed (2)
```

Counterfactual, reverting only `createSwitch.tsx` to its state on `main` and rebuilding the package (these tests resolve `@tamagui/switch` from `dist`):

```
 ✓ Switch.web.test.tsx > Switch.Thumb travel > should move the thumb towards the trailing edge in ltr
 × Switch.web.test.tsx > Switch.Thumb travel > should mirror the thumb travel in rtl
   → expected 40 to be -40 // Object.is equality
```

The RTL thumb translates `+40` instead of `-40`, which is the defect. The LTR guard stays green through both runs, so the test is not passing or failing for an unrelated reason.

Full CI test jobs after the change:

- `bun turbo run test:web --filter='!@tamagui/kitchen-sink' --concurrency=1`: `@tamagui/components-test` 32 passed (6 files). The only failing task is `integration`, which cannot launch Playwright locally because the browser binaries are not installed; unrelated to this change.
- `bun turbo run test:native --filter=@tamagui/static-tests --filter=@tamagui/core-test --filter=@tamagui/components-test --concurrency=1`: 3 successful, 3 total.

`bun run check:deps` (manypkg), oxlint and oxfmt are clean on the changed files. `@tamagui/use-direction` was added to `@tamagui/switch` dependencies, which is the only reason `bun.lock` is touched.

### DIFF
--- a/bun.lock
+++ b/bun.lock
@@ -3142,6 +3142,7 @@
         "@tamagui/stacks": "workspace:*",
         "@tamagui/switch-headless": "workspace:*",
         "@tamagui/use-controllable-state": "workspace:*",
+        "@tamagui/use-direction": "workspace:*",
         "@tamagui/use-previous": "workspace:*",
       },
       "devDependencies": {
--- a/code/ui/components-test/Switch.web.test.tsx
+++ b/code/ui/components-test/Switch.web.test.tsx
@@ -1,0 +1,49 @@
+import '@testing-library/jest-dom'
+import 'vitest-axe/extend-expect'
+
+import { getDefaultTamaguiConfig } from '@tamagui/config-default'
+import { TamaguiProvider, createTamagui } from '@tamagui/core'
+import { Switch, SwitchStyledContext } from '@tamagui/switch'
+import { DirectionProvider } from '@tamagui/use-direction'
+import { render } from '@testing-library/react'
+import { describe, expect, it } from 'vitest'
+
+const conf = createTamagui(getDefaultTamaguiConfig())
+
+const FRAME_WIDTH = 40
+const THUMB_TEST_ID = 'switch-thumb'
+
+// the thumb reads the measured frame width off context, and jsdom reports no
+// layout, so feed the width in directly to keep the travel distance non-zero
+function ThumbTest({ dir, active }: { dir: 'ltr' | 'rtl'; active: boolean }) {
+  return (
+    <TamaguiProvider config={conf} defaultTheme="light">
+      <DirectionProvider dir={dir}>
+        <SwitchStyledContext.Provider active={active} frameWidth={FRAME_WIDTH}>
+          <Switch.Thumb unstyled data-testid={THUMB_TEST_ID} />
+        </SwitchStyledContext.Provider>
+      </DirectionProvider>
+    </TamaguiProvider>
+  )
+}
+
+// renders unchecked then checks it, so the thumb starts at its resting edge
+// and we read the travel it makes towards the other edge
+function getCheckedTravel(dir: 'ltr' | 'rtl') {
+  const rendered = render(<ThumbTest dir={dir} active={false} />)
+  rendered.rerender(<ThumbTest dir={dir} active />)
+  const thumb = rendered.getByTestId(THUMB_TEST_ID)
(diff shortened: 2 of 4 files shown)


giaBaoJS:
Converting this to a draft. The approach does not hold up.

`useDirection()` reads a React context that nothing in the repo populates. No `DirectionProvider` is mounted anywhere in `code/`, and nothing bridges `I18nManager.isRTL` or the document direction into it. The property this PR tries to reconcile with, `alignSelf: 'flex-start'`, flips against the ambient layout direction instead: CSS `direction` on web, and Yoga's direction (fed by `I18nManager.isRTL`) on native. Those are two independent signals, so negating the translate off `useDirection()` does not line the two halves up.

Measured in jsdom, reading `translateX` off `Switch.Thumb`:

- `document.documentElement.dir = 'rtl'` with no provider: `useDirection()` returns `'ltr'` and the translate stays `+40`. That is the setup in #2970, where the reporter is on Android with `I18nManager` RTL, so this PR does not affect the reported case. The `Fixes #2970` line was wrong and I have removed it.
- LTR document with `DirectionProvider dir="rtl"`: the anchor stays on the left but the translate becomes `-40`, so the thumb travels left out of the frame. That case renders correctly on `main` today, so it would be a regression.

The test does not catch either problem. It asserts only that `rtl === -ltr` on `translateX`, and it renders the thumb outside `Switch` with no frame, so it never checks where the thumb lands. It stays green when I invert the `alignSelf` expression, and when I restrict the negation to the unchecked branch.

I am reworking it to derive one RTL flag from the direction the layout engine actually uses, and to drive both the anchor and the translate from that flag so they cannot disagree, following the way `Slider` converts direction into physical `left`/`right` edges. Sorry for the churn.

PR #4209, opened 2026-09-08
fix(core): stop stripping onScroll on native so ScrollView handlers fire

### LINKED ISSUE #4163: Bug: `onScroll` no longer triggers on native after updating to Tamagui v2
### Summary

After updating to Tamagui v2, `onScroll` no longer fires on native when using a plain Tamagui `ScrollView`. The ScrollView still scrolls, but the `onScroll` callback is never invoked. Wrapping the same ScrollView with reanimated's `createAnimatedComponent` makes `onScroll` work again.

We have a screen that relies on `onScroll` to track scroll position. After upgrading to Tamagui v2, the handler silently stopped firing on native. No error, no warning, the list just scrolls without the callback ever running. After further debugging we narrowed it down to Tamagui's prop handling on native.

### Minimal reproduction

Repo: https://github.com/idrakimuhamad/tamagui-onscroll-repro

The repro screen renders two identical ScrollViews side by side:

1. **Plain Tamagui ScrollView** — `styled(TamaguiScrollView, { flex: 1 })` with an `onScroll` handler
2. **AnimatedScrollView** — `Animated.createAnimatedComponent(TamaguiScrollView)` with the same `onScroll` handler

Both lists scroll. But:

| ScrollView | Scrolls? | `onScroll` fires? |
|---|---|---|
| Plain Tamagui `ScrollView` | ✅ | ❌ (counter stays `0`) |
| `AnimatedScrollView` (reanimated) | ✅ | ✅ (counter climbs) |

### Root cause

I asked my agent to dig, and it found out that Tamagui's `styled()` core strips `onScroll` on native. The prop is listed in `webPropsToSkip.native.ts`:

```ts
// node_modules/@tamagui/web/src/helpers/webPropsToSkip.native.ts
export const webPropsToSkip = {
  ...
  onScroll: 1,   // <-- stripped on native
  ...
}
```

Reanimated's `createAnimatedComponent` re-injects `onScroll` via its `PropsFilter`, which is why the wrapped ScrollView works.

### Question / reasoning

Why is `onScroll` in `webPropsToSkip` for native? `onScroll` is a legitimate native `ScrollView` prop AFAIK, it's not web-only. Stripping it on native means any plain Tamagui `ScrollView` silently loses scroll tracking, which is surprising and hard to debug (no error, just a dead callback).

Is this intentional? If so, what is the supported way to receive `onScroll` on a plain Tamagui `ScrollView` on native? Do we need to wrap every scrollable in `createAnimatedComponent` (which pulls in reanimated as a hard requirement), or should `onScroll` be removed from the native skip list?

### Expected behavior

`onScroll` should fire on a plain Tamagui `ScrollView` on native, matching the web behavior. If it must be stripped, there should be a documented, non-reanimated way to opt back in.

### Environment

- Tamagui 2.7.6
- Expo SDK 55
- React Native 0.83.2
- react-native-reanimated 4.2.2
- iOS (Expo Go)

giaBaoJS:
Fixes #4163

## What breaks

On native, `onScroll` never reaches the underlying React Native component. A plain Tamagui `ScrollView` scrolls normally but the handler is never called, with no error or warning. The reporter's workaround was to wrap the ScrollView in reanimated's `createAnimatedComponent`, whose `PropsFilter` re-injects the prop.

The same strip hits `Input` on native: `code/ui/input/src/Input.native.tsx` destructures `onScroll` and passes it into the styled `TextInput` frame, where it is dropped again.

## Root cause

`code/core/web/src/helpers/webPropsToSkip.native.ts:48` lists `onScroll: 1` among the web-only event handlers.

`code/core/web/src/helpers/skipProps.ts:29` merges that object into `skipProps` when `TAMAGUI_TARGET === 'native'`, and `getSplitStyles.tsx:385` drops any prop found in `skipProps` for a non-HOC component. `ScrollView` is a plain `styled(ScrollViewNative, ...)` (`code/ui/scroll-view/src/ScrollView.tsx:6`), so it takes that path and loses the prop before render.

`onScroll` is not web-only: it is a real prop on RN `ScrollView`, `FlatList` and `TextInput`. On web the list is empty (`webPropsToSkip.ts`), so web already forwards it, which is why the behaviour diverges by platform.

## The fix

Remove `onScroll` from the native skip list. This follows the existing precedent in the same file, where `pointerEvents` is explicitly excluded from the spread of `webOnlyStylePropsView` because it is a valid RN prop.

`code/core/web/types/helpers/webPropsToSkip.native.d.ts` is regenerated to match, since `types/` is tracked here and `SKIP_JS=1 bun run build` in `code/core/web` would otherwise leave the tree dirty with exactly that one-line deletion. It is the only file in `types/` that moves.

The skip list is global to native styled components, so this also lets `onScroll` reach RN `View` and `Text` hosts, where it simply never fires. Rendering `<View onScroll>`, `<Text onScroll>` and `<styled(View) onScroll>` under `TAMAGUI_TARGET=native` produces no throw and no warning, and the handler lands on the host element in all three cases (the only `console.error` in that run is the unrelated `react-test-renderer is deprecated` notice).

## How this was verified

Two tests added to `code/core/core-test/getSplitStyles.native.test.tsx`, one for a bare `View` and one behind a `styled()` wrapper.

Counterfactual, reverting only `webPropsToSkip.native.ts` and rebuilding. Rebuilding `@tamagui/web` alone is a no-op for these tests, because `code/packages/vite-plugin-internal/src/getConfig.ts` aliases both `@tamagui/core` and `@tamagui/web` to `@tamagui/core/native-test`, so `@tamagui/core` has to be rebuilt:

```
 x getSplitStyles > onScroll is forwarded on native
   -> expected undefined to be [Function onScroll] // Object.is equality
 x getSplitStyles > onScroll survives a styled() wrapper on native
   -> expected undefined to be [Function onScroll] // Object.is equality
 Tests  2 failed | 15 passed | 2 skipped (19)
```

With the fix restored: `17 passed | 2 skipped (19)`.

Full `unit-tests` CI job locally, after `bun run build:js`:

- `test:native` for `@tamagui/core-test`, `@tamagui/components-test`, `@tamagui/static-tests`: 3/3 tasks green (97 passed / 7 expected fail / 11 skipped, 27 passed, 32 passed).
- `test:web --filter='!@tamagui/kitchen-sink'`: `@tamagui/core-test` 210 passed, `@tamagui/components-test` 30 passed, `@tamagui/build` 14 passed, `create-tamagui` 24 passed. The `next15-plus-cli-optimize` starter fails on this machine with `error: Script not found "tamagui"`, which is a local CLI linking problem unrelated to this change.

## Follow-up, not in this PR

`onChange` sits in the same list, but it is not the same case and should probably stay skipped.

`code/ui/input/src/Input.native.tsx` destructures `onChange` at line 27 and never forwards it. Instead `handleChangeText` (lines 184 to 189) synthesizes a web-shaped event from `onChangeText`:

```tsx
onChange({ target: { value: text }, type: 'change' } as any)
```

So on native, Tamagui deliberately gives `onChange` the web payload, `e.target.value`. React Native's own `TextInput.onChange` is `NativeSyntheticEvent<TextInputChangeEventData>`, where the value is at `e.nativeEvent.text`. Unskipping `onChange` would let the raw RN event reach handlers on styled RN hosts while `Input` keeps synthesizing the web-shaped one, so the same prop name would carry two different payload shapes depending on which component you used. `onScroll` has no such shim anywhere, which is why it is safe to unskip and `onChange` is not.

### DIFF
--- a/code/core/core-test/getSplitStyles.native.test.tsx
+++ b/code/core/core-test/getSplitStyles.native.test.tsx
@@ -351,6 +351,19 @@ describe('getSplitStyles', () => {
     expect(style?.padding).toBeUndefined()
     expect(style?.paddingTop).toBeUndefined()
   })
+
+  test('onScroll is forwarded on native', () => {
+    const onScroll = () => {}
+    const { viewProps } = getSplitStylesFor({ onScroll })
+    expect(viewProps.onScroll).toBe(onScroll)
+  })
+
+  test('onScroll survives a styled() wrapper on native', () => {
+    const StyledView = styled(View, { backgroundColor: 'red' })
+    const onScroll = () => {}
+    const { viewProps } = getSplitStylesFor({ onScroll }, StyledView)
+    expect(viewProps.onScroll).toBe(onScroll)
+  })
 })
 
 describe.skip('getSplitStyles - pseudo prop merging', () => {
--- a/code/core/web/src/helpers/webPropsToSkip.native.ts
+++ b/code/core/web/src/helpers/webPropsToSkip.native.ts
@@ -45,7 +45,8 @@ export const webPropsToSkip = {
   onChange: 1,
   onInput: 1,
   onBeforeInput: 1,
-  onScroll: 1,
+  // onScroll is not listed here: it is a real React Native prop on
+  // ScrollView, FlatList and TextInput, so skipping it breaks them.
   onCopy: 1,
   onCut: 1,
   onPaste: 1,
--- a/code/core/web/types/helpers/webPropsToSkip.native.d.ts
+++ b/code/core/web/types/helpers/webPropsToSkip.native.d.ts
@@ -32,7 +32,6 @@ export declare const webPropsToSkip: {
     onChange: number;
     onInput: number;
     onBeforeInput: number;
-    onScroll: number;
     onCopy: number;
     onCut: number;
     onPaste: number;


natew:
Thanks for the detailed writeup — the diagnosis is exactly right, but this one already landed on main in fd108d6e1a ("fix(core): preserve native ScrollView onScroll", Fixes #4163), which is the same one-line removal of `onScroll: 1` from `webPropsToSkip.native.ts`. That is why this is showing as conflicting.

Your follow-up note on `onChange` is the genuinely useful part and I want to keep it on the record: `Input.native.tsx` synthesizes a web-shaped `{ target: { value } }` event from `onChangeText`, so unskipping `onChange` would let the raw RN `NativeSyntheticEvent` reach handlers on styled RN hosts while `Input` keeps emitting the web shape — the same prop name carrying two payloads depending on which component you used. Agreed that it should stay skipped until that shim is reconciled.

Closing as already fixed.

PR #4213, opened 2026-09-10
set minH for header popover menu

andrewnoyes:
Currently, on tamagui.dev, when you open the menu in the header, its height doesn't allow for the menu contents to be viewed easily: 
Screenshot_2026-09-10_08-10-28

I just added a minHeight that matches its minWidth, and the menu contents are more visible: 
Screenshot_2026-09-10_08-10-51

### DIFF
--- a/code/tamagui.dev/features/site/header/Header.tsx
+++ b/code/tamagui.dev/features/site/header/Header.tsx
@@ -554,6 +554,7 @@ const HeaderLinksPopoverContent = React.memo((props: { active: ID | '' }) => {
       bg="$background06"
       backdropFilter="blur(40px)"
       maxH="90vh"
+      minH={360}
       maxW={360}
       minW={360}
       elevation="$2"


natew:
Thanks — the symptom is real, but I think this lands one layer above the actual cause, and it has a side effect on the other menus.

`HeaderLinksPopoverContent` sizes itself from a per-menu map, but only on the fine-pointer branch:

```tsx
const pointerFine = !isTouchable
const heights = {
  core: Math.min(maxHeight, 1300),
  ui: Math.min(maxHeight, 1300),
  theme: data?.user ? 300 : 240,
  menu: Math.min(maxHeight, isOnlyShowingMenu ? 1000 : 520),
}
...
{pointerFine ? (
  <YStack height={heights[active]} maxHeight="90vh" overflow="hidden">
    ...
) : (
  <YStack p="4">   {/* no height at all */}
    ...
)}
```

Two things follow. First, `menu` is already 520 on a fine pointer, so `minH={360}` cannot be what changed your screenshot — you must be hitting `pointerFine === false`, where the content is auto-height and collapses. That comes from `isTouchable`:

```ts
export const isWebTouchable = isClient && (`ontouchstart` in window || navigator.maxTouchPoints > 0)
```

which is true on any touchscreen laptop, not just phones. So a desktop with a touch digitizer gets the unheighted branch for every menu. That is the bug worth fixing.

Second, `Popover.Content` is shared by all four menus, so a blanket `minH={360}` also puts a floor under `theme` (240 logged out, 300 logged in) on the fine-pointer path, leaving 60–120px of empty background below the theme menu.

Could you confirm whether the machine in the screenshot is touch-capable? If so the targeted fix is to give the non-`pointerFine` branch a height rather than flooring the shared content — roughly:

```tsx
<YStack p="4" minH={heights[active]} maxH="90vh">
```

which keeps each menu at its intended size on both paths. Happy to take that version.

(Unrelated, but the map still carries a `compiler: 117` entry and the ID union is `core | ui | theme | menu`, so that one is dead.)

PR #4228, opened 2026-09-19
site: align browser tests with deployed v3 site

natew:
Update homepage navigation assertions to the current v3 labels/routes and make the no-JS SSR layout check wait for the remote extracted stylesheet before measuring.

Validated against the deployed v3 Railway preview: docs and homepage Playwright specs, 36/36 passed.


natew:
Folded into #4227 so v3-beta receives one merge and cuts only one beta release.

